PR #1, opened 2020-11-04
full explicit remove memset

tresreid:
removed the memset for the full explicit case. Doesn't seem to affect objects further down the line. tested by running minidoublets with full explicit and segments with unified. same output as when running minidoublets in unified memory. 
Also removed cudaFree for temp objects. 
Fixes all issues raised by cuda-memcheck

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -48,25 +48,25 @@ SDL::Event::~Event()
 
 #ifdef Explicit_MD
     mdsInTemp->freeMemory();
-//    cudaFree(mdsInTemp);
+    cudaFreeHost(mdsInTemp);
 #else
     mdsInGPU->freeMemory();
 #endif
 #ifdef Explicit_Seg
     segmentsInTemp->freeMemory(); 
-//    cudaFree(segmentsInTemp);
+    cudaFreeHost(segmentsInTemp);
 #else
     segmentsInGPU->freeMemory(); 
 #endif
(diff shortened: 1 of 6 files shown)


beiwang2003 (commented):
cudaFree should be there to make sure that things are deleted properly. If debugger compliant about anything related to cudaFree, it is likely that we have bugs in other places. If we remove memset, where the nMDs array being initialized then? We may see correct results now because the compiler tries to initialize it to zero. How about in the case the compiler choose not to initialize the allocated memory by default? 

I've tried the case w/o -DFull_Explicit option, cuda-memcheck issues errors at cudaFree call. If we switch to all unified memory (removing MEMFLAGS), everything runs normal and no errors in cuda-memcheck. 

I think there is a bug somewhere in the explicit memory implementation.

beiwang2003 (changes requested):
ok bugs are identified. cudaMallocHost should pair with cudaFreeHost (but not cudaFree)

beiwang2003 (changes requested):
Again, removing "cudamemset" is not the right way of fixing the issue. cudamemset takes the first argument as pointer to device address (*), but not (**). Try to modify how the first argument is passing to the function. That will fix that issue.

tresreid:
I've made the changes as @beiwang2003 has suggested

beiwang2003 (approved):

beiwang2003:
merge to PR

PR #8, opened 2020-12-11
Explicit hits

tresreid:
load hits using explicit memory management. 
now 3 methods of loading hits:
1) addHitToEventGPU: loads hits via kernel when applicable. takes ~400 ms
https://github.com/tresreid/TrackLooper/blob/33c835b9fca1bb8a7bef8bcbc019f820725f6944/bin/doAnalysis.cc#L665
2) addHitToEventOMP: loads hits to explicit memory via memcpy. makes hits on CPU either serially (~300 ms) or using OMP (~80 ms). This is the best version and should be the default. 
https://github.com/tresreid/TrackLooper/blob/33c835b9fca1bb8a7bef8bcbc019f820725f6944/bin/doAnalysis.cc#L666
3) original method: adds hits to unified memory
https://github.com/tresreid/TrackLooper/blob/33c835b9fca1bb8a7bef8bcbc019f820725f6944/bin/doAnalysis.cc#L668

image


beiwang2003:
@tresreid which commit is the one that introduces hits with pixels (with many cudaMalloc calls)? Since the code is hanging when running with cuda-memcheck, I am trying to narrow down the problem. If it is caused by the hits with pixels, we need to mark it as an issue and move forward for this PR.

tresreid:
390ac37 or possibly 6c04c21. 
If you comment out https://github.com/tresreid/TrackLooper/blob/explicit_hits/bin/doAnalysis.cc#L747
and 
https://github.com/tresreid/TrackLooper/blob/explicit_hits/bin/doAnalysis.cc#L753
you should be able to run cuda-memcheck again and there will be a normal amount of cuda calls.

beiwang2003:
> [390ac37](https://github.com/SegmentLinking/TrackLooper/commit/390ac37c6a79c9464f37a2a29881e282d7f125f3) or possibly [6c04c21](https://github.com/SegmentLinking/TrackLooper/commit/6c04c21cabd41ee5e48c8f8e5684c00023bff1a0).
> If you comment out https://github.com/tresreid/TrackLooper/blob/explicit_hits/bin/doAnalysis.cc#L747
> and
> https://github.com/tresreid/TrackLooper/blob/explicit_hits/bin/doAnalysis.cc#L753
> you should be able to run cuda-memcheck again and there will be a normal amount of cuda calls.

Great. I will mark it as an issue and move forward with this PR.

beiwang2003:
This PR is mainly for addressing explicit memory management. Pixel is under development in this version. So there are issues with Pixel tracking in this version.

beiwang2003:
After commenting out Pixel calls, the cuda-memcheck is able to run (not hanging) and reports 4 error messages. Here is the details. @tresreid can you take a look?
Screen Shot 2020-12-16 at 10 36 51 AM

tresreid:
Addressed in issue #11

Issue #11, opened 2020-12-16
cuda-memcheck hanging for PR#8


beiwang2003:
cuda-memcheck is hanging for: https://github.com/SegmentLinking/TrackLooper/commit/ed6e48e3133cd4e435e8dc8e6a3ffaecc218b739. This is potentially due to large numbers of cudaMalloc and cudaFree to implement the Pixel segments. 
If we comment out https://github.com/tresreid/TrackLooper/blob/explicit_hits/bin/doAnalysis.cc#L747
and
https://github.com/tresreid/TrackLooper/blob/explicit_hits/bin/doAnalysis.cc#L753
we should be able to run cuda-memcheck again and there will be a normal amount of cuda calls.

tresreid:
Hi @beiwang2003 . The issue is due to this line: https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L1493
You would need to comment out the "+ 1" since that refers to the pixel index and we have commented out loading the pixels to get cuda-memcheck to run. Commenting this out doesn't report errors to cuda-memcheck. Idk if this will cause an issue if the pixels were included (since cuda-memcheck hangs at the pixel loading), but we can revisit this when the pixels are redone in a better manner. I think that will avoid the hanging issue since there would be less cuda calls.

beiwang2003:
@tresreid the links in the first comment are broken. Can you fix it?

beiwang2003:
@tresreid commenting out https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L1493 does not help with cuda-memcheck. The code still failed with the same segmentation violation error.

tresreid:
@beiwang2003 did you comment out the whole line or just the +1?

beiwang2003:
> @beiwang2003 did you comment out the whole line or just the +1?

@tresreid they whole line.

tresreid:
just the +1 is what you need to comment out

PR #12, opened 2020-12-16
Revert "Explicit hits"

beiwang2003:
Reverts SegmentLinking/TrackLooper#8 (since cuda-memcheck complaint OOB for TrackCandidate. Need to address before merge.


tresreid:
addressed in issue #11

PR #13, opened 2020-12-17
Pixel loading

tresreid:
Now loads the hits and pixels to explicit memory at the same time with fewer kernels launched and memory allocations. To load the pixels in the MD and segment objects, those objects now have to be allocated in the load hits call. This also gives back cuda-memcheck functionality when loading hits (no hanging).
The print out statements are the same as before (excluding known instabilities) but i find these results questionable as they don't seem to change whether or not pixels are loaded. @bsathian , can you please verify that the pixels are working correctly here and that the output is the same as the CPU versions?
The Hit output was also removed (reports 0) as it doesn't seem meaningful anymore. 

Average Times  Hits       MD   Segments Triplets  pTracklet  Tracklet  Tracks    Total  Total(no load)
explicit    avg:   531.28    11.38   155.28    68.41     9.18   322.84   117.31   1215.67    675.21
explicit C avg:   533.79    11.77   162.99    69.57     2.69   333.46   121.38   1235.65    699.16
Unified    avg:   513.76    22.50   183.10   156.30    13.52   222.03   102.05   1213.24    685.97
unified C avg:   531.53    13.62   168.53    92.23    18.48   290.56   104.68   1219.63    669.61

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -404,34 +404,23 @@ void SDL::Event::addPixelSegmentToEvent(std::vector<unsigned int> hitIndices, fl
   cudaDeviceSynchronize();
   cudaFree(hitIndices_dev); // added by tres
 }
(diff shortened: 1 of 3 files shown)


tresreid:
Hit loading can be made faster by utilizing OMP if desired. Example results for explicit no cache
avg:   204.16    22.71   132.82    76.92    13.61   278.68   117.49    846.40    628.63
Also not that right now all the unified versions use omp to set the number of objects to 0. idk if this is desired or not

tresreid:
add issues will be addressed in #15

PR #15, opened 2020-12-18
Fixing bugs from previous pull request

bsathian:
Multiple bug fixes

1. Pixel segments were not being written into memory due to a counter issue. I had used nSegments[pixelModuleIndex] directly as a counter since I was using unified memory, but now we need to use the tid instead
2. Since pixel segments were not being written into memory, Pixel quadruplet kernel threads were simply starting and immediately stopping, which explains the "too-good-to-be-true" performance of the pixel quadruplets. (Anti climax : The kernel takes anywhere from 4-10 seconds now, at least we matched!). This got compounded by the fact that we didn't have a sanity check in the form of a prinout of the number of pixel quadruplets produced
3. Some other fixes to ensure appropriate memory locations existed for the counters and other arrays so that track candidates can run without a hitch

The task is not over yet, because I still see some multiplicity issues between my personal repo (unified memory) and this one (approx 5-10%) , but hey, we're at least in the ballpark now!

### DIFF
--- a/bin/doAnalysis.cc
+++ b/bin/doAnalysis.cc
@@ -18,21 +18,21 @@ void addPixelSegments(SDL::Event& event, int isimtrk, std::vector<float> trkX, s
     std::vector<float> px_vec;
     std::vector<float> py_vec;
     std::vector<float> pz_vec;
-    std::vector<float> vecX;
-    std::vector<float> vecY;
-    std::vector<float> vecZ;
-    std::vector<int> hitIdxInNtuple_vec0;
-    std::vector<int> hitIdxInNtuple_vec1;
-    std::vector<int> hitIdxInNtuple_vec2;
-    std::vector<int> hitIdxInNtuple_vec3;
+//    std::vector<float> vecX;
+//    std::vector<float> vecY;
+//    std::vector<float> vecZ;
+//    std::vector<int> hitIdxInNtuple_vec0;
+//    std::vector<int> hitIdxInNtuple_vec1;
+//    std::vector<int> hitIdxInNtuple_vec2;
+//    std::vector<int> hitIdxInNtuple_vec3;
     std::vector<unsigned int> hitIndices_vec0;
     std::vector<unsigned int> hitIndices_vec1;
     std::vector<unsigned int> hitIndices_vec2;
     std::vector<unsigned int> hitIndices_vec3;
-    std::vector<float> ptIn_vec;// (ptIn,ptErr,etaErr,pixelSegmentDeltaPhiChange)
-    std::vector<float> ptErr_vec;// (ptIn,ptErr,etaErr,pixelSegmentDeltaPhiChange)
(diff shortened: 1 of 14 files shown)


tresreid:
can you add in "event_times[ev][4] " in this line: https://github.com/bsathian/TrackLooper-1/blob/pr13/bin/doAnalysis.cc#L896

tresreid:
The number of objects produced has changed but not the number in each layer. Also the number of objects produced in each layer should add up to the total number of objects produced. Can you find the cause and fix it?

tresreid:
If adding the module map will speedup the pixels, can you add that in this PR?

tresreid:
there is a "cudaFreeHost" missing. Can you find what isn't being freed and fix it?

tresreid:
https://github.com/bsathian/TrackLooper-1/blob/pr13/SDL/Tracklet.cu#L22 please fix the parenthesis here

bin/doAnalysis.cc
```diff
@@ -824,33 +880,33 @@ int main(int argc, char** argv)
...
     float avg_plet = 0.;
     float avg_tlet = 0.;
     float avg_track = 0.;
     float avg_tot = 0.;
     float avg_noload = 0.;
     for(int ev=0;ev<ana.n_events;ev++){
-        float total =event_times[ev][0]+event_times[ev][1] +event_times[ev][2]+event_times[ev][3]+event_times[ev][4]+ event_times[ev][5]+event_times[ev][6]+event_times[ev][7];
-        float no_load =event_times[ev][1] +event_times[ev][2]+event_times[ev][3]+ event_times[ev][5]+event_times[ev][6]+event_times[ev][7];
+        float total =event_times[ev][0]+event_times[ev][1] +event_times[ev][2]+event_times[ev][3]+event_times[ev][4]+ event_times[ev][5]+event_times[ev][6];//+event_times[ev][7];
+        float no_load =event_times[ev][1] +event_times[ev][2]+event_times[ev][3]+ event_times[ev][5]+event_times[ev][6];//+event_times[ev][7];
```
slava77, line 896: @tresreid 
I think that your comment was referring to this line https://github.com/SegmentLinking/TrackLooper/pull/15#issuecomment-748220363

you can also click on the "+/-" box above and make a suggestion
```suggestion
        float no_load =event_times[ev][1] +event_times[ev][2]+event_times[ev][3]+event_times[ev][4]+ event_times[ev][5]+event_times[ev][6];//+event_times[ev][7];
```

slava77:
> can you add in "event_times[ev][4] " in this line: https://github.com/bsathian/TrackLooper-1/blob/pr13/bin/doAnalysis.cc#L896

I'm curious why these are not in-line comments (click on "Files changed")

bsathian:
> If adding the module map will speedup the pixels, can you add that in this PR?

we don't have it yet. We'll add it later

PR #16, opened 2020-12-23
introduce new CUDA grid method for MD, Segments, Triplets and Trackle…

beiwang2003:
Use flags in Makefile to select the CUDA grid launching methods. 
NESTED_PARA: nested parallelism
NEWGRID_*: new grid launching method

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -103,7 +103,7 @@ void SDL::Event::resetObjectsInModule()
 {
     resetObjectRanges(*modulesInGPU,nModules);
 }
-// add hits via kernel method. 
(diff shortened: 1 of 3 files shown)


slava77:
>  NESTED_PARA: nested parallelism

isn't `dynamic parallelism` a more appropriate name for this?
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-dynamic-parallelism

bsathian (approved):

bsathian:
The multiplicities of the objects between the nested parallelism method and the new grids seem to match. So I'm mergint this

PR #18, opened 2020-12-24
Adding CPU code, efficiency code, and setting benchmark

sgnoohc:


sgnoohc:
I am in the process of adding a doc to reproduce the benchmark point. one more commit is on its way.

sgnoohc:
The large number of files changed is because I deleted a bunch of unnecessary files from rooutil/.

PR #23, opened 2021-01-15
added clip feature and integrated cut value ntuple producer (activate by building with -d flag)

bsathian:


SDL/Event.cu
```diff
@@ -806,13 +806,13 @@ void SDL::Event::createSegmentsWithModuleMap()
     for (int i=0; i<nLowerModules; i++) {
       unsigned int innerLowerModuleIndex = modulesInGPU->lowerModuleIndices[i];
       unsigned int nConnectedModules = modulesInGPU->nConnectedModules[innerLowerModuleIndex];
-      unsigned int nInnerMDs = nMDs[innerLowerModuleIndex];
+      unsigned int nInnerMDs = nMDs[innerLowerModuleIndex] > N_MAX_MD_PER_MODULES ? N_MAX_MD_PER_MODULES : nMDs[innerLowerModuleIndex];
```
slava77, line 809:
```suggestion
      unsigned int nInnerMDs = std::min(nMDs[innerLowerModuleIndex],N_MAX_MD_PER_MODULES);
```
is shorter.
However, is there a `-1` missing?

slava77: >  However, is there a `-1` missing?

no, probably not

slava77: BTW, if the checks are done at write time, perhaps they are not really required at read time

bsathian:
Efficiency plots here : http://uaf-10.t2.ucsd.edu/~bsathian/SDL/efficiencies_20210117/

I'm merging this so that I don't stack PRs on top of each other

PR #24, opened 2021-01-19
Track candidate memory reduction

bsathian:
Turns out there are no track candidates that start at barrel layers 5 and 6, and endcap layers 2,3,4,5. These constitute around 8000 out of the 13000 lower modules. Given that track candidates take up around 6 GB, not allocating memory for these outer layers can save us around 3-3.5 GB. 

Observations post implementation:-
As anticipated, only 5313 modules even have track candidates, and our memory footprint has come down to 12333MiB (12.07 GiB) from 16139MiB (15.76 GiB) - a 23% reduction in memory footprint

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -2213,9 +2213,18 @@ __global__ void createTrackCandidatesFromInnerInnerInnerLowerModule(struct SDL::
                 }
                 else
                 {
-		    unsigned int trackCandidateIdx = innerInnerInnerLowerModuleArrayIndex * N_MAX_TRACK_CANDIDATES_PER_MODULE + trackCandidateModuleIdx;
+//		    unsigned int trackCandidateIdx = innerInnerInnerLowerModuleArrayIndex * N_MAX_TRACK_CANDIDATES_PER_MODULE + trackCandidateModuleIdx;
+                    if(modulesInGPU.trackCandidateModuleIndices[innerInnerInnerLowerModuleArrayIndex] == -1)
+                    {
+                       printf("Track candidates: no memory for module at module index = %d\n",innerInnerInnerLowerModuleArrayIndex);
 
-                    addTrackCandidateToMemory(trackCandidatesInGPU, trackCandidateType, innerObjectIndex, outerObjectIndex, trackCandidateIdx);
+                    }
+                    else
+                   {
+                        unsigned int trackCandidateIdx = modulesInGPU.trackCandidateModuleIndices[innerInnerInnerLowerModuleArrayIndex] + trackCandidateModuleIdx;
+                        addTrackCandidateToMemory(trackCandidatesInGPU, trackCandidateType, innerObjectIndex, outerObjectIndex, trackCandidateIdx);
+                    }
+
                 }
             }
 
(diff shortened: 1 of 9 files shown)


slava77:
> Turns out there are no track candidates that start at barrel layers 5 and 6, and endcap layers 2,3,4,5.

do we really have TCs starting in B4?
I can't make a straight line case starting at B4 and crossing 5 layers in 
SDL_pattern_2021-01-18 at 8 35 01 PM

bsathian:
I see that there are 4 types of track candidates possible starting at layer 4 
(http://uaf-10.t2.ucsd.edu/~bsathian/SDL/cats_from_gpu_PU200.txt, search for "[4")

slava77:
```
[4, 5, 7, 8, 8, 9] 
[4, 7, 7, 8, 9, 10]
[4, 7, 8, 9, 9, 10]
```
all of them look kind of impossible for 1 GeV. Do these patterns ever show up in the muon sample?

slava77:
> The text file [1] was produced using the PU 200 sample and tabulates the different types of outer layer TCs possible and I see that there are 4 types of track candidates starting at layer 4.
> 
> [1] http://uaf-10.t2.ucsd.edu/~bsathian/SDL/cats_from_gpu_PU200.txt, search for "[4"

I guess we can take a note to follow up on these in the future; I added an item in the todo list gdoc

beiwang2003:
I've observed similar things for segment, triplet and tracklet creation. For example, for segments, 9688 out of 13k lower module has non-zero connected modules, 6259 out of 13k for both Triplet and Tracklet. Can we apply the memory reduction for the others too? @bsathian

SDL/Event.cu
```diff
@@ -2209,9 +2213,18 @@ __global__ void createTrackCandidatesFromInnerInnerInnerLowerModule(struct SDL::
                 }
                 else
                 {
-		    unsigned int trackCandidateIdx = innerInnerInnerLowerModuleArrayIndex * N_MAX_TRACK_CANDIDATES_PER_MODULE + trackCandidateModuleIdx;
+//		    unsigned int trackCandidateIdx = innerInnerInnerLowerModuleArrayIndex * N_MAX_TRACK_CANDIDATES_PER_MODULE + trackCandidateModuleIdx;
```
slava77, line 2216: here and in the following: why are these old lines still needed?

slava77:
>  bsathian requested your review on this pull request. 

I don't have much to say from just a visual inspection.
It would be nice to see some before/after printout to be more obvious that the physics/output part is unchanged.

beiwang2003:
> > bsathian requested your review on this pull request.
> 
> I don't have much to say from just a visual inspection.
> It would be nice to see some before/after printout to be more obvious that the physics/output part is unchanged.

beiwang2003:
I've run both nvprof and cuda-memcheck, passed. I also printed out the produced output for a single event. They agree exactly (for the first event).

slava77:
@beiwang2003 
did you close this by accident or intentionally ?

beiwang2003:
> @beiwang2003
> did you close this by accident or intentionally ?

Sorry. I was trying to edit my early comment, but accidentally close it by mistake.

slava77:
I suppose this should be merged soon.
If I were the one to do it, I'd still rather see https://github.com/SegmentLinking/TrackLooper/pull/24#pullrequestreview-571381768 and related cleanup to be done.

Is there anything else pending for this PR?

beiwang2003:
> I suppose this should be merged soon.
> If I were the one to do it, I'd still rather see [#24 (review)](https://github.com/SegmentLinking/TrackLooper/pull/24#pullrequestreview-571381768) and related cleanup to be done.
> 
> Is there anything else pending for this PR?

I don't have anything. It looks fine.

SDL/Event.cu
bsathian: I don't really trust myself to do things right and I was lazy to do git diffs, and so I had commented out the old lines. I can remove them now and push a new commit

tresreid:
Are the validation plots here or posted somewhere?

slava77:
> Are the validation plots here or posted somewhere?

IIUC, the claim is that this is a technical update, not affecting the physics results.
Also, there was a confirmation in https://github.com/SegmentLinking/TrackLooper/pull/24#issuecomment-762937975

tresreid:
It seems like everyone is ok with this so i'm going to merge.

PR #26, opened 2021-01-28
pT4 map and higher pt threshold for the pixel segments

bsathian:
I've added a crude pixel quadruplet map - which is basically not considering very outer layer segments i.e., segments formed between a PS and a 2S module, or between two 2S modules. In addition, I've also increased the pT threshold of the pixel segments. The results are summarized in these sets of slides - http://uaf-10.t2.ucsd.edu/~bsathian/SDL_slides/20210119-pixelModuleMap.pdf and http://uaf-10.t2.ucsd.edu/~bsathian/SDL_slides/20210124-trimmingPixels.pdf

Effects :
1. We see a small drop in efficiency with the pixel module map, especially near the turn on region. Using a module map improves our timing performance by a factor of 2
2. We don't see any significant effect in efficiency by increasing the pT thresholds. This however improves the timing performance by another factor of 2

These two together implies pixel T4s are now created in 1.2 seconds (as opposed to around 5 seconds in the past)


bsathian:
PS : This should incorporate the previous PR by Tres

code/core/trkCore.cc
```diff
@@ -1288,7 +1288,7 @@ float addInputsToLineSegmentTracking(SDL::Event &event, bool useOMP)
         float py = p3LH.Y();
         float pz = p3LH.Z();
 
-        if ((ptIn > 0.7) and (fabs(p3LH.Eta()) < 3))
+        if ((ptIn > 0.99) and (fabs(p3LH.Eta()) < 3))
```
slava77, line 1291: I thought that there was supposed to be `1.0 - 2*ptErr`

bsathian: next PR will have it (or I'll add that commit to this PR here later today)

PR #28, opened 2021-01-29
implement new grid to the last two kernels: createPixelTracklets and createTrackCandidates

beiwang2003:
Due to pixelTracklets, the thread number we launch for createTrackCandidatesInGPU kernel is ~260x more than the case w/o pixelTracklets. It may worth thinking about separating track candidates function for pixel tracklets. 

Also, I've found that --maxrregcounts=64 flag caused significant slow down for createTrackCandidates kernels. I thus remove it for now. At the end, we may need to enforce the register usage per kernel.

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -2710,6 +2710,7 @@ __global__ void createTrackCandidatesInGPU(struct SDL::modules& modulesInGPU, st
   int outerObjectIndex = 0;
   short trackCandidateType;
   bool success;
+
   //step 1 tracklet-tracklet
   if(innerObjectArrayIndex < nTracklets)
     {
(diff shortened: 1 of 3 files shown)


tresreid:
Hi @beiwang2003 , it seems like running out of the box fails around the pT4 stage. Running with the command "./bin/sdl -n 10 -v 1 -m 7 -i PU200" my output looks like this: 
image

Can you please double check that this runs?

beiwang2003:
I changed the Makefile to compile with the old approach, but forgot to change it back. Can you try it again? @tresreid

tresreid:
I think there are still issues here. Some of the versions don't compile. In Particular explicit newgrid.

beiwang2003:
@tresreid @bsathian I've fixed the compiling errors and a few other things. I should be good now.

tresreid:
So I have the timing results for this new version here (i know unified_newgrid is missing). It seems like even with Balaji's module make the new grid version for pT4 and track candidates is still (far) worse off than than without it. Do we really want to include this? Or maybe keep the code but not use new grid for those two objects? 
I haven't checked the actual code yet though.
image

beiwang2003:
> So I have the timing results for this new version here (i know unified_newgrid is missing). It seems like even with Balaji's pixel module map make the new grid version for pT4 and track candidates is still (far) worse off than than without it. Do we really want to include this? Or maybe keep the code but not use new grid for those two objects?
> I haven't checked the actual code yet though.
> image

My understanding is Balaji's pixel module map won't be able to fix the performance issue in the new grid. We will have to consider separating the normal and pixel track to fix this. @bsathian Do you have plan to implement it soon?

tresreid:
Ah, right. We had discussed this yesterday. I had completely forgotten. Sorry.

PR #30, opened 2021-02-05
Add fake rate and duplicate rate plots to the benchmarking scripts

sgnoohc:
- CPU benchmark points are updated.
- Fake rate and duplicate rate plots for the Track Candidate is included.
- with -l option one can also write out lower level objects efficiency/duplicate/fakerate plots for T4+T4x, and T3. (pT4 not included yet)
- with -w option (0 or 1) one can control whether to write the output ntuple or not. -w 0 will turn it off and speed things up for timing result studies.
- bin/sdl.cc now contains the main running workflow (runMD, runT4, etc. etc.) so that the developer can focus on one file to get the overview instead of looking at two different places.
- code/core/write_sdl_ntuple.cc only contains ntuple writing related functions.
- ... And some other changes that I can't remember off the top of my head as I am writing this PR


tresreid:
was there a reason you changed " (ana.verbose == 1) " to " (ana.verbose >= 1) " in trkCore?

PR #31, opened 2021-02-16
Separate kernels for pixel Track Candidate and regular Track Candidate creation

bsathian:


beiwang2003:
For L1390 in Event.cu, should we use nLowerModules-1 instead of nLowerModules? @bsathian 
Screen Shot 2021-02-16 at 10 31 00 AM

beiwang2003:
There are running errors in testing the code:
1. compile: sdl_make_tracklooper -x, run: nvprof ./bin/sdl -n 1 -v 1 -m 7 -I PU200, error:segmentation violation
2. compile: sdl_make_tracklooper -xc, run:nvprof ./bin/sdl -n 1 -v 1 -m 7 -I PU200, error:cudaErrorMemoryAllocation: out of memory
Can you take a look? @bsathian

tresreid:
I guess bei beat me to it, but yeah the explicit versions do not work (particularly the validation fails). I think this also fails for the current master, though. I do see that the explicit validation does work back at commit 0b2eaf4d7b6b4c9a695b1fd43122825650fe4c96 if that helps.

bsathian:
The latest commit in this chain should fix the segmentation faults

tresreid:
some of the timing results are shown here: 
image
Additionally, the validation plots are here: https://www.classe.cornell.edu/~mgr85/SegmentLinking_keep/PR31/efficiencies/
It seems like the validation is working for the explicit cases just fine. And looking at the code, I do think this is running the newer hit loading method, which as we see from the timing results, is much faster. Given that the new hit loading method should be able to support unified memory, i think we should switch to this in all cases and remove all other hit loading methods to avoid further issues. Though, I could do this in a new PR after this is merged.

beiwang2003:
> methods

What's the problem you are running? Can you also put the running command? @tresreid

tresreid:
@beiwang2003 I'm not sure what you mean. But I'm talking about the different hit loading methods found 
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L208
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L313
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L452
and for pixel LS
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L490
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L563
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L596
with the associated helper kernels.
The unified versions by default use the 3rd hit loading method whereas the explicit version uses the 2nd (labeled OMP). The OMP one is the fastest and can support unified memory as well. So i'm saying we should be using the OMP version in all cases and delete the others to avoid further confusion. 
The previous issue was we weren't sure if the omp version loaded the correct simtrks to use for validation. I thought i had fixed it in a previous PR but the more recent PR's weren't running the explicit validation. However, it seems to be working just fine now so i think that issue was indeed fixed. So we should have no problem in using only the OMP hit loading method for all cases.

beiwang2003:
> @beiwang2003 I'm not sure what you mean. But I'm talking about the different hit loading methods found
> https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L208
> https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L313
> https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L452
> and for pixel LS
> https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L490
> https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L563
> https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Event.cu#L596
> with the associated helper kernels.
> The unified versions by default use the 3rd hit loading method whereas the explicit version uses the 2nd (labeled OMP). The OMP one is the fastest and can support unified memory as well. So i'm saying we should be using the OMP version in all cases and delete the others to avoid further confusion.
> The previous issue was we weren't sure if the omp version loaded the correct simtrks to use for validation. I thought i had fixed it in a previous PR but the more recent PR's weren't running the explicit validation. However, it seems to be working just fine now so i think that issue was indeed fixed. So we should have no problem in using only the OMP hit loading method for all cases.

What I mean is just the running command, .e.g, ./bin/sdl -n 1 -v 1 -m 7 -I PU200. I am testing the new grid approach and want to make sure we are running the same problem.

tresreid:
Oh well for what i posted here i just used the typical validation and timing setup. 
sdl_validate_efficiency muonGun
sdl_timing PU200

I think philip changed some of the flag options in the last PR. So the command to run for an individual event as you are doing is 
/bin/sdl -n 1 -v 1 -w0 -I PU200.
The -m 7 is no longer there ( -m 5 is the defualt). the -w 0 flag does what -m 7 used to do (turns off the writing for validation). idk if that fixes your problem though. I at least did not see a segfault after balaji's recent fix when running the validation and timing.

tresreid:
I just double checked and confirmed that it works. if you had an out of memory error on lnx7188, it's possible that someone (myself or possibly Dan) else was also using the gpu as you were using it and the large memory used by this maxed out the gpu? This is just spitballing though. do you still see the issue today? You can change which gpu you are using by putting
CUDA_VISIBLE_DEVICES=x
in front of the running command where x is 0 or 1 depending on which gpu you want to select.

tresreid:
@beiwang2003 , are you still seeing issues or can this be merged?

beiwang2003:
> @beiwang2003 , are you still seeing issues or can this be merged?

Please go ahead with merging. I will then open another PR for the new grid update.

PR #32, opened 2021-02-22
Hit Loading and additional cleanup

tresreid:
switches to using the best working hit loading method. Removes all other hit loading methods to avoid confusion.
changes verbosity levels to something more useful. 0: no printout, 1: timing printout, 2: object multiplicity printout
removes "doAnalysis" from makefile and bin.
validaition results are here: https://www.classe.cornell.edu/~mgr85/SegmentLinking_keep/PR32/efficiencies/


tresreid:
timing results are here:
image

PR #38, opened 2021-04-06
pT4 with pixel Map

tresreid:
This loads and uses the pixel module map for the newgrid method only. 
memory usage:
Unified:   6GB         cache: 10 GB           grid:   6GB        cache grid: 10GB
Explicit: 12GB          cache: 27GB           grid: 12GB      cache grid: 27GB
Validation plots can be found here: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR38/
image


tresreid:
image
With updated pixelMap

bsathian:
I've checked it. Everything looks good. I'm merging this

PR #41, opened 2021-04-21
pixel Map full fix (with safe guards)

tresreid:
validation here: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR41/
This fixes an issue with the pixel map indexing and employs a few safe guards to ensure no crashing.


tresreid:
I think the issue was caused by the order the pixel line segments were filled. With the pixel map, the superbins are made after the other pixel information is pushed back into vectors. However, the superbin creation uses "continue"'s when the pt or eta values of the pLS are out of range. so the superbin vectors are not always the same size as the rest of pixel information. So when running the map, in a few cases, you could be looking at values that aren't properly set which can end up crashing in the kernel. I don't know why we only see the crash in a very specific setup though. This has been fixed and I've put in a few safe guards to avoid crashing just in case. I think these safe guards can be removed later if we choose. The validation looks about the same as before.

tresreid:
image

bsathian:
seems to work well, merging

PR #42, opened 2021-04-21
T5 creation

bsathian:


tresreid:
@bsathian  Can you add the quintuplet validation to to list of plots on the front page of the validation plots. And also post a link to the validation results you get for T5 along with the timing results?

tresreid:
@bsathian the cache allocation versions do not compile. Can you please fix this

SDL/Quintuplet.cu
```diff
@@ -0,0 +1,646 @@
...
+
+
+void SDL::createQuintupletsInUnifiedMemory(struct SDL::quintuplets& quintupletsInGPU, const unsigned int& maxQuintuplets, const unsigned int& nLowerModules, const unsigned int& nEligibleModules)
+{
+    unsigned int nMemoryLocations = maxQuintuplets * nEligibleModules;
+    std::cout<<"Number of eligible T5 modules = "<<nEligibleModules<<std::endl;
+
+#ifdef CACHE_ALLOC
+    cudaStream_t stream = 0;
+    quintupletsInGPU.tripletIndices = (unsigned int*)cms::cuda::allocate_managed(nMemoryLocations * 2 * sizeof(unsigned int));
```
tresreid, line 135: quintupletsInGPU.tripletIndices = (unsigned int*)cms::cuda::allocate_managed(nMemoryLocations * 2 * sizeof(unsigned int),stream);
You need the stream here

SDL/Quintuplet.cu
```diff
@@ -0,0 +1,646 @@
...
+}
+
+void SDL::createQuintupletsInExplicitMemory(struct SDL::quintuplets& quintupletsInGPU, const unsigned int& maxQuintuplets, const unsigned int& nLowerModules, const unsigned int& nEligibleModules)
+{
+    unsigned int nMemoryLocations = nEligibleModules * maxQuintuplets;
+#ifdef CACHE_ALLOC
+    cudaStream_t stream = 0;
+    int dev;
+    cudaGetDevice(&dev);
+    cudaMalloc(&quintupletsInGPU.tripletIndices, 2 * nMemoryLocations * sizeof(unsigned int));
```
tresreid, line 178: Shouldn't these also use cache allocation?

SDL/Makefile
```diff
@@ -21,13 +21,13 @@ CXX                  = nvcc
 CXXFLAGS             =  -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_52 -I/mnt/data1/dsr/cub 
 LD                   = nvcc 
 SOFLAGS              = -g -shared --compiler-options -fPIC --cudart shared -arch=compute_52
-PRINTFLAG            = -DAddObjects -DT4FromT3 #-DWarnings
+PRINTFLAG            = -DAddObjects -DDO_QUINTUPLET #-DT4FromT3 -DWarnings
```
tresreid, line 24: -DT4FromT3 should be on by default (at least when running T4s)

tresreid: When running T4 side by side with T5, This is useful to have a comparison of times.

tresreid:
Also it seems odd that the new grid approach for T5 is slower than the nested version. This is the opposite for every other object

tresreid:
There is a missing free in the unified caching versions.

SDL/Quintuplet.cu
bsathian: I copied this from Track candidates (https://github.com/SegmentLinking/TrackLooper/blame/master/SDL/TrackCandidate.cu#L93-L95)

here for some reason some objects are not cache allocated

tresreid: Those objects exceed the maximum size for cache allocation so they are excluded. The quintuple objects can be allocated using caching just fine

bsathian:
http://uaf-10.t2.ucsd.edu/~bsathian/SDL/T5_efficiency/efficiencies/new_efficiencies/

PR #43, opened 2021-04-23
CPU T5s

sgnoohc:
Performance for muon: http://uaf-10.t2.ucsd.edu/~phchang//analysis/sdl/code/TrackLooper__/workarea/2021_04_19_T5CPU_w_Radius/efficiencies/eff_plots__CPU_645f858_muonGun/index.html

- The track candidates for the CPU is currently = combinations with T3 and T4s but not T5.
- The future plan is to make track candidates = {T5, pT2}.
- The html for the lower level objects have been updated to include T5s as well.
- The write_sdl_ntuple.cc for the cpu's are done, but i noticed for GPU it's not done yet. Mb @bsathian can work on it?

### DIFF
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -210,6 +210,9 @@ void SDL::CPU::Event::addTripletToEvent(SDL::CPU::Triplet tp, unsigned int detId
 
     // And get the layer andd the triplet to it
     getLayer(layerIdx, subdet).addTriplet(&(triplets_.back()));
+
+    // Link segments to mini-doublets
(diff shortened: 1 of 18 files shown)


bsathian:
http://uaf-10.t2.ucsd.edu/~bsathian/SDL/T5_efficiency/efficiencies/new_efficiencies/eff_plots__GPU_explicit_05cd16c_muonGun/mtv/

provides an apples-to-apples comparison (200 events). I see very little difference between CPU and GPU implementations. So I'm merging this

PR #50, opened 2021-05-05
Memory trimming

bsathian:


beiwang2003:
how much memory reduced?

bsathian:
around 6 GB of memory should be reduced, since the memory requirements for TCs are brought down heavily.

PR #52, opened 2021-05-11
Final states

tresreid:
Uses makefile toggles to choose the final state and tracklet optionsl
in tracklooper/Makefile: -DDO_QUINTUPLET and -DDO_QUADRUPLET chooses whether or not to Run the T4 or T5 functions. excluding them can save memory and time.
in SDL/Makefile: -DFINAL_{pT2,T5,T3T4} chooses which objects to include as a track candidate. The necessary toggle in the tracklooper makefile is also necessary to run. each version is independent of each other and can be stacked. 
T3T4 includes pT2.


tresreid:
validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR52/

PR #57, opened 2021-05-26
pT3 newgrid + bugfixes

bsathian:


bsathian:
plots here

http://uaf-10.t2.ucsd.edu/~bsathian/SDL/pT3_efficiency/pT3_newgrid_bugfix/

tresreid:
I think you are missing 5 cudaFrees and 1 cudaFreeHost

tresreid:
The fix to the pT2 newgrid method does show some improvement but the results are still not exactly the same as the nested version.

tresreid:
Free mismatch will be handled later.

PR #60, opened 2021-06-01
pT3 newgrid timing bugfix

bsathian:


bsathian:
Modified the kernel launch parameters so that pT3 newgrid does not take 180ms (instead takes only 50ms)

bsathian:
Before fix

timing

after fix

timing

PR #61, opened 2021-06-02
Implementing pT3 in CPU and adding the T5 fixes

sgnoohc:


sgnoohc:
http://uaf-10.t2.ucsd.edu/~phchang/talks/PhilipChang20210602_LST_Update.pdf

PR #68, opened 2021-07-08
T5 regression and chi-squared based cuts, with pT5 as bonus (only the baseline selections)

bsathian:
efficiency plots here : http://uaf-10.t2.ucsd.edu/~bsathian/SDL/T5_regression/20210708_updates/


slava77:
> efficiency plots here : http://uaf-10.t2.ucsd.edu/~bsathian/SDL/T5_regression/20210708_updates/

it would be useful to have a reference (before this PR) for a comparison

PR #70, opened 2021-07-26
Hit matching

tresreid:
Fixes errors from pT5 
adds T5 duplicate removal
only keeps T5s from first 2 layers. 

Validation plots (T5 only in final state) https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR70/
For PU200 first event: T5 multiplicity 4337 -> 706
For muonGun first event: T5 multiplicity 108-> 33
It pretty consistently takes 10 extra ms for the duplicate removal but cuts the current pT5 creation time in half.


tresreid:
Before duplicate removal: 
image
After duplicate removal:
image

PR #71, opened 2021-07-28
preliminary pT3 duplicate cleaning

tresreid:
This is not the final pT3 duplicate removal. It is intended to help with understanding the effect on pT5 creation when pT3s are cleaned.
pT3 multiplicity: 24028 - > 7220
track candidate final state in validation plots: T5 + pT3
Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR71/
Requirements: 4 matched pixel hits and 4+ matched triplet hits. No eta, phi requirement. Save smaller r-phi track score only.

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -3977,7 +3977,7 @@ __global__ void createPixelTripletsInGPUFromMap(struct SDL::modules& modulesInGP
         float eta_pix = segmentsInGPU.eta[pixelSegmentArrayIndex];
         float phi_pix = segmentsInGPU.phi[pixelSegmentArrayIndex];
         float pt = segmentsInGPU.ptIn[pixelSegmentArrayIndex];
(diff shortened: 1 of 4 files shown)


bsathian:
A few comments

 Cleaning the pT3s and T5s blindly results in pT5 efficiency plummeting (https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR71/PU200_AfterRemoval/eff_plots__GPU_unified_newgrid_3eab7dcDIRTY_PU200/mtv/pT5_AllTypes__pt_eff.pdf)

This happens mainly due to pT3s being cleaned (because cleaning the T5s didn't affect the pT5s). So, a few permuations were tried (summaries here : http://uaf-10.t2.ucsd.edu/~bsathian/SDL_slides/20210729-DuplicateCleaning.pdf, http://uaf-10.t2.ucsd.edu/~bsathian/SDL_slides/20210803-crossCleaningUpdate.pdf) and the best one has the setup described in the second set of slides (code here : https://github.com/bsathian/TrackLooper-1/tree/pr_71)

In short, this procedure goes the following way:-

1. Clean T5s
2. Construct pT5s (without cleaning the pT3s)
3. When cleaning pT3s, instead of blindly dropping the pT3 with the higher track score, we also check if the said pT3 is part of a pT5. If so, the other pT3 is dropped by default

PR #74, opened 2021-08-13
Inclusion of pLS, pT5 and pT3 cross cleaning and duplicate removal

tresreid:
This cleans duplicate pLS (using dR), pT5 (using hit matching in outer tracker and pLS ID matching) and pT3 (using 4+ outer hits and pLS ID matching) to remove duplicates. Also removes pT3 cross cleaning. 
Produces ~20% duplicate rate  for all TCs
Note that none of these methods are optimized for physics performance or time. 
Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR74/

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1494,6 +1494,18 @@ void SDL::Event::createTrackCandidates()
 #endif
     }
 
+#ifdef FINAL_pT5
+    printf("Adding pT5s to TC collection\n");
+    unsigned int nThreadsx_pT5 = 1;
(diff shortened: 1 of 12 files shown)


tresreid:
image

tresreid:
Additional plots for pT5 and pLS added to front page. 
pT3 efficiency is now with respect to sim tracks that are not found by pT5s.

tresreid:
image
updated timing

PR #76, opened 2021-08-31
pT3 refining cuts

bsathian:
http://uaf-10.t2.ucsd.edu/~bsathian/SDL/pT3_new_cuts/eff_comparison_plots__d160b9aDIRTY90pc_explicit_newgrid_ea2ce14_explicit_newgrid_PU200/

"reference" in the above plots refer to the current master

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -4118,8 +4118,6 @@ __global__ void createPixelTripletsInGPUFromMap(struct SDL::modules& modulesInGP
     int segmentModuleIndex = seg_pix_gpu_offset[offsetIndex];
     int pixelSegmentArrayIndex = seg_pix_gpu[offsetIndex];
     if(pixelSegmentArrayIndex >= nPixelSegments) return;
-    if(segmentsInGPU.isDup[pixelSegmentArrayIndex]) return;
-    if(segmentsInGPU.partOfPT5[pixelSegmentArrayIndex]) return; //don't make pT3s for those pixels that are part of pT5
     if(segmentModuleIndex >= connectedPixelSize[pixelSegmentArrayIndex]) return;
 
     unsigned int tripletLowerModuleIndex; //index of the module that connects to this pixel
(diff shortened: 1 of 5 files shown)


tresreid:
It looks like there is only one r-phi chi2 score for pT3s whereas there are two for pT5s (inward and outward) and 2 for T5s (anchor hits and non-anchor). Typically, using both is useful for duplicate removal and I would expect it would also be useful for reducing fakes. Can you do the other r-phi score for pT3s as well (i assume the inward one, which I recall performed better for pT5s but that does have more hits to work with...)

bsathian:
Update : Added inward pT3 cuts. 
Plots here : http://uaf-10.t2.ucsd.edu/~bsathian/SDL/pT3_new_cuts/both_cuts_90pc/eff_comparison_plots__16250e7_explicit_newgrid_ea2ce14_explicit_newgrid_PU200/
Also, cutting this harsher results in significant efficiency loss for the pT5s (since pT3 selections are applied to the pT5s also) : http://uaf-10.t2.ucsd.edu/~bsathian/SDL/pT3_new_cuts/both_cuts_85pc/eff_comparison_plots__16250e7DIRTY_explicit_newgrid_ea2ce14_explicit_newgrid_PU200/pT5_AllTypes__pt.pdf

so I won't cut this further (90% is the final)

tresreid:
Why are pt3 selections applied to pt5? There shouldn't be a reason for this. These should have separate cuts.

tresreid:
Other than separating the pt5 and pt3 cuts, I think the fake rate for pt3s looks much better. And I think fixing back the pT5 cuts will bring back the TC efficiency. So once that is done I think this can be merged. we'd get a much more flat ~20% fake rate.

bsathian:
http://uaf-10.t2.ucsd.edu/~bsathian/SDL/pT3_new_cuts/both_cuts_90pc_no_pT3_cuts_for_pT5/eff_plots__GPU_explicit_newgrid_16250e7DIRTY_muonGun/mtv/TC_AllTypes__pt_eff.pdf

http://uaf-10.t2.ucsd.edu/~bsathian/SDL/pT3_new_cuts/both_cuts_90pc_no_pT3_cuts_for_pT5/eff_comparison_plots__16250e7DIRTY_explicit_newgrid_ea2ce14_explicit_newgrid_PU200/TC_AllTypes__pt.pdf

PR #77, opened 2021-09-07
Baseline: duplicate removal and cross cleaning

tresreid:
Applies duplicate removal and cross cleaning for all objects. 
Validation plots are here: https://www.classe.cornell.edu/~mgr85/SegmentLinking_keep/PR77/


tresreid:
Timing
image

tresreid:
TC % breakdown:
image

PR #86, opened 2021-10-01
Bei's initial timing improvements

tresreid:
Validation: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/baselinePlots/PU200_bei/
Bei's timing improvements include inline funcitons and using fastmath


tresreid:
image

PR #94, opened 2021-11-18
Multistreaming

tresreid:
Reduces memory usage (kernels.cuh)
Restructures the setup of the code
-- reads root files upfront
--runs over events in parallel CPU threads using openMP
-- each CPU thread takes its own cuda stream and event
--all kernels, memcpy and memsets are Asynchronous. uses cudaStreamSynchronize instread of device sync
number of streams set from command line with -s <nStreams> flag. Default =1
nBlocks set to 80 in kernels.cuh


tresreid:
Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR94/

tresreid:
Timing:
image

slava77:
> Timing:

is the throughput column something to read or mostly ignore for now? (unified s=1 being the best looks odd; here I'm assuming that as in other places the throughput is the number of events per unit time)

the "effective time" shown on Tuesday was a bit more intuitive. It may be better to add a column for it.

tresreid:
So kernel and kernel short replace the previous times we have look at until now. Loop is the time it takes to run over all the events divided by the number of events. And throughput (i guess inverse throughput) is the effective time = (kernel_short/ kernel ) * loop. So explicit cache with 6 streams works the best giving an effective time of 30 ms per event.

tresreid:
Changed the timing format to be less confusing
image

beiwang2003:
Can you elaborate what kernel, kernel (short), loop, effective mean in the timing table?

beiwang2003:
@tresreid Can you elaborate what kernel and kernel (short) mean in the timing table?

tresreid:
@beiwang2003 Kernel is the sum of all the kernel times (what was previously timing) whereas kernel short(previously timing short) is the sum of the kernel times we care about (it excludes the hit loading and pLS duplicate removal). Kernel short was what we were previously using to report our times. Loop time is the time it takes to run over all the events divided by the number of events. For a single stream should be close to the kernel time but may include some overhead. Effective time (which i just changed from throughput) is kernel_short / kernel * loop. It is the timing we can expect for the part of the code we care about over multiple streams. This is the time we want to use now. For a single stream it should hopefully be close to the kernel short time.

sgnoohc:
Tested unified and explicit on phi3 on PU200 ttbar 50 events and I seem to get reasonable output from the validation plots. LGTM!

PR #95, opened 2021-11-22
Pt0p8

sgnoohc:
Performance plots: http://uaf-10.t2.ucsd.edu/~phchang/efficiencies/eff_plots__GPU_explicit_f004f4a_PU200/

```
Timing summary
Evt     Hits         MD       LS      T3       TC       T5       pT3      pT5      pLS      Total     Total(short)
   avg     788.       4.      18.      19.      18.      43.      23.      31.      30.      971.      154.+/- 77.   explicit
```


tresreid:
Can you fix the zoomed efficiency plot. It's hard to see where the efficiency plateaus.

tresreid:
Thanks. How many events are you running over in the validation? 50?

sgnoohc:
500

tresreid:
Do you think you can show the efficiency plots without the eta extension in the plots (or is there a place to change that easily). I'm wondering if that accounts for the efficiency drop.

sgnoohc:
I believe it is the cause of efficiency "drop." But I suppose at the end it is what it is.

At the end of the day we will be comparing efficiency between ours and some alternative (e.g. CKF).
Since they (CKF or rest of CMS) produces MTV plots with a strict definition, I think we should do the same.

We can always work towards the "relative" direction. (improve efficiency/FR/DR in certain area to be +/-/- direction.)

tresreid:
This is turned on by default right? I pulled this along with the multistream stuff and I get these times which are comparable to the previous multistreaming results from PR #94 .
image

sgnoohc:
ah. i forgot. it's -8 option during compilation

tresreid:
This doesn't seem to work in conjunction with PR94. It seems to fail around the segment production step. So at the very least both PRs should not be merged at once.

sgnoohc:
I wonder if it has to do with larger multiplicities due to lowered pt threshold..?

sgnoohc:
NOTE: The 0.8 GeV threshold will make the multiplicities somewhat blow up and require some tuning to make it run.

PR #96, opened 2021-11-30
pt0p8 Patch

tresreid:
changed max object values and validation/ timing scripts to support the reduction in the pixel pt thresholds.
Note that this only allows for up to 2 streams for the explict cache version. Previously, explicit cache with 4-6 streams was the fastest time.


tresreid:
Validation:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR96/

tresreid:
Timing
image

sgnoohc:
I can reproduce results from both lnx7188 and phi3. I am merging.
I noticed explicit_cached stream=4 fails with out of memory, but mb this is OK.

PR #99, opened 2022-01-05
Half Precision Storage

tresreid:
Replaces many instances of stored floats as stored __half. Toggleable by BOTH makefiles. base and dPhi can be changed without a drop to the efficiency. The other switches do decrease the efficiency, but I suspect this can be recovered (to varying degrees) by redoing the threshold cuts. Integers and Booleans were not changed. This only goes down to the segment level, not mini doublets (pT5, pT3, T5, T3 and segments are changed). Values are currently promoted back to full precision floats when used.
Explicit cache data usage goes from ~12 GB to ~8 GB with the baseline change and ~7 GB with the other cuts. The baseline allows for 4 concurrent streams. 
Validation plots for the baseline change with 200 events:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR99/base/
Validation plots for a variety of changes with 30 events: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR99/extra


tresreid:
image

PR #97, opened 2021-12-10
Track extensions

bsathian:
http://uaf-10.t2.ucsd.edu/~bsathian/SDL_plots/20211210-PullRequest/eff_plots__GPU_unified_efcd8c0_PU200/mtv/

http://uaf-10.t2.ucsd.edu/~bsathian/SDL_plots/20211210-PullRequest/eff_plots__GPU_unified_efcd8c0_PU200/mtv/TC_v_TCE_AllTypes__pt.pdf

PR #101, opened 2022-01-12
Quick Extension Update

tresreid:
This updates PR #97 by fixing the validation plotting and timing to include the track extensions.
There is a switch withing the sdl_validate_efficiency script to toggle the extensions. track extensions are also included on the front page of the validation plots. They revert to the original track candidates when the switch is not used. 
Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR100/


tresreid:
image

tresreid:
image

PR #102, opened 2022-01-14
Half storagev3

tresreid:
Replacement for PR #99 .
half storage base including track extensions. Does not have an implementation for floats that can change efficiency.


tresreid:
validation plots: https://www.classe.cornell.edu/~mgr85/SegmentLinking_keep/PR102/

tresreid:
image

PR #105, opened 2022-01-25
Memory optimization

bsathian:
NOTE : Still in development. Validation plots and timing will be posted after everything is done


bsathian:
PR_timing

slava77:
> A few ms here and there, but not a lot of gains :(

please also post the "before" results

PR #107, opened 2022-02-14
Changed Minidoublet creation

tresreid:
Pulls out hits from minidouble creation kernel. Some small timing improvements.


tresreid:
current master timing:
image

tresreid:
New timing:
image

PR #109, opened 2022-02-16
reduce threads for removal of pixel objects

tresreid:
This reduces the number of threads used to launch the removeDupPixel triplet and quintuplet kernels. This is associated with a 10 ms speedup for the best time.


tresreid:
old timing:
image

PR #108, opened 2022-02-15
Module coalescing

bsathian:
http://uaf-10.t2.ucsd.edu/~bsathian/SDL_plots/moduleCoalescing_PR/eff_plots__GPU_explicit_0991b84_PU200/

PR #110, opened 2022-02-16
bug fix, explicit transfer to host, Fixed cache efficiencies

tresreid:
Changed the explicit to host transfers for validation for the Md and segments to match the change from nModules->nLowerModules. This caused a sync error. not sure how previous validation was done but this fixes the issue. 

Cache efficiency is now fixed by adding in memsets for Boolean values that were not set by default.

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1194,7 +1194,7 @@ void SDL::Event::createSegmentsWithModuleMap()
     int nonZeroModules = 0;
   #ifdef Explicit_Module
     unsigned int* nMDs;// = (unsigned int*)malloc(nModules*sizeof(unsigned int));
-    cudaMallocHost(&nMDs, (nModules +1)* sizeof(unsigned int));
+    cudaMallocHost(&nMDs, (nLowerModules +1)* sizeof(unsigned int));
     cudaMemcpyAsync((void *)nMDs, mdsInGPU->nMDs, nLowerModules*sizeof(unsigned int), cudaMemcpyDeviceToHost,stream);
 
     unsigned int* module_nConnectedModules;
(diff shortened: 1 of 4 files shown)


tresreid:
Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR109/

slava77:
> Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR109/

it's been a while since I looked at the plots : 
image

is this real? CPU up to 20% more efficient than GPU.

slava77:
> Validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR109/

another thing, this PR is 110, while the link is 109. Is the link a reference or actually this PR?

slava77:
also, are the variants supposed to have the same efficiency or is something broken?

https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR109/eff_plots__GPU_explicit_cache_9a119ab_PU200/mtv/TC_AllTypes__etazoomcoarse_eff.png
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR109/eff_plots__GPU_explicit_9a119ab_PU200/mtv/TC_AllTypes__etazoomcoarse_eff.png
explicit and explicit cache
look quite different
image
image

tresreid:
@slava77 , 
caching and non caching version should have the same efficiency but they do not and it's been like this for a while, since before changing the cache binning. I don't know the cause. 
I don't normally follow the CPU version, I was under the impression that it was out of date. In any case, the GPU and CPU versions have not matched for a very long time, so this isn't a recent occurrence. If this comparison is made by just matching point by point (which I think it is), I think the results make sense since the GPU has a lower efficiency after going down to the 0.8 GeV pt threshold.
This PR was supposed to be part of #109 but balaji merged it before It was ready. I found the bug when i was running validation for #109 , which was caused by #108 . I do not know how balaji made the validation plots for #108 with this issue. That's why the plots are labeled 109 and not 110. I moved the plots before realized the other version was already merged.

tresreid:
The cache/non-cache difference is caused at the pT3 level it seems

bsathian:
I think when I merged my PR with master after merging your PR there, things changed. However, I made the efficiency plots before the merge and so it went smoothly.

tresreid:
Post cache fix validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR110

tresreid:
image
New Timing

PR #111, opened 2022-02-24
Half integer modules + Merging Tres' PR

bsathian:
Plots : http://uaf-10.t2.ucsd.edu/~bsathian/SDL_plots/uint16_pr/efficiencies/

timing_new


bsathian:
Screenshot from 2022-02-24 20-19-32

Timing after merging Tres' PR

slava77:
> Screenshot from 2022-02-24 20-19-32
> 
> Timing after merging Tres' PR

Do I understand correctly that there is a slow-down?

In #112 vs the plot above 

imageimage

In particular, the fastest time
image
image

tresreid:
There might just be some variation in the run. No speedup but no real slow down either. I think this is good to go so i am merging. 
image
image
image

PR #112, opened 2022-02-24
Hit migration and improved duplicate removal steps

tresreid:
Removed Track extension flag so that TCE is always done
Ensure hits indicies are set for objects up to TC 
Changed duplicate removal steps (T5,pT5,pT3) to used saved hit indices instead of navigating the chain of objects to get the hits
Changed first pass of T5 duplicate removal. Compares fewer T5s but overall recovers good efficiency with similar (slightly better) final fake and duplicate rates. Makes slightly more pT5s.


tresreid:
Validation plots : https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR112/

tresreid:
timing
image

SDL/Kernels.cu
```diff
@@ -877,19 +878,149 @@ __device__ int inline checkHitsT5(unsigned int ix, unsigned int jx,struct SDL::m
     }
     return nMatched;
 }
+__device__ int inline checkHitspT5(unsigned int ix, unsigned int jx,struct SDL::pixelQuintuplets& pixelQuintupletsInGPU)
+{
+    unsigned int hits1[14];// = {-1,-1,-1,-1,-1,-1,-1,-1,-1,-1};
+    unsigned int hits2[14];// = {-1,-1,-1,-1,-1,-1,-1,-1,-1,-1};
+
+
+    hits1[0] = pixelQuintupletsInGPU.hitIndices[14*ix];
```
bsathian, line 887: Instead of creating new memory locations, we can initialize a pointer to the first hit and then index starting from there. I'll make this change in my PR

bsathian: On second thought, looks like changing this to direct accessing from global memory using pointers increases the timing of pT5 by a full 8ms for the first event (50%). So I won't make this change

PR #113, opened 2022-02-28
working TCE time reduction

tresreid:
Redoes the kernel launch and distribution of work for threads in the TCE creation kernel.
Validation: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR113/

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1425,26 +1425,28 @@ void SDL::Event::createExtendedTracks()
     createTrackExtensionsInUnifiedMemory(*trackExtensionsInGPU, nTrackCandidates * N_MAX_TRACK_EXTENSIONS_PER_TC, nTrackCandidates, stream);
 #endif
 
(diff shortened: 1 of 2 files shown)


tresreid:
timing: 
image

slava77:
I'm curious about the apparently bad scaling of the TCE kernel vs the number of streams. Everything else seems to be roughly linear. Perhaps it's just round-offs. 
@tresreid 
Please add one more significant digit to the printout values below 2ms.

tresreid:
image

slava77:
> image

thanks. Well, this doesn't look like a round-off; OTOH, perhaps there is an issue in the timing resolution at low values.
e.g. 0.06 for s=1 seems too small. 
Is this an average over many events? how many?

tresreid:
Ok Actually i don't think this is real. I think i missed up somewhere.

PR #114, opened 2022-03-02
Extension kernel cleanup

tresreid:
redoes the division of work for each thread in the track extensions.
blocks launched are close to the previous values without the host calculations. 
Bei found the issue with the sync. The previous stream sync did not work because the kernel launch was not launched on the stream. This is now fixed and the timing seems ok.


tresreid:
Validation plots:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR114/

tresreid:
timing: 
image

PR #115, opened 2022-03-04
Change Behavior for Storing Maximum Values of Objects

VourMa:
This PR changes the behavior of the `n[X]` struct members to not exceed the maximum value of the `[X]` object, as set in `Kernels.cuh`.

To accommodate occupancy studies, a new struct member is introduced, `totOccupancy[X]`.

### DIFF
--- a/SDL/Kernels.cu
+++ b/SDL/Kernels.cu
@@ -688,6 +688,7 @@ __global__ void createPixelQuintupletsInGPUFromMap(struct SDL::modules& modulesI
 
             if(success)
             {
+                atomicAdd(pixelQuintupletsInGPU.totOccupancyPixelQuintuplets, 1);
                 if(*pixelQuintupletsInGPU.nPixelQuintuplets >= N_MAX_PIXEL_QUINTUPLETS)
                 {
 #ifdef Warnings
--- a/SDL/Kernels.cu
+++ b/SDL/Kernels.cu
@@ -133,6 +133,7 @@ __global__ void createSegmentsInGPU(struct SDL::modules& modulesInGPU, struct SD
 
             if(success)
             {
+                atomicAdd(&segmentsInGPU.totOccupancySegments[innerLowerModuleIndex],1);
                 if(segmentsInGPU.nSegments[innerLowerModuleIndex] >= N_MAX_SEGMENTS_PER_MODULE)
                 {
 #ifdef Warnings
--- a/SDL/Kernels.cu
+++ b/SDL/Kernels.cu
@@ -68,16 +68,15 @@ __global__ void createMiniDoubletsInGPU(struct SDL::modules& modulesInGPU, struc
 
             if(success)
             {
-                unsigned int mdModuleIndex = atomicAdd(&mdsInGPU.nMDs[lowerModuleIndex],1);
-                if(mdModuleIndex >= N_MAX_MD_PER_MODULES)
+                if(mdsInGPU.nMDs[lowerModuleIndex] >= N_MAX_MD_PER_MODULES)
                 {
-                    #ifdef Warnings
-                    if(mdModuleIndex == N_MAX_MD_PER_MODULES)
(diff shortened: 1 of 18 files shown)


VourMa:
Link to validation plots:
https://www.classe.cornell.edu/~evourlio/SDL_GPU/PR115_Validation/

tresreid:
This looks good to me. Can you run timing and we can merge after the meeting today.

VourMa:
The timing is already run, included in the validation link above. Let me also post the results here to have a clearer picture:

**Before PR:**
image

**After PR:**
image

Let's also discuss any details in the meeting and merge after. Thanks!

tresreid:
The fastest times are typically with explicit cache. Do you have that? I don't see it in the link.

VourMa:
No, I haven't run that. I will do it now and post the results.

VourMa:
### Timing for explicit_cache:

**Before PR:**
image


**After PR:**
image

VourMa:
I added a commit to deal with some comments, as discussed in the meeting yesterday.

I have also tested that setting the variables `N_MAX_MD_PER_MODULES`, `N_MAX_SEGMENTS_PER_MODULE`, `N_MAX_TRIPLETS_PER_MODULE`, `N_MAX_QUINTUPLETS_PER_MODULE`, `N_MAX_TRACK_EXTENSIONS_PER_TC`, `N_MAX_PIXEL_MD_PER_MODULES`, `N_MAX_PIXEL_TRIPLETS`, `N_MAX_PIXEL_QUINTUPLETS` to values <10% of their current ones does **not** crash the code.

I noticed that low values for the variable `N_MAX_PIXEL_SEGMENTS_PER_MODULE` does crash the code, however. Dealing with this variable is not part of this PR, so I propose to go ahead and merge it and create an issue for this specific variable.

SDL/Kernels.cu
```diff
@@ -68,16 +68,16 @@ __global__ void createMiniDoubletsInGPU(struct SDL::modules& modulesInGPU, struc
...
-                    if(mdModuleIndex == N_MAX_MD_PER_MODULES)
-                        printf("Mini-doublet excess alert! Module index =  %d\n",lowerModuleIndex);
-                    #endif
+#ifdef Warnings
+                    printf("Mini-doublet excess alert! Module index =  %d\n",lowerModuleIndex);
+#endif
                 }
                 else
                 {
+                    unsigned int mdModuleIndex = atomicAdd(&mdsInGPU.nMDs[lowerModuleIndex],1);
```
slava77, line 80: I think that this part is incorrect: `mdModuleIndex` should be computed first so that it's a fixed value in a current thread. As written, between a load on L72 where the check is made and the `atomicAdd` call the value could've been already incremented in another thread.
So, an occasional over-write is possible, especially in region1 (2S barrel layers).
Also, it's even more likely that `totOccupancyMDs` and `nMDs` are out of sync in these cases.

tresreid: I'm not sure if this has already been said, but couldn't we put the check on totOccupancy instead of nMDs. As I understand, currently two threads can run the nMD check and then both would pass and increment the counter. So you need the atomic add to happen before the check. But then nMDs would keep incremented (Even when the value isnt saved). But just put the check on totOccupancy which you want to keep incrementing anyway. then nMDs inside the if statement would not be incremented if totOCcupancy is already above the Max and those values would not be written either.

VourMa: I don't think this has been already mentioned. I am not sure that it will work but I can try for sure. Thanks for the suggestion.

slava77: It may work.

I could imagine the following though.
 Let's say before two concurrent threads A and B tot = M-3, nM=M-3 (where M is max).
- A: tot_new = M-1 (happened second), nM_new = M-2 (happened first)
- B: tot_new = M-2 (happened first), nM_new=M-1 (happened second)

but I guess we do not care much if tot and nM can be out of sync.

PR #116, opened 2022-03-08
Variable occupancy

bsathian:
MD and segments work, T3s have issues

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -546,7 +546,7 @@ void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::
     const int loopsize = x.size();// use the actual number of hits instead of a "max"
     uint16_t nLowerModules;
     cudaMemcpyAsync(&nLowerModules,modulesInGPU->nLowerModules,sizeof(uint16_t),cudaMemcpyDeviceToHost,stream);
-cudaStreamSynchronize(stream);
+    cudaStreamSynchronize(stream);
     //printf("loopsize %i\n",loopsize);
 
     if(rangesInGPU == nullptr)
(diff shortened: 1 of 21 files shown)


bsathian:
Splitting into two PRs - MD + Segment, and T3  following Slava's suggestion

bsathian:
Update : Now we have variable occupancy for MD, Segment and T3. In addition, T5s are only created for barrel layers 1 and 2, and endcap layers 1 (and only for those modules which have nonzero T3s). On first glance, it looks like T3 memory footprint has halved. 

Screenshot from 2022-03-11 17-50-20


http://uaf-10.t2.ucsd.edu/~bsathian/SDL_plots/variable_pr/eff_plots__GPU_explicit_c17aaa8_PU200/

bsathian:
The timing has increased, possibly due to accessing more array locations containing starting module locations (and associated stalls). I'll investigate this

bsathian:
T
Screenshot from 2022-03-17 21-11-47
iming afer the corrections

slava77:
> T Screenshot from 2022-03-17 21-11-47 iming afer the corrections

looking at the single-thread explicit cache, 
- pT3 time is back to 9 ms as it was before this PR
- T5 time is still quite high at 22 ms (was 19 ms before this PR).

Is this correct, or did I miss something?

tresreid:
I think a T5 time of 22 ms for explicit cache single stream is pretty standard. There is some variation with these reported times. looking back at the last 2 PRs, I see 19, 20 and 21. For a more accurate depiction of a timing change, i think you would need to look at the kernel runtimes from the profiler before and after. those numbers are more "stable"

tresreid:
This is ready to be reviewed then?

PR #123, opened 2022-03-18
Hitpreload

tresreid:
Uses the -l option in compiling to toggle hit preloading on.

Preloading initializes a large Hit struct and fills it with the hits for all events before runtime. During runtime, the hits stage will copy the hits for that event to a smaller hit array used throughout the code. Pixel line segments are loaded in the same way. 
Things to be done:
- save precalculated variables for the hit
- save pre-duplicate cleaned pLS
- investigate remaining large hit runtime
- possibly preload hits via streaming


tresreid:
old timing:
image

tresreid:
new timing:
image

slava77:
> new timing: image

where does the time counter start in the combined contributions above (more specifically, Event and Loop) relative to the preload start/end?
I'm assuming that the preload is still synchronous.

tresreid:
> > new timing: image
> 
> where does the time counter start in the combined contributions above (more specifically, Event and Loop) relative to the preload start/end? I'm assuming that the preload is still synchronous.

The preloading happens before any of this timing is done. However, the hits column does include both the loading of the PLS (as per usual) as well as copying the hits for that event from the large hit arrays

tresreid:
Validation:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR123/

tresreid:
preloading for 200 events explicit (cache) uses about 5( 6.4) GB.

slava77:
> During runtime, the hits stage will copy the hits for that event to a smaller hit array used throughout the code.

why is this extra copy necessary? Can it be just an update of a pointer to start reading from the same place as the destination of preload?
Are the hits preloaded as structs or SoAs?

tresreid:
The hits are preloaded in the same container as previously. One can have some offset to start at the correct value in the large array, but then this would need to be propagated throughout the code. Setting the pointer did not work for me originally, but after fixing some bugs it might work now. I didn't think the copy would be very problematic since for a single stream this takes 1ms. I haven't investigated why this blows up for multithreading yet though.

tresreid:
Looking at the produced timeline (for 8 streams) it does seem like the large hits time is truly produced by the D2D memcpy. However, this processes usually overlaps with kernels in another stream. So it is not clear to me that this actually hurts our throughput. at least, not unless all processes have this transfer coincidently overlapping at the same time (obviously we should still try to reduce this).

tresreid:
> Looking at the produced timeline (for 8 streams) it does seem like the large hits time is truly produced by the D2D memcpy. However, this processes usually overlaps with kernels in another stream. So it is not clear to me that this actually hurts our throughput. at least, not unless all processes have this transfer coincidently overlapping at the same time (obviously we should still try to reduce this).

Oh, but this is a device to device load which would probably use up some bandwidth on the gpu that could otherwise be used in a kernel. So that would make sense why this slows down for multithreading and can overall hurt performance.

slava77:
Is the D2D copy per event done before the kernel starts?
If so, I'd try to avoid it as much as possible; our assumption is that data is already available in an event.

tresreid:
There isn't really a kernel for the hit loading. d2d happens withing the run though. I'm looking for alternatives. just assigning the pointer doesn't seem to work, but I think that is because i was doing it on the host and these pointers are only on the device (despite the struct itself being on the host). I haven't had time to check that yet.

tresreid:
an update with an assignment to the hit pointer as opposed to using memcpy. Though this does remove the large D2D copies, this did not correspond to a large reduction in the hit loading time for multiple streams. There is some small improvement overall to the throughput ,though. Cleanup and full validation still needs to be done, but first few events looked ok.
image

tresreid:
timing with pixels only: 
image
timing with hits only:
image

So I guess the remaining issue is mostly with pixel loading as opposed to the actual hit loading.

tresreid:
Final times:
image

tresreid:
Final Validation:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR123/eff_plots__GPU_explicit_2805c14_PU200/#1.5

tresreid:
Additional validation plots with cut at 2.4 eta
PU200: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR123_eta2p4/
Muon:  https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR123_eta2p4_muon/

bsathian:
Things seem to run on my side. Merging.......

PR #126, opened 2022-03-23
Script for validating the efficiency of duplicate objects

VourMa:
This PR includes a validation script to print info on the duplication of matching in the sim track, TC, pT5, pT3 and T5 collections.

A table with the results can be found [here](https://docs.google.com/spreadsheets/d/1RGS98NUFv17jSe4R5w2qu6XCqOczH30wNN3FWd92s7U/edit?usp=sharing).

### DIFF
--- a/efficiency/python/dupObjEff.py
+++ b/efficiency/python/dupObjEff.py
@@ -331,13 +331,13 @@ def printTCComp(objType,dict_):
 
   print ""
   print "Total "+objType+" multiplicity in TC collection = %d" %dict_["N_"+objType]
-  print objType+" fakes = %d (%.2f%%)" %( dict_["N_"+objType+"fakes"], float(dict_["N_"+objType+"fakes"])/float(dict_["N_"+objType])*100 )
-  print objType+" duplicates with pT5 = %d (%.2f%%)" %( dict_["Ndup_"+objType+"WpT5"], float(dict_["Ndup_"+objType+"WpT5"])/float(dict_["N_"+objType])*100 )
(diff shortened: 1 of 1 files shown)


tresreid:
Is there a pt/eta requirement for "good sim". I find it strange that we are still getting <40% here.

tresreid:
Also it might be good to add the total fake rate

VourMa:
> @VourMa Is there a pt/eta requirement for "good sim". I find it strange that we are still getting <40% here.

Indeed, it is low. There is an |η| < 2.4 cut but no pT cut. Given that we expect a high number of very low-pT tracks in the PU200 sample, I can see how this becomes an issue.
On the other hand, I have also run the MuonGun and MuonGun_highE samples (see other sheets in the GoogleSheet). In the latter, the matching efficiency is 100%, which makes sense, given the above.

VourMa:
> Also it might be good to add the total fake rate

By this, do you mean the addition of cells B12, B18, B24, B30 in the sheet?

tresreid:
> > Also it might be good to add the total fake rate
> 
> By this, do you mean the addition of cells B12, B18, B24, B30 in the sheet?

Yeah, just to have everything easily accessible. so we would have the efficiency, duplicate rate (and breakdown) and fake rate for everything we make all in one place. A summary line.

VourMa:
> > > Also it might be good to add the total fake rate
> > 
> > 
> > By this, do you mean the addition of cells B12, B18, B24, B30 in the sheet?
> 
> Yeah, just to have everything easily accessible. so we would have the efficiency, duplicate rate (and breakdown) and fake rate for everything we make all in one place. A summary line.

Yes, this is going to be useful. I added this to the sheet and to the script.

PR #129, opened 2022-04-08
Add file exists check feature

sgnoohc:
Added a function called check_file_exists where it will check whether the file exists and abort if not.

### DIFF
--- a/code/core/trkCore.h
+++ b/code/core/trkCore.h
@@ -16,15 +16,6 @@
 #include <numeric>
 #include <filesystem>
 
-inline void check_file_exists(const std::string& name)
-{
-    if (not std::filesystem::exists(name.c_str()))
-    {
-        std::cout << "ERROR: Could not find the file = " << name.c_str() << std::endl;
-        abort();
-    }
-}
-
 enum TrackletType {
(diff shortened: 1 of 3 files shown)


code/core/trkCore.h
```diff
@@ -14,6 +14,16 @@
...
 #include <cppitertools/zip.hpp>
 #include <numeric>
+#include <filesystem>
+
+inline void check_file_exists(const std::string& name)
+{
+    if (not std::filesystem::exists(name.c_str()))
+    {
+        std::cout << "ERROR: Could not find the file = " << name.c_str() << std::endl;
+        abort();
```
pwittich, line 24: as we discussed in the meeting, replace `abort()` with `exit(non-zero status)`

sgnoohc:
I upgraded the function to expand to absolute path and check whether the file exists.
And i changed abort to exit(2).

We never used exit status code properly. ... so far it has always been exit(1).
So I thought I might as well start using it now. :)

tresreid:
would be nice to have the paths set directly in the setup and then used everywhere but i can get that in another PR. So i think this is good. The missing file is reported at least.

PR #131, opened 2022-04-18
Removing instances of hard-coded 1 GeV, and removing junk legacy code

bsathian:
Things in this PR :-

1. New header file and source file (Constants.cuh and Constants.cu) which will be used everywhere to deal with hardcoded constants. This has a switch to toggle between 0.8 GeV and 1 GeV (ptCut variable). The ptCut variable is used instead of 1 GeV or 0.8 GeV
2. Removed legacy code (AnalysisInterface, Study.cc etc)


bsathian:
Efficiency plots : https://www.classe.cornell.edu/~bsathian/pr_generalize_pt/efficiencies/eff_plots__GPU_explicit_cache_1a2a419_PU200/mtv/TC_AllTypes__pt_eff.pdf

Comparison with current master :

bsathian:
timing

PR #130, opened 2022-04-11
make md occupancy cut in different eta modules

YonsiG:

PR #133, opened 2022-04-20
Revert "make md occupancy cut in different eta modules"

YonsiG:
Reverts SegmentLinking/TrackLooper#130


YonsiG:
I'm so sorry that again I did something wrong and accidentally merged the pull request....... Can I still revert it now?

slava77:
@tresreid 
was the missing free the only problem in #130?
If so, perhaps it's more straightforward just to have a new PR (soon) with a bugfix, instead of introducing a revert cycle of commits

PR #134, opened 2022-04-20
Bug fix

YonsiG:
for explicit free variable memory and unified version bug fix


slava77:
it would be nice to rebase or just remake the update after checking out the SegmentLinking:master.
This would get rid of 6 other commits in this PR

PR #138, opened 2022-05-02
Host Caching Allocator

GNiendorf:
This pull request replaces most of the CudaMallocHost calls in the code with the host caching allocator that is already present in the codebase.

See the new timing below
full_new

Compared to a run from the current version of the code
current_full

From what I have seen over multiple timing runs, the host allocator gives a little over 10% speedup for s=2,4,6, and 8.

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -94,18 +94,21 @@ SDL::Event::~Event()
     if(trackCandidatesInGPU){trackCandidatesInGPU->freeMemory(stream);}
     if(trackExtensionsInGPU){trackExtensionsInGPU->freeMemory(stream);}
 #endif
-    if(rangesInGPU != nullptr){cudaFreeHost(rangesInGPU);}
-    if(mdsInGPU != nullptr){cudaFreeHost(mdsInGPU);}
-    if(segmentsInGPU!= nullptr){cudaFreeHost(segmentsInGPU);}
-    if(tripletsInGPU!= nullptr){cudaFreeHost(tripletsInGPU);}
-    if(trackCandidatesInGPU!= nullptr){cudaFreeHost(trackCandidatesInGPU);}
-    if(hitsInGPU!= nullptr){cudaFreeHost(hitsInGPU);}
+    if(rangesInGPU != nullptr){cms::cuda::free_host(rangesInGPU);}
+    if(mdsInGPU != nullptr){cms::cuda::free_host(mdsInGPU);}
+    if(segmentsInGPU!= nullptr){cms::cuda::free_host(segmentsInGPU);}
+    if(tripletsInGPU!= nullptr){cms::cuda::free_host(tripletsInGPU);}
+    if(trackCandidatesInGPU!= nullptr){cms::cuda::free_host(trackCandidatesInGPU);}
+    if(hitsInGPU!= nullptr){//cms::cuda::free_host(hitsInGPU);
+        cudaFreeHost(hitsInGPU);}
 
-    if(pixelTripletsInGPU!= nullptr){cudaFreeHost(pixelTripletsInGPU);}
-    if(pixelQuintupletsInGPU!= nullptr){cudaFreeHost(pixelQuintupletsInGPU);}
+    if(pixelTripletsInGPU!= nullptr){//cms::cuda::free_host(pixelTripletsInGPU);
+        cudaFreeHost(pixelTripletsInGPU);}
+    if(pixelQuintupletsInGPU!= nullptr){cms::cuda::free_host(pixelQuintupletsInGPU);}
 
-    if(quintupletsInGPU!= nullptr){cudaFreeHost(quintupletsInGPU);}
-    if(trackExtensionsInGPU != nullptr){cudaFreeHost(trackExtensionsInGPU);}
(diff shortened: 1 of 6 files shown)


tresreid:
image
you are missing 4 Host frees and somewhere have 2 extra frees... Can you please take a look?

tresreid:
image
and for unified version this becomes missing 2 host and 2 regular frees. You can use nvprof to count them

SDL/Event.cu
```diff
@@ -326,25 +329,28 @@ void SDL::Event::resetEvent()
...
     tripletsInGPU = nullptr;}
-      if(quintupletsInGPU){cudaFreeHost(quintupletsInGPU);
+      if(quintupletsInGPU){cms::cuda::free_host(quintupletsInGPU);
       quintupletsInGPU = nullptr;}
-    if(trackCandidatesInGPU){cudaFreeHost(trackCandidatesInGPU);
+    if(trackCandidatesInGPU){cms::cuda::free_host(trackCandidatesInGPU);
     trackCandidatesInGPU = nullptr;}
-    if(pixelTripletsInGPU){cudaFreeHost(pixelTripletsInGPU);
+    if(pixelTripletsInGPU){//cms::cuda::free_host(pixelTripletsInGPU);
+    cudaFreeHost(pixelTripletsInGPU);
```
tresreid, line 348: Is there a reason this isn't using cms::cuda::free_host?

SDL/Event.cu
```diff
@@ -1830,25 +1837,25 @@ void SDL::Event::createPixelTriplets()
...
-    cudaMallocHost(&connectedPixelIndex_host, nInnerSegments* sizeof(unsigned int));
+    connectedPixelSize_host = (unsigned int*)cms::cuda::allocate_host(nInnerSegments* sizeof(unsigned int), stream);
+    connectedPixelIndex_host = (unsigned int*)cms::cuda::allocate_host(nInnerSegments* sizeof(unsigned int), stream);
+
     unsigned int* connectedPixelSize_dev;
     unsigned int* connectedPixelIndex_dev;
-    cudaMalloc(&connectedPixelSize_dev, nInnerSegments* sizeof(unsigned int));
-    cudaMalloc(&connectedPixelIndex_dev, nInnerSegments* sizeof(unsigned int));
-    //unsigned int max_size =0;
+    connectedPixelSize_dev = (unsigned int*)cms::cuda::allocate_host(nInnerSegments* sizeof(unsigned int), stream);
```
tresreid, line 1855: This should not be a host allocation but the normal cudaMalloc. Same with the next line. I think that is the cause of the extra frees you have

tresreid: Yeah, after changing this back, you are only missing 2 more host frees, which makes sense.

SDL/Module.cu
```diff
@@ -526,23 +532,21 @@ void SDL::loadModulesFromFile(struct modules& modulesInGPU, uint16_t& nModules,
...
     cudaStreamSynchronize(stream);
 
-    cudaFreeHost(host_detIds);
-    cudaFreeHost(host_layers);
-    cudaFreeHost(host_rings);
-    cudaFreeHost(host_rods);
-    cudaFreeHost(host_modules);
-    cudaFreeHost(host_subdets);
-    cudaFreeHost(host_sides);
-    cudaFreeHost(host_eta);
```
tresreid, line 536: What happened to the host_eta and host_r frees here? I think these are the missing 2

SDL/Event.cu
GNiendorf: There are 5 variables or so that currently cause errors when using the host allocator. I know why "hitsInGPU" causes an error, but I have not looked too deeply into this variable yet. Since it will take a good chunk of time to go through each of these edge cases and find the best fix for each of them (and they probably won't substantially increase speedup), I think it makes sense to fix them in a later PR.

SDL/Module.cu
GNiendorf: Thank you for pointing this out, just added the fix.

SDL/Event.cu
GNiendorf: Thanks Tres, see the fix I just added.

GNiendorf:
Output from nvprof showing that there are now an equal number of malloc/free's.
fixed_see

The timing benefits are unchanged, here is an example run.
updated_timing

SDL/Event.cu
tresreid: ok makes sense

pwittich: Maybe if you provide a list of the variables that are being difficult @tresreid or @bsathian can give a hint why they are different than the others

SDL/Event.cu
pwittich: Any reason this shouldn’t use the caching allocator (gpu version)?

tresreid:
Validation:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR138_Gavin/eff_plots__GPU_explicit_cache_2901e1c_PU200/

SDL/Event.cu
tresreid: I think you're correct and we should use the gpu caching allocator for this

GNiendorf:
For my own learning purposes (and future pull requests), here is also a link to the validation plots:
https://www.classe.cornell.edu/~gsn27/www/PR138/

PR #139, opened 2022-05-03
2D duplicate kernels + Improved segment launching grid 

bsathian:
temp_timing

For segments, going to one extreme (thread dimensions 1024, 1, 1) seems to help with memory coalescing, given that our cache hit rate is now 95% (from 67% previously). In addition, I've implemented Matt's suggestions albeit for pT3 and pT5 duplicate cleaning kernels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ movetoweb.sh
 *.log
 
 *.nfs*
+.directoryhash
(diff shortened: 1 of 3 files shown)


tresreid:
image
8 stream timing with the master pull. which looks slightly better

bsathian:
https://www.classe.cornell.edu/~bsathian/dup_cleaning_2d_pr/efficiencies/eff_plots__GPU_explicit_cache_5709042_PU200/

PR #140, opened 2022-05-03
Move  create segment kernel

tresreid:
moves the createSegmentsInGPU kernel to Segment.cu. This reduces the maximum live register usage for the kernel to under 64 registers.


tresreid:
validation: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR140/eff_plots__GPU_explicit_cache_49d7479_PU200/

tresreid:
timing:
image

PR #141, opened 2022-05-04
Register reduction

tresreid:
Moves Kernel calls and used functions to all be within the same file.
Merges PixelQuintuplet file into PixelTriplet fit. 
all Kernels now use < 64 registers.
line 650 in SDL/PixelTriplets.cu does cause a change in the multiplicity and it is unclear as to why. Changing this back to the original computeChiSquared( ...) fixes this at the cost of using more registers for pT5 and pT3 and a bit of timing (not really that noticeable though).


tresreid:
Validation (with full register reduction): https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR141/eff_plots__GPU_explicit_cache_fcf3752_PU200/#1.5
Validation (with better efficiency): 
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR141v2/eff_plots__GPU_explicit_cache_fcf3752DIRTY_PU200/#1.5

tresreid:
image

PR #143, opened 2022-05-08
Fix for maximum number of objects

VourMa:
This PR properly applies the maximum number of objects, as tested by printouts.

**Timing**
Master:
timing_master

This PR:
timing_fixObjectMaxNumber

**Validation**
Master:
https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/fixObjectMaxNumber/eff_plots__GPU_explicit_cache_1cad1b0_PU200/

This PR:
https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/fixObjectMaxNumber/eff_plots__GPU_explicit_cache_84a74b2_PU200_fixObjectMaxNumber/

The efficiency is slightly lower (<1%) for this PR, which makes sense since we are now properly cropping the number of objects.

### DIFF
--- a/SDL/Kernels.cu
+++ b/SDL/Kernels.cu
@@ -759,8 +759,8 @@ __global__ void createT3T3ExtendedTracksInGPU(struct SDL::modules& modulesInGPU,
 
         if(success and nLayerOverlaps[0] == 2)
         {
-            atomicAdd(&trackExtensionsInGPU.totOccupancyTrackExtensions[nTrackCandidates], 1);
-            if(trackExtensionsInGPU.nTrackExtensions[nTrackCandidates] >= N_MAX_T3T3_TRACK_EXTENSIONS)
+            unsigned int totOccupancyTrackExtensions = atomicAdd(&trackExtensionsInGPU.totOccupancyTrackExtensions[nTrackCandidates], 1);
+            if(totOccupancyTrackExtensions >= N_MAX_T3T3_TRACK_EXTENSIONS)
             {
 #ifdef Warnings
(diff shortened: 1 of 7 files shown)


tresreid:
This is what I'm seeing from the multiplicies in the first event. Other events I think look ok though. Could you run this and double check please?
image

VourMa:
I confirm the issue, which I tracked down to the triplets. I think it's unrelated to this PR. It seems that we are missing some `cudaMemsetAsync` [here](https://github.com/SegmentLinking/TrackLooper/blob/0a31961831cfe6c09c30cbed8e5828b0b246922c/SDL/Triplet.cu#L144), including the one for `totOccupancyTriplets`?

Can you please confirm?

tresreid:
Yes that seems to fix it. Can you push that change to this PR? And also check to make sure this doesn't happen in other files. Lastly, in the unified section, there are memcpys that are within an ifdef for track extensions. Can you please remove the ifdef lines and keep the memcpys

VourMa:
> This is what I'm seeing from the multiplicies in the first event. Other events I think look ok though. Could you run this and double check please? image

The issue mentioned above has been fixed, as it can be seen in the screenshot below:
Screenshot 2022-05-10 at 8 28 33 PM

I am attaching the updated timing (no real change):
Screenshot 2022-05-10 at 8 48 54 PM

The validation plots have also been updated (no real change as well). After inspecting the plots, it seems that efficiency differences are observed for all of the objects (as expected) and the efficiency can both decrease or increase slightly (due to the undefined behavior of overwriting).

PR #144, opened 2022-05-10
Max blocks check yonsi

YonsiG:
This is after changing block size to 13296 for SG, T3, and also take care of T5, make every module have a thread (in sg and T3) and make every triplet has a thread in T5


YonsiG:
image

YonsiG:
http://uaf-10.t2.ucsd.edu/~yonsi/SDL_GPU_plots/blocksize_change/eff_comparison_plots__0a31961_explicit_cache_d91ae79_explicit_cache_PU200/

tresreid:
I think this is not obsolete due to all the other PRs

PR #145, opened 2022-05-10
Shared memory segments

bsathian:
A v1 implementation of shared memory for the segment kernel. This has barely matched up to the existing implementation in terms of timing
shared_mem_timing_v2


bsathian:
The two profiler reports : One with shared memory, and the "reference" one without shared memory can be found here :-

https://www.classe.cornell.edu/~bsathian/sharedMemory_profilers/

SDL/Segment.cu
```diff
@@ -918,27 +922,68 @@ __global__ void SDL::createSegmentsInGPUv2(struct SDL::modules& modulesInGPU, st
     int blockxSize = blockDim.x*gridDim.x;
     int blockySize = blockDim.y*gridDim.y;
     int blockzSize = blockDim.z*gridDim.z;
+
+    __shared__ short subdets[MAX_CONNECTED_MODULES], layers[MAX_CONNECTED_MODULES], sides[MAX_CONNECTED_MODULES], rods[MAX_CONNECTED_MODULES], rings[MAX_CONNECTED_MODULES], innerSubdet, innerLayer, innerSide, innerRod, innerRing;
```
slava77, line 926: would performance get any better if full words are used here (32 bit, as `unsigned int`)?
My reading of https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#shared-memory-5-x suggests that the bank conflicts appear only for separate reads inside the bank.

SDL/Segment.cu
```diff
@@ -918,27 +922,68 @@ __global__ void SDL::createSegmentsInGPUv2(struct SDL::modules& modulesInGPU, st
...
+
     unsigned int nConnectedModules = modulesInGPU.nConnectedModules[innerLowerModuleIndex];
+    if(nConnectedModules == 0) continue;
+    //shared memory stuff
+    if(threadIdx.x == 0 and threadIdx.z == 0)
+    {
+        for(unsigned int i = threadIdx.y; i < nConnectedModules; i+=blockDim.y)
+        {
+            subdets[i] = modulesInGPU.subdets[modulesInGPU.moduleMap[innerLowerModuleIndex * MAX_CONNECTED_MODULES + i]];
+            layers[i] = modulesInGPU.layers[modulesInGPU.moduleMap[innerLowerModuleIndex * MAX_CONNECTED_MODULES + i]];
```
slava77, line 944: just in case the compiler is not optimizing away the inner index computation, does it get any faster with a temp
```suggestion
            auto moduleIdx = modulesInGPU.moduleMap[innerLowerModuleIndex * MAX_CONNECTED_MODULES + i];
            subdets[i] = modulesInGPU.subdets[moduleIdx];
            layers[i] = modulesInGPU.layers[moduleIdx];
```
... (similar to the remaining lines)

PR #151, opened 2022-05-10
pLS 2D Kernel

bsathian:
Converts the checkHitspLS kernel to a 2D one without any change in multiplicity or efficiency.

Timing
pLS_2D


tresreid:
This does remove the second pass of duplicate removal. but that seems to be fine

PR #152, opened 2022-05-11
resize pixelTriplet kernels

slava77:
for `SDL::createPixelTripletsInGPUFromMapv2` (on V100; 20 events) time reduced from  3.04ms to 1.23 ms by changing to  nThreads(32,4,1); nBlocks(1,4096,1);
- Here the outer y loop typically goes over 0.2 - to 0.5 million entries;
the innermost loop varies : 83% is below 16 (not counting zeros); around 8% goes above 32; 1% is above 128. So, there will still be some tails of lonely threads.
- Still, a factor of 2.5 speedup is almost for free
- I went up to 16384 blocks in y and down to only 32-x in block size; these had worse times

removeDupPixelTripletsInGPUFromMap has a bit odd behavior; it's a loop of N*N, where N is around 300 to 800. It writes to bool isDup[N] to the outer index.
The cost of computation inside the kernel thread seems small.
I find that the smallest time is achieved with nBlocks smaller than number of SMs.
nBlocks(1,40); nThreads(32,32) runs faster than nBlocks(1,80); nThreads(32,32): time changes from 0.117ms to 0.085ms.
- times go up by a factor of 2 with more blocks
- I'm guessing there is something with memory coherency in cache that could slow things down: either the writes happen to the same cache line but different location, or perhaps even the same location. 
We are writing only "1" in this case and it wouldn't matter which thread value makes it. Is there a way to tell it to the compiler or GPU runtime?

Since my setup is on phi3, I think that it would be better to get reference timing values from lnx/Cornell.

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1859,10 +1859,11 @@ cudaStreamSynchronize(stream);
     free(segs_pix);
     //cudaFreeAsync(segs_pix_gpu,stream);
     cudaFree(segs_pix_gpu);
+
 #ifdef Warnings
     unsigned int nPixelTriplets;
-    cudaMemcpyAsync(&nPixelTriplets, &(pixelTripletsInGPU->nPixelTriplets),  sizeof(unsigned int), cudaMemcpyDeviceToHost,stream);
-    cudaStreamSynhronize(stream);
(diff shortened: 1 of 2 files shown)


tresreid:
validation including PR #143:  https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR152/eff_plots__GPU_explicit_cache_516229d_PU200/

PR #153, opened 2022-05-11
change thread extension thread balance

tresreid:
quick change to the number of blocks/ threads in track extension creation. shaves off 1ms


tresreid:
image

PR #154, opened 2022-05-11
speedup in startup and hit loading

slava77:
for 20 events one stream
Real Time (as of 3dbc4c3) :      23.9 s->  13.1 s.

From igprof report https://slava77sk.web.cern.ch/slava77sk/reco/cgi-bin/igprof-navigator/sdl/ig.pp.3dbc4c3.rerun2 vs https://slava77sk.web.cern.ch/slava77sk/reco/cgi-bin/igprof-navigator/sdl/ig.pp.16776d3-addInputs2.m8clx (in perf counts, which I think are about the same as seconds)
- `addInputsToLineSegmentTrackingPreLoad`: 2.31 -> 1.95
- `SDL::initModules`: 3.46 -> 2.07
- `SDL::initHits`: 10.27 -> 1.43

PR #156, opened 2022-05-12
go back to regular std::map to resolve regression in removeDupQuintupletsInGPUv2 timing

slava77:
as reported in https://github.com/SegmentLinking/TrackLooper/pull/154#issuecomment-1125030693
it looks like the `removeDupQuintupletsInGPUv2` kernel slowed down by almost a factor of 2.

After printing out the modules appearing in this kernel calls, it looks like the order of modules has changed. For the original map the order of modules should follow the order of detIds. For the `unordered_map` the order can be somewhat random. It seems like this dup cleaning kernel is more effective if the modules are ordered somewhat in the direction of the detIds.

(perhaps we can find an even better order for modules at some point to make the duplicate cleaning even better)

I reverted back to using std::map. This will slow down the pixel map loading by about a factor of 2, but it's manageable.

### DIFF
--- a/SDL/Module.cu
+++ b/SDL/Module.cu
@@ -1,10 +1,10 @@
 #include "Module.cuh"
 #include "ModuleConnectionMap.h"
 #include "allocate.h"
-std::unordered_map <unsigned int, uint16_t> *SDL::detIdToIndex;
-std::unordered_map <unsigned int, float> *SDL::module_x;
-std::unordered_map <unsigned int, float> *SDL::module_y;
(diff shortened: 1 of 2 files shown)


tresreid:
image

tresreid:
image
without the additional recent PRs

slava77:
with the unordered map most filled modules with T5s end up in the range from around 4K to 13K, while with std::map they are from 2K to 7K.

tresreid:
image
timing with everything included

PR #158, opened 2022-05-12
Second pass T5 duplicate removal

tresreid:
rearranges loop order. Changes loops over lower modules to loops over eligible T5 modules (results in more T5s i think). 
I found that only 1 thread in the inner most loops actually worked the best so I took off the parallelism there and moved it to the outer most loops with the modules. This seems to have given the best result.


tresreid:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR158/eff_plots__GPU_explicit_cache_bd99a1c_PU200/
There is a higher efficiency and duplicate rate here. But the duplicate rate is still fairly low I think.

tresreid:
image

tresreid:
image

PR #159, opened 2022-05-12
speedups in pT3 

slava77:
- "massage" the implementation in `createPixelTripletsInGPUFromMapv2` and called methods to reduce the number of registers: 
    - the number of used registers as in ptxas info reduced from 138 to 124. 
        - At some point rather similar code had 118 registers, but I lost track of that version; it was not significantly faster 
    - pT3 time changed from 5.5 to 5.2 ms
    - `createPixelTripletsInGPUFromMapv2` reduced from 1.25 ms to 1.02 ms
- remove (pLS X connectedModules) remapping/flattening done on CPU and simply use another dimension in the grid to do it on the fly; this saves the cost of mem copy of about 8MB of data
    - this provides the largest improvement in timing: pT3 time (on phi3) goes from 5.2 ms to 1.7 ms (compared from 8b578a3b385 to dacd609f8)

The validation was limited: looking at just the number of produced pT3s and other objects

### DIFF
--- a/SDL/Constants.cu
+++ b/SDL/Constants.cu
@@ -5,23 +5,24 @@
 #include "Constants.cuh"
 
 //defining the constant host device variables right up here
-CUDA_CONST_VAR float SDL::miniMulsPtScaleBarrel[6] = {0.0052, 0.0038, 0.0034, 0.0034, 0.0032, 0.0034};
-CUDA_CONST_VAR float SDL::miniMulsPtScaleEndcap[5] = {0.006, 0.006, 0.006, 0.006, 0.006}; 
+CUDA_CONST_VAR const float SDL::miniMulsPtScaleBarrel[6] = {0.0052, 0.0038, 0.0034, 0.0034, 0.0032, 0.0034};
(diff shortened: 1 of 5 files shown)


slava77:
time for 20 evts on phi3: (ef72c6c is old, dacd609 is new)
```
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short             Rate
   old      3.0      4.0      3.5      3.5      3.7      5.0      6.4      5.5      6.9      5.8      47.1      39.1+/-  6.6      52.2   explicit_cache[s=1]
   new      2.8      3.8      3.4      3.4      3.7      5.0      6.5      1.7      6.3      5.6      42.2      34.5+/-  5.8      45.8   explicit_cache[s=1]
```

slava77:
my timing diff apparently includes the impact of #155 
So, only the pT3 column change is relevant.

tresreid:
validation looks good: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR159/eff_plots__GPU_explicit_cache_dacd609_PU200/#1.5

tresreid:
image

PR #162, opened 2022-05-14
keep track of pLS hit idXs in a vector of uint4

slava77:
… and use it directly in checkHitspLS for overlap checks

```
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short             Rate
   old      2.8      3.8      3.4      3.4      3.8      5.0      6.7      1.7      4.2      5.7      40.5      32.8+/-  5.6      44.2   explicit_cache[s=1]
   new      2.7      3.9      3.4      3.4      3.7      2.6      6.5      1.8      3.7      5.8      37.4      32.1+/-  5.3      40.9   explicit_cache[s=1]
```
old is 4226916; new is 86ec271

the impact is on pLS and TC calls, deriving from `checkHitspLS` speedup.

The average time in `checkHitspLS` is down from 3.10ms -> 1.65ms (it is called twice per event); the gain per event is then around 2.9 ms

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1046,7 +1046,7 @@ __global__ void addPixelSegmentToEventKernel(unsigned int* hitIndices0,unsigned 
       addMDToMemory(mdsInGPU, hitsInGPU, modulesInGPU, hitIndices2[tid], hitIndices3[tid], pixelModuleIndex, 0,0,0,0,0,0,0,0,0,outerMDIndex);
 #endif
 
-    int hits1[4];
+    unsigned int hits1[4];
     hits1[0] = hitsInGPU.idxs[mdsInGPU.anchorHitIndices[innerMDIndex]];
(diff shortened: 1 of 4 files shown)


slava77:
I checked with `-v 2` that the multiplicity doesn't change

slava77:
also, `checkHitspLS` is now using just 32 registers

tresreid:
https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR162/eff_plots__GPU_explicit_cache_86ec271_PU200/#1.5

tresreid:
image

sgnoohc:
Out of personal curiosity...
Are these timing result based on A100? or V100?

slava77:
> Out of personal curiosity... Are these timing result based on A100? or V100?

v100

PR #163, opened 2022-05-14
add early skips in checkHitspLS

slava77:
```
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short             Rate
   old      2.7      3.9      3.4      3.4      3.7      2.6      6.5      1.8      3.7      5.8      37.4      32.1+/-  5.3      40.9   explicit_cache[s=1]
   new      2.8      4.0      3.6      3.5      3.7      1.3      5.8      1.8      3.6      5.8      36.0      31.9+/-  5.5      39.8   explicit_cache[s=1]
```
old is 21cf443 new is a5876c4

checkHitspLS average time changes from 1.65ms to 0.96ms (recall that it's average out of 2 calls), so, the total should be down by 1.4ms, as observed:
- the first commit takes it down to 0.99ms
- the second commit to 0.96ms

admittedly, most of the gain is in pLS, which we sort of expect to probably not need a cleaning pass if patatracks come cleaned already

PR #164, opened 2022-05-19
pLS new score

bsathian:
Fixes pLS score computation, but does not use it while adding pLS in TC collection

https://www.classe.cornell.edu/~bsathian/pLS_PR_new/efficiencies/eff_plots__GPU_explicit_cache_41dd58fDIRTY_PU200/mtv/TCE_AllTypes_h_duplrate_numer_pt_eff.pdf


bsathian:
pLS_timing

bsathian:
Comparison plots: https://www.classe.cornell.edu/~bsathian/pLS_PR_new/efficiencies/eff_comparison_plots__41dd58fDIRTY_explicit_cache_72b6266_explicit_cache_PU200/TCE_AllTypes__etacoarse.png

slava77:
> Comparison plots: https://www.classe.cornell.edu/~bsathian/pLS_PR_new/efficiencies/eff_comparison_plots__41dd58fDIRTY_explicit_cache_72b6266_explicit_cache_PU200/TCE_AllTypes__pt.pdf

do we have an html page setup to display these comparison plots?
As we have seen previously from pre-PR slides the fake rate goes up. This can be an acceptable regression considering that it's a bug fix, but more work may be needed to make tighter selections.
It would be more clear after the categorization plot is made with this PR.



For the timing plot here
> pLS_timing

vs in #163:  in #163

it's nice to see little impact on the total or even individual times

slava77:
> This can be an acceptable regression considering that it's a bug fix, but more work may be needed to make tighter selections.

I implied that this selection review would happen in a follow up PR(s)

PR #170, opened 2022-05-23
Module type for CMSSW_12_2_0_pre2 geometry

VourMa:
This PR adds the module (layer) type information in the centroid data file and gets that information from there, instead of parsing the `detId`. No further changes are applied downstream.

Only the CMSSW_12_2_0_pre2 geometry is affected by these changes.

**Validation plots**
Current master (726ef8c):
- [Old geometry](https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/moduleType/eff_plots__GPU_explicit_cache_726ef8c_PU200_oldGeom/)
- [New geometry](https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/moduleType/eff_plots__GPU_explicit_cache_726ef8c_PU200_newGeom/)

This PR (f0f5b32):
- [Old geometry](https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/moduleType/eff_plots__GPU_explicit_cache_f0f5b32_PU200_oldGeom/)
- [New geometry](https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/moduleType/eff_plots__GPU_explicit_cache_f0f5b32_PU200_newGeom/)

PR #171, opened 2022-05-29
Host Allocator Cleanup

GNiendorf:
This PR removes a few variables from the host caching allocator that were being freed by cudaFreeHost(). These variables all throw errors if you try to free them with the host caching allocator. Since all of these specific variables are only allocated/freed once, this should not affect the timing, but it's not clear how cudaFreeHost() is handling these cases (and so they should be removed).

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -808,7 +808,8 @@ std::vector<std::vector<short>>&    out_isQuad_vec
     for(int evt=0; evt < static_cast<int>(out_trkX.size()); evt++)
     {
         struct SDL::hits* hitsInGPU_event;
-        hitsInGPU_event = (struct SDL::hits*)cms::cuda::allocate_host(sizeof(struct SDL::hits), modStream);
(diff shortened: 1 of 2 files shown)


GNiendorf:
Current timing below
timing_fix_current
Timing with these changes
timing_fix

There are no substantial timing differences as shown above.

VourMa:
Does this PR build on top of #170, which also applies changes for the host caching allocator to fix crashes? Is it based on the recent discussions on Skype or is it independent of them?

GNiendorf:
> Does this PR build on top of #170, which also applies changes for the host caching allocator to fix crashes? Is it based on the recent discussions on Skype or is it independent of them?

This is independent of those changes. The changes made in this PR just clean up certain variables which were being allocated with cms::cuda::allocate_host() but then freed with cudaFreeHost() because they were causing errors if you tried to free them with cms::cuda::free_host().

GNiendorf:
Performance plots
https://www.classe.cornell.edu/~gsn27/www/PR171/

PR #181, opened 2022-06-29
Hit Loading Speedup

GNiendorf:
This PR speeds up hit loading by moving calculations done in addHitToEvent to the GPU. I decided to not make changes to the hit preloading code until we decide if we want to keep it in the codebase. I did however move the timing/validation scripts back over to not using the preloading flag by default.

Validation plots:
https://www.classe.cornell.edu/~gsn27/www/PR181/

Timing with no preloading:
PRTiming

Also, the "Real Time" for 200 events drops from 42s (w/ preloading) to 30s (w/o preloading) under this PR for single stream.


pwittich:
When @tresreid comes back we can decide if this should be ported to the hit loading or if we should get rid of it.

tresreid:
I think it makes sense to drop the hit preloading. I think this is a better version overall. Given the timing results, the preloading isn't necessary anymore

GNiendorf:
Validation plots
https://www.classe.cornell.edu/~gsn27/www/PR181_4/

Timing
removepre

I will upload comparison validation plots to the current master soon.

GNiendorf:
Comparison plots:
https://www.classe.cornell.edu/~gsn27/www/PR181_comp/

TC_AllTypes__ptzoom
TC_AllTypes__pt

GNiendorf:
I checked all of my variables on the GPU against their CPU-calculated values and couldn't find any major outliers. I think what's going on is that when I calculate a certain variable on the GPU, like sqrt(hit_x*hit_x+hit_y*hit_y), it differs slightly from its CPU version because CUDA's math functions are calculated differently than the standard library's math functions. I also noticed that one variable which is calculated on both the CPU and GPU with the same _global_ function agrees much better than the other variables (up to the last decimal compared to only 4-5 decimal places). To check that this is causing the efficiency differences, I made a hybrid CPU/GPU version of the code where I first calculate the variables on the CPU and upload them to the GPU like what is done on the master branch, but then override them with my GPU kernels. If I comment out only the variables that differ by math precision (but override everything else), then I get much better agreement to master.

https://www.classe.cornell.edu/~gsn27/www/PR181_comp2/

TC_AllTypes__ptzoom2

slava77:
> Comparison plots: https://www.classe.cornell.edu/~gsn27/www/PR181_comp/
> 
> The differences are small, but I will try to see where they are coming from.
> 
> TC_AllTypes__ptzoom

Does this comparison use all of the events available? If the differences are from a random shuffle, it would be best to compare with the smallest uncertainty (largest sample).

SDL/Event.cu
```diff
@@ -574,464 +566,151 @@ void SDL::Event::resetObjectsInModule()
...
-    cms::cuda::free_host(module_moduleType);
-    cms::cuda::free_host(host_x);
-    cms::cuda::free_host(host_y);
-    cms::cuda::free_host(host_z);
-    cms::cuda::free_host(host_detId);
-    cms::cuda::free_host(host_idxs);
-
+    // No stream synchronize needed after second kernel call. Saves ~100 us.
+    moduleRangesKernel<<<MAX_BLOCKS,256,0,stream>>>(nLowerModules, modulesInGPU, hitsInGPU);
+    //std::cout << cudaGetLastError() << std::endl;
```
slava77, line 709: Is the sync not needed because the next operation is `addPixelSegmentToEvent`), which does not rely on `hitsInGPU.hitRanges*`  (the only values modified in this kernel)?

It would be better to be clear about this, to reduce a chance of errors in the future

GNiendorf:
I removed the extra semicolon and explained why a sync is not needed after the second kernel. I also moved hitsInGPU to the host caching allocator, since previously the hit preloading code was causing an error that prevented it from being moved.

Validation plots with all events available:
https://www.classe.cornell.edu/~gsn27/www/PR181_postclean1/

Current master with all events available:
https://www.classe.cornell.edu/~gsn27/www/PR181_masterfull/

Comparison with all events available:
https://www.classe.cornell.edu/~gsn27/www/PR181_compfull/

New Timing:
t1

zoom1

GNiendorf:
Peter pointed out that these should be cosf and sinf since the inputs are floats. I tried sincosf but that gave random runtime errors. Doesn't affect the timing substantially, but it's still good to move them over to the correct functions.

SDL/Event.cu
```diff
@@ -574,464 +564,154 @@ void SDL::Event::resetObjectsInModule()
...
+            float phi = 0;
+            // Unclear why these are not in map, but CPU map returns phi = 0 for all exceptions.
+            if (found_index != -1)
+                phi = geoMapPhi[found_index];
+            float cos_phi = cosf(phi);
+            hitsInGPU->highEdgeXs[ihit] = ihit_x + 2.5 * cos_phi;
+            hitsInGPU->lowEdgeXs[ihit] = ihit_x - 2.5 * cos_phi;
+            float sin_phi = sinf(phi);
+            hitsInGPU->highEdgeYs[ihit] = ihit_y + 2.5 * sin_phi;
+            hitsInGPU->lowEdgeYs[ihit] = ihit_y - 2.5 * sin_phi;
```
pwittich, line 644: All these float should be put in `2.5f` to avoid the conversion to doubles, otherwise switching to `cosf` and `sinf` probably doesn’t matter.

slava77:
> Hey @slava77 can you check if my changes address all the concerns you had when you get a chance? Thank you!

yes, my comments were resolved. Thanks.

PR #183, opened 2022-07-08
Host and Device Allocator Fix

GNiendorf:
This PR removes an extra stream (called modStream) by moving code to the default stream (stream=0). I also specified the stream for a few memsets to avoid conflicts with other streams. This seems to have cleared up previous issues with the host and device allocators. I then moved over variables to the allocators that were previously causing errors. I will upload validation plots soon.

New Timing
fix_time

Current Master
master

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -526,16 +526,13 @@ void SDL::initModules(const char* moduleMetaDataFilePath)
 {
     if(modulesInGPU == nullptr)
     {
-    cudaStream_t modStream;
-    cudaStreamCreate(&modStream);
-        //modulesInGPU = (SDL::modules*)cms::cuda::allocate_host(sizeof(struct SDL::modules), modStream);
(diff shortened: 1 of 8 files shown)


GNiendorf:
Validation plots: https://www.classe.cornell.edu/~gsn27/www/PR183_valfull/#1.6

Comparison to master: [Here](https://www.classe.cornell.edu/~gsn27/www/PR183_compfull/eff_comparison_plots__daa3a63DIRTY_explicit_cache_9c806acDIRTY_explicit_cache_PU200/)

tresreid:
do you understand why this seems to be faster (particularly pT5s)?

GNiendorf:
> do you understand why this seems to be faster (particularly pT5s)? This doesn't look like the typical variation in the timing. I haven't seen it that fast in a while.

It's because I moved over variables to the allocators that were previously causing errors. I don't understand exactly why the scaling improves though.

PR #185, opened 2022-07-24
Explicit cut value ntuple

bsathian:
Given that unified is pretty much dead, and the cut value debug ntuple part of the code was not working. This PR creates appropriate memory in the explicit framework. 

compiling with options `-mxc8d` adds the extra branches in the output ntuples


tresreid:
image
with the debug flag, there are a few frees missing. the normal non-debug instance is ok.

bsathian:
https://www.classe.cornell.edu/~bsathian/cut_value_debug_plots/efficiencies/eff_plots__GPU_explicit_cache_cutvalue_3086853DIRTY_PU200/

bsathian:
timing

PR #186, opened 2022-07-29
Fix for trkAlgo iterations

VourMa:
This PR is a fix on removing the extra tracking algorithm iterations in one more place in the code. Prior to this PR, the issue of higher fake rate seems to be evident even in the CMSSW 11 geometry (somehow this was missed in previous PRs?). I am not sure if a recent change caused that. However, this PR fixes the issue in any geometry, as can be seen in the validation plots (I hope that the labels are self-explanatory):
https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/trkIterAlgoRevisited/


tresreid:
timing for cmssw 11:
image

tresreid:
image
CMSSW 2

PR #190, opened 2022-08-29
Make CMSSW 12 geometry default + Cleanup

VourMa:
This PR makes the CMSSW 12 geometry the default and removes the compilation flag. Similarly, the pT>0.8 GeV configuration becomes the default and the relevant compilation flag is removed. Some cleanup is also done.

Comparison plots (names+git tags should be self-explanatory):
https://uaf-10.t2.ucsd.edu/~evourlio/SDL/moveToCMSSW12Geometry/

Timing before all changes (phi3):
```
Total Timing Summary
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short           Rate
   avg      6.3      3.8      5.0      4.4      3.8      1.4      2.5      2.1      4.9      6.5      40.7      33.0+/-  4.6      42.8   explicit_cache[s=1]
   avg     11.1      5.1      4.6      6.8      6.7      1.9      4.6      3.4      8.0      8.0      60.2      47.2+/-  7.5      32.7   explicit_cache[s=2]
   avg     21.1      7.9      6.8     12.6     13.5      3.3     10.2      6.5     16.9     12.3     111.1      86.6+/- 14.1      29.9   explicit_cache[s=4]
   avg     31.2     10.7      9.9     18.0     20.2      4.5     16.6      9.3     24.2     16.1     160.7     125.0+/- 17.5      28.3   explicit_cache[s=6]
   avg     49.9     12.3     11.4     25.1     27.7      6.0     24.4     12.5     31.1     19.2     219.7     163.7+/- 26.9      29.1   explicit_cache[s=8]
```

Timing after all changes (phi3):
```
Total Timing Summary
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short           Rate
   avg      6.4      4.1      4.2      4.6      3.9      1.3      2.2      1.9      4.4      6.4      39.3      31.6+/-  5.8      41.2   explicit_cache[s=1]
   avg     10.8      5.7      4.6      7.1      6.6      1.7      4.3      3.1      7.5      8.0      59.3      46.8+/-  8.5      32.2   explicit_cache[s=2]
   avg     21.5      7.8      6.3     12.3     13.3      3.3      9.5      5.9     15.3     11.0     106.3      81.5+/- 12.5      28.2   explicit_cache[s=4]
   avg     33.9     10.9      8.9     20.2     21.2      4.5     16.5     10.1     23.2     14.6     163.9     125.5+/- 19.2      28.9   explicit_cache[s=6]
   avg     49.2     12.5     11.3     25.5     27.1      5.3     20.4     12.1     32.6     20.1     216.0     161.5+/- 25.7      28.4   explicit_cache[s=8]
```


slava77:
How many events do we have with the 12X? Is it the same 500 as in 104X/old?

tresreid:
This is the timing I got on lnx7188. Closer to the previous times for cmssw 11:
image

tresreid:
This uses the same input file as before correct? If so, there is still a bit of an efficiency drop compared to old. Otherwise, This passes the usual checks.

tresreid:
The stacked plots seem to be missing a plot though.

VourMa:
> This uses the same input file as before correct? If so, there is still a bit of an efficiency drop compared to old. Otherwise, This passes the usual checks.

When you say "old", do you refer to the configuration before removing things? If yes, then the input is indeed the same and same amount of events. This PR is purely technical, so I don't see where variation may come from, except from the order things run on the GPU. Do you think there's another source?

> The stacked plots seem to be missing a plot though. The efficiency verse pt

You are right, I didn't notice that. I will rerun the plots to fix this.

VourMa:
> The stacked plots seem to be missing a plot though. The efficiency verse pt

Turns out this is a long standing typo in the html file. I am fixing it with an additional commit to this PR. Thanks for catching it.

slava77:
> If you are talking about ttbar, then no, the statistics is not the same. The 12X sample has 175 events. If we want more, I had volunteered to produce some but, if I recall correctly, Philip suggested that this production may be a good exercise for Matt. I don't know how it evolved from there but we can revisit this.

Doesn't Gavin have more data already available? (was it 6K events?)

tresreid:
> 

Well my concern isn't so much the this PR's cleanup vs the previous master but rather the difference between the cmssw 12 results vs cmssw 11. There is, and has been, a decrease in the efficiency in using cmssw 12 which i thought was still being understood and addressed. Going from 90% efficiency to 85%. I'm a bit concerned about making it the default now when this drop in efficiency remains. We can't really say that we have comparable efficiency to current patatrack+CFK.

VourMa:
> Doesn't Gavin have more data already available? (was it 6K events?)

I think he has but I don't know the details of where this is stored. We can ask and make it available to all.

VourMa:
> Well my concern isn't so much the this PR's cleanup vs the previous master but rather the difference between the cmssw 12 results vs cmssw 11. There is, and has been, a decrease in the efficiency in using cmssw 12 which i thought was still being understood and addressed. Going from 90% efficiency to 85%. I'm a bit concerned about making it the default now when this drop in efficiency remains. We can't really say that we have comparable efficiency to current patatrack+CFK.

The difference has been there since a long time with no obvious way to fix it, at least to my knowledge.

In any case, we got "complaints" by the TRK people for using such an old geometry for the tracker when using it for comparisons, so I don't think that sticking with the old geometry makes sense, just for saying that we are comparable with the current patatrack+CKF. But let's talk about it at the meeting if there are other proposals.

tresreid:
I know this difference has always existed, but I was under the impression that it was being addressed and cmssw 12 would become the default once it was. The issue with this becoming the default now is also that we lose the cmssw 11 geometry option. Which means it would be harder to go back to make comparisons. To me it seems like we would be accepting this drop in efficiency and I don't think that is what we should do.

tresreid:
i mean, maybe we can regain some efficiency by returning the cuts since they were all set for cmssw 11 geo, but i think the differences went down to which hits we were seeing right?

tresreid:
Also, just looking at the cmssw 12 efficiency plots alone, it seems clear there is some kind of issue..., the turn on has a noticeable drop around 1 GeV that seems out of place. The plateau starts to drop for high pt tracks (>10 GeV). These plots just don't look correct at the moment. I think this should be addressed before we make this the default that we share outside of our group....

GNiendorf:
> > Doesn't Gavin have more data already available? (was it 6K events?)
> 
> I think he has but I don't know the details of where this is stored. We can ask and make it available to all.

The 6k event sample is located here:

/mnt/data1/gsn27/samples/CMSSW_12_5_0_pre3/src/ttbar_CMS12_9000

slava77:
Can we aim to conclude on merging this PR some time in the Tuesday meeting or soon after in this first week of October?

VourMa:
For documentation and for reference, here are the validation plots after the PR was merged:

https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/afterMovingToCMSSW12Geometry/eff_plots__GPU_explicit_cache_236179b_PU200_n200/

And here is the timing on lnx7188 (close to the timings of CMSSW 10 geometry):
```
Total Timing Summary
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short           Rate
   avg      4.8      3.4      3.6      3.5      3.1      1.1      1.8      1.6      4.0      5.4      32.5      26.5+/-  4.1      35.0   explicit_cache[s=1]
   avg      7.3      5.1      4.7      6.1      5.5      1.7      3.7      2.6      6.7      6.6      50.0      40.9+/-  8.5      31.0   explicit_cache[s=2]
   avg     16.4      7.4      6.5     11.2     10.8      2.8      8.2      5.4     13.9      9.9      92.5      73.4+/- 10.7      25.1   explicit_cache[s=4]
   avg     25.5     10.1      8.2     17.2     17.1      3.7     13.6      8.5     20.2     12.7     136.9     107.7+/- 17.8      24.6   explicit_cache[s=6]
   avg     36.1     11.6     10.5     21.3     20.6      4.8     19.2     11.3     26.0     16.6     178.0     137.1+/- 21.4      24.1   explicit_cache[s=8]
```

An extensive set of comparisons is linked in (the now closed) issue #191.

Issue #188, opened 2022-08-17
Code Crashes with Multiple Input Files


GNiendorf:
Steps to reproduce:
sdl_make_tracklooper -m8c2
./bin/sdl -n 1100 -v 0 -s 8 -o test.root --input /mnt/data1/gsn27/samples/CMSSW_12_5_0_pre3/src/ttbar_CMS12_9000/

Extra info: This bug does not occur if run with "-w 0"

![output](https://user-images.githubusercontent.com/25272611/185182008-1b711020-4895-46a5-b089-740b362c36bd.png)

slava77:
the crash is apparently on the host side.
Can you run this under gdb and then get a stack trace? (type `where`, after it aborts)

GNiendorf:
Stack trace below. Not sure what's going on yet.

![trace](https://user-images.githubusercontent.com/25272611/185677869-7d58baee-590e-46f0-8f0d-dc31fcb029a6.png)

GNiendorf:
Here's where the original error is being thrown:

https://github.com/SegmentLinking/TrackLooper/blob/f87e39b2e64fc5354cb88e3284cc05f871ce8211/code/core/trkCore.cc#L970-L978

slava77:
> Here's where the original error is being thrown. It seems like this error occurs if you run over multiple input files, so I get the same error with n=1100 events since each input file stores 1000 events:
> 
> https://github.com/SegmentLinking/TrackLooper/blob/f87e39b2e64fc5354cb88e3284cc05f871ce8211/code/core/trkCore.cc#L970-L978

the stack trace points to https://github.com/SegmentLinking/TrackLooper/blob/0b170ae16a79841802d23e2551a6d08aa6dd19bd/code/core/write_sdl_ntuple.cc#L1020
Is your version of the code the same as in master for this file?

GNiendorf:
> the stack trace points to
> 
> https://github.com/SegmentLinking/TrackLooper/blob/0b170ae16a79841802d23e2551a6d08aa6dd19bd/code/core/write_sdl_ntuple.cc#L1020
> 
> 
> Is your version of the code the same as in master for this file?

Sorry I was not clear, the stack trace points to that location, but if you look at the original screenshot I posted this trkCore.cc hitidx check throws an "Error" before the code actually crashes.

PR #199, opened 2022-10-09
Allow multi-file loop

sgnoohc:
Figured out why when looping over multifile we get segfault and out of bounds error.
The issue was that the `trk.GetEntry(evt)` won't work unless the correct tree in a correct file was initialized to the reader class via `trk.Init(tree)`.
So fixed few lines.


GNiendorf:
Corresponding issue: https://github.com/SegmentLinking/TrackLooper/issues/188

PR #200, opened 2022-10-09
Fix tc eff only

sgnoohc:
The broken feature of plotting only the TC is fixed.

### DIFF
--- a/efficiency/bin/sdl_plot_efficiency
+++ b/efficiency/bin/sdl_plot_efficiency
@@ -124,12 +124,13 @@ END
 )
 
 FULLSTRSAMPLE=$(python -c "$PYTHON_CODE")
-SAMPLE=${FULLSTRSAMPLE//"TObjString = "/}
+SAMPLE=PU200
+# SAMPLE=${FULLSTRSAMPLE//"TObjString = "/}
 
-if [[ "$SAMPLE" == *"/"* ]]; then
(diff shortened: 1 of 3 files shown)


efficiency/bin/sdl_plot_efficiency
```diff
@@ -124,12 +124,13 @@ END
...
+# SAMPLE=${FULLSTRSAMPLE//"TObjString = "/}
 
-if [[ "$SAMPLE" == *"/"* ]]; then
-    SAMPLE=${SAMPLE//\//_-_}
-    SAMPLE=${SAMPLE//./_}
-fi
+# if [[ "$SAMPLE" == *"/"* ]]; then
+#     SAMPLE=${SAMPLE//\//_-_}
+#     SAMPLE=${SAMPLE//./_}
+# fi
```
VourMa, line 133: What do these changes help with? It seems to me like this is a leftover of some check but I am not sure if there is some actual reason behind them

sgnoohc: Ah thanks for catching.
This is accident.

tresreid:
newest validation plots: https://www.classe.cornell.edu/~mgr85/www/SegmentLinking_keep/PR200/eff_plots__GPU_explicit_cache_a7daf44_PU200/

PR #202, opened 2022-10-10
Added files for making fewer plots

MatthewDittrich:


VourMa:
I know that this is a draft PR but I would like to understand what's the strategy for the new steps. By a first look at the code, I can see that the added files are largely a copy of the `efficiency/bin/sdl_plot_efficiency` and `efficiency/python/plot_performance.py` scripts. If that's the case, why not modify the code that's already there with the extra functionality that this PR adds, instead of duplicating it? It will make more difficult to review this PR but also to maintain multiple scripts consistently.

sgnoohc:
Closing as this is already merged to PR212

PR #206, opened 2022-10-12
First Alpaka Kernel Rewrite + Initial Setup

GNiendorf:
Initial setup that allows Alpaka code to run alongside CUDA code, and the first rewrite of a kernel (ModuleRangesKernel) in Alpaka.


VourMa:
Hi Gavin,
I took a look at the PR. However, it is a bit difficult to have a clear picture of the changes, since this PR seems to reapply all of the recent changes to the master.

Should we not freeze the developments of the `alpaka_move` branch, so that we can easily track down what changes with the migration? I understand that this could mean a lot of work at end, which may not be preferable.

The other solution would be to rebase both the `SegmentLinking/alpaka_move` branch and yours to the master developments, so that the differences between between them are only the alpaka related ones. To me, this seems like the optimal solution but I leave it to you and others to comment on the best way to do this.

GNiendorf:
Hey @VourMa I rebased alpaka_move so that now only the relevant commits are shown.

VourMa:
Merging this to have a basis for our alpaka setup in the common `SegmentLinking/alpaka_move` branch.

The plan would be to have a draft PR from `SegmentLinking/alpaka_move` to `SegmentLinking/master` to keep track of the whole migration (rebasing when needed) and have any discussion on our strategy there. We will be making incremental PRs from other branches to `SegmentLinking/alpaka_move`.

PR #209, opened 2022-10-14
Alpaka Migration

VourMa:
Draft PR to keep track of the entirety of the updates needed for the alpaka migration. We will use this PR to have discussions on the implementation details.


SDL/Event.cu
```diff
@@ -599,20 +600,47 @@ __device__ int binary_search(
...
+        struct SDL::modules *modulesInGPU,
+        struct SDL::hits *hitsInGPU,
+        TIdx const & nLowerModules) const
+    -> void
     {
-        uint16_t upperIndex = modulesInGPU->partnerModuleIndices[lowerIndex];
-        if (hitsInGPU->hitRanges[lowerIndex * 2] != -1 && hitsInGPU->hitRanges[upperIndex * 2] != -1)
+
+        TIdx const gridThreadIdx(alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u]);
+        TIdx const threadElemExtent(alpaka::workdiv::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc)[0u]);
```
VourMa, line 619: image
Mapping of Alpaka Abstraction to Hardware from [Alpaka – An Abstraction Library for Parallel Kernel Acceleration](https://arxiv.org/abs/1602.08477).

It seems that the elements do not have a clear mapping on a GPU. From the same paper:

> The element level represents an n-dimensional set of elements and unifies the data parallel capabilities of modern hardware architectures e.g. vectorization on thread level. [...] Alpaka therefore currently relies on compiler recognition of vectorizable code parts. Code is refactored in such a way that it includes primitive inner loops over a fixed number of elements.

Doing tests with a varying number of elements per thread, it seems that elements play a small role even on a GPU. However, I would possibly leave this level of abstraction as a next-to-leading order opportunity for optimization and work with the clear mappings up to threads for starters.

VourMa, line 619: To switch from elements to threads:
```suggestion
        TIdx const threadElemExtent(alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0u]);
```

SDL/Event.cu
```diff
@@ -599,20 +600,47 @@ __device__ int binary_search(
...
+        TIdx const threadFirstElemIdx(gridThreadIdx * threadElemExtent);
+
+        if(threadFirstElemIdx < nLowerModules)
         {
```
VourMa, line 623: To switch from elements to threads:
```suggestion
        TIdx const threadElemExtent(alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0u]);
```

SDL/Event.cu
```diff
@@ -599,20 +600,47 @@ __device__ int binary_search(
...
+            {
+                uint16_t upperIndex = modulesInGPU->partnerModuleIndices[i];
+                if (hitsInGPU->hitRanges[i * 2] != -1 && hitsInGPU->hitRanges[upperIndex * 2] != -1)
+                {
+                    hitsInGPU->hitRangesLower[i] =  hitsInGPU->hitRanges[i * 2]; 
+                    hitsInGPU->hitRangesUpper[i] =  hitsInGPU->hitRanges[upperIndex * 2];
+                    hitsInGPU->hitRangesnLower[i] = hitsInGPU->hitRanges[i * 2 + 1] - hitsInGPU->hitRanges[i * 2] + 1;
+                    hitsInGPU->hitRangesnUpper[i] = hitsInGPU->hitRanges[upperIndex * 2 + 1] - hitsInGPU->hitRanges[upperIndex * 2] + 1;
+                }
+            }
```
VourMa, line 639: To switch from elements to threads:
```suggestion
```

SDL/Event.cu
```diff
@@ -599,20 +600,47 @@ __device__ int binary_search(
...
-            hitsInGPU->hitRangesLower[lowerIndex] =  hitsInGPU->hitRanges[lowerIndex * 2]; 
-            hitsInGPU->hitRangesUpper[lowerIndex] =  hitsInGPU->hitRanges[upperIndex * 2];
-            hitsInGPU->hitRangesnLower[lowerIndex] = hitsInGPU->hitRanges[lowerIndex * 2 + 1] - hitsInGPU->hitRanges[lowerIndex * 2] + 1;
-            hitsInGPU->hitRangesnUpper[lowerIndex] = hitsInGPU->hitRanges[upperIndex * 2 + 1] - hitsInGPU->hitRanges[upperIndex * 2] + 1;
+            // Calculate the number of elements to compute in this thread.
+            // The result is uniform for all but the last thread.
+            TIdx const threadLastElemIdx(threadFirstElemIdx+threadElemExtent);
+            TIdx const threadLastElemIdxClipped((nLowerModules > threadLastElemIdx) ? threadLastElemIdx : nLowerModules);
+
+            for(TIdx i(threadFirstElemIdx); i<threadLastElemIdxClipped; ++i)
```
VourMa, line 629: To switch from elements to threads:
```suggestion
            for(auto i = gridThreadIdx; i < nLowerModules ; i += threadElemExtent)
```
This line makes it clear how straightforward a migration from cuda to alpaka is if we are to use the abstractions up to threads.

SDL/Event.cu
```diff
@@ -709,11 +737,50 @@ void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::
...
+    alpaka::vec::Vec<Dim, Idx> const extent(numElements);
+
+    // Let alpaka calculate good block and grid sizes given our full problem extent
+    alpaka::workdiv::WorkDivMembers<Dim, Idx> const workDiv(
+        alpaka::workdiv::getValidWorkDiv<Acc>(
+            devAcc,
+            extent,
+            elementsPerThread,
+            false,
+            alpaka::workdiv::GridBlockExtentSubDivRestrictions::Unrestricted));
```
VourMa, line 769: If we wanted to have manual (cuda-like) work divisions:
```suggestion
    Idx const blocksPerGrid = 5; // The equivalent of MAX_BLOCKS but using the value from the alpaka automatic work division, which gives 5x speed-up.
    Idx const threadsPerBlock = 1024; // The equivalent of 256 but using the value from the alpaka automatic work division, which gives 5x speed-up.
    Idx const elementsPerThread = 1;
    
    Idx const numElements(nLowerModules);
    
    auto workDiv = alpaka::workdiv::WorkDivMembers<Dim, Idx>{blocksPerGrid, threadsPerBlock, elementsPerThread};
```

SDL/Event.cu
slava77: there is a limited degree of vectorization possible with reduced precision on more recent architectures.
The paper is kind of old.
Maybe some effect can be seen in cases with half-precision data.

GNiendorf:
Commenting here to keep a log of changes. I just rebased to bring the branch up to master and added a commit (fc737f7) to fix a couple things after the rebase. 

Performance plots:
https://www.classe.cornell.edu/~gsn27/www/www/fc737f7/mtv/var/

Here are the timing results on the Cornell machine:
Screenshot 2023-01-24 at 12 55 40 PM

Current master timing results:
Screenshot 2023-01-24 at 12 59 29 PM

GNiendorf:
New Timing at rebase 424136a (slightly faster than before)
Screenshot 2023-05-02 at 1 52 38 PM

Validation Plots: Soon
Comparison to Mater: Soon

GNiendorf:
There are now 3 cases of isnan() in Quintuplet.cuh following the rebase. These either have to be removed, or we need to update the Alpaka version to accommodate for them.

Timing after rebase:
Screenshot 2023-05-17 at 1 25 39 PM


Validation Plots: [Here](https://www.classe.cornell.edu/~gsn27/www/www/rebase_new/mywork_1b12ee-PU200/summary/)
Comparison to Master: [Here](https://www.classe.cornell.edu/~gsn27/www/www/rebase_new/mywork_1b12ee-PU200_cb4e59-PU200/summary/)

VourMa:
> There are now 3 cases of isnan() in Quintuplet.cuh following the rebase to bring alpaka_move up to date with master.

Where do these come from? Is it something to worry about?

GNiendorf:
> > There are now 3 cases of isnan() in Quintuplet.cuh following the rebase to bring alpaka_move up to date with master.
> 
> Where do these come from? Is it something to worry about?

At first glance it looks like kind of questionable code. I'm not sure if we should be relying on isnan() calls in the algorithm, seems like it's just asking for trouble.

https://github.com/SegmentLinking/TrackLooper/blob/1b12ee5b8be13f9ebf6c9530873c3635eaf039f3/SDL/Quintuplet.cu#L797-L798

https://github.com/SegmentLinking/TrackLooper/blob/1b12ee5b8be13f9ebf6c9530873c3635eaf039f3/SDL/Quintuplet.cu#L839

VourMa:
Marking this "ready for review" but the review will take some time and probably come in doses, just because of the sheer magnitude of the PR.

By the way, I can see conflicts. @GNiendorf could you please take care of those? Or do you want to waiting for merging Jonathan's open PR (only one to be merged in before we hold developments for a bit so that the migration is finalized), before dealing with conflicts?

GNiendorf:
> Marking this "ready for review" but the review will take some time and probably come in doses, just because of the sheer magnitude of the PR.
> 
> By the way, I can see conflicts. @GNiendorf could you please take care of those? Or do you want to waiting for merging Jonathan's open PR (only one to be merged in before we hold developments for a bit so that the migration is finalized), before dealing with conflicts?

Yes those conflicts will go away with the rebase. I should have it done by today.

VourMa (commented):
First set of comments is there (now in the correct PR) - to be continued...

I have two general questions:
1. What happened to the `PrintUtils.cc/h` files? They were there in PR#292 but I can no longer find them in this one - not even as deleted. Maybe they were lost during rebasing? If so, let's be careful so that the history stays and we can go back to whatever file we want.
2. I noticed a couple of `.cc/.h` files under the `SDL` folder. Are they supposed to be there? I thought that would become a purely `.cu/.cuh`-file folder...
...or not? What is the proper file extension for Alpaka files?

SDL/PrintUtil.cc
GNiendorf, line 1: @VourMa They were moved to /cpu/ in the last PR, but then I removed the cpu directory so now they just show as deleted from SDL. I haven't force pushed the rebase yet.

code/alpaka_interface/AllocatorConfig.h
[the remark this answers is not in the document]
GNiendorf: Yup, these are the replacements of the CUDA caching allocator for Alpaka. These are all taken from CMSSW. Sounds good, I will add that.

Makefile
```diff
@@ -46,13 +47,13 @@ cutvalue_primitive: $(ROOUTIL) efficiency $(EXES)
 
 
 bin/doAnalysis: bin/doAnalysis.o $(OBJECTS)
-	$(LD) $(PT0P8) $(T3T3EXTENSION) $(LDFLAGS) $^ $(ROOTLIBS) $(EXTRACFLAGS) $(CUTVALUEFLAG) $(PRIMITIVEFLAG) $(EXTRAFLAGS) $(DOQUINTUPLET) -o $@
+	$(CC) $(PT0P8) $(T3T3EXTENSION) $(LDFLAGS) $^ $(ROOTLIBS) $(EXTRACFLAGS) $(CUTVALUEFLAG) $(PRIMITIVEFLAG) $(EXTRAFLAGS) $(DOQUINTUPLET) $(ALPAKAINCLUDE) -o $@
```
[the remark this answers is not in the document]
GNiendorf: Oh yeah, good catch. I'll fix this.

setup_cgpu.sh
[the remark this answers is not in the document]
GNiendorf: Yup, when I force push the rebase this will be brought up to date. I think these files could be reconciled, but there are some unique paths in each that would have to be resolved. Let's look at this more after the rebase and see what we can do.

SDL/PrintUtil.cc
VourMa: Right, I missed them - thanks!

VourMa:
> Those are old files which are being used in various places on the host.

So, are they gonna be removed from there? Removed altogether? Or do they need to be also migrated to an alpaka-like syntax?

GNiendorf:
> > Those are old files which are being used in various places on the host.
> 
> So, are they gonna be removed from there? Removed altogether? Or do they need to be also migrated to an alpaka-like syntax?

They can stay there for now with no changes. I will just change the makefile slightly when I move all extensions over to .cc/.h. The last time I did this (never pushed it though) I just had them compiled with NVCC and everything was fine.

GNiendorf:
@VourMa Rebase is complete. No more conflicts for now. I will keep pushing changes while we wait for the last open PR to be merged. I'll also post updated timing and validation plots soon.

setup_cgpu.sh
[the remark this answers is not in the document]
GNiendorf: sdl_run also has to be changed a bit for the cornell machine. I don't use sdl_run but will make that commit soon.

GNiendorf:
Cornell Machine, Master Branch
Screenshot 2023-07-18 at 9 27 45 PM
Cornell Machine, This PR
Screenshot 2023-07-18 at 9 28 25 PM

VourMa:
Thanks for the report, @GNiendorf! Things look good at this stage. I will try to do the rest of the review today.

> I will keep pushing changes while we wait for the last open PR to be merged.

What other changes should we expect? Since I have started the review, I would prefer not to run after a constantly moving `HEAD`.

GNiendorf:
> Thanks for the report, @GNiendorf! Things look good at this stage. I will try to do the rest of the review today.
> 
> > I will keep pushing changes while we wait for the last open PR to be merged.
> 
> What other changes should we expect? Since I have started the review, I would prefer not to run after a constantly moving `HEAD`.

Just what we’ve talked about (changing file names and adjusting makefile to make that work, and some more random cleanup) before I rebase again for the last open PR when it gets merged in.

SDL/Constants.cuh
```diff
@@ -1,78 +1,122 @@
...
-#define __F2H_seg __float2half  
-#define __H2F_seg __half2float  
-typedef __half FPX_seg;
-#else
-#define __F2H_seg
-#define __H2F_seg
-typedef float FPX_seg; 
+
+#ifndef ALPAKA_ACC_GPU_CUDA_ENABLED
+struct uint4
```
VourMa, line 47: Where is this used? Maybe some comment would help, because it seems very, very generic.

VourMa:
> Just what we’ve talked about (changing file names and adjusting makefile to make that work, and some more random cleanup) before I rebase again for the last open PR when it gets merged in.

Ok, makes sense. However, I am worrying that renaming stuff could mess up the change log, as I see, so let's do it as the very last step, after I have finished reviewing.

bin/sdl.cc
```diff
@@ -524,9 +519,6 @@ void run_sdl()
     std::cout << "Time for event creation = " << timeForEventCreation << " ms\n";
     printTimingInformation(timevec, full_elapsed, avg_elapsed);
 
-    SDL::cleanModules();
-    SDL::freeEndCapMapMemory();
```
VourMa, line 528: This happens automatically now, because they are alpaka buffers?

SDL/LST.cc
```diff
@@ -10,8 +10,7 @@ void SDL::LST::eventSetup() {
     SDL::initModules(path);
 }
 
-void SDL::LST::run(cudaStream_t stream,
```
VourMa, line 13: I am afraid this won't work for CMSSW. We need to pass the queue managed by the framework, so that the kernels run there. I guess one solution would be to have another constructor which is accepting and using an externally provided queue (I haven't tested this and the tests should ideally go along with the fix)?

SDL/Makefile
```diff
@@ -16,33 +16,30 @@ LIB=libsdl.so
...
-ROOTCFLAGS           = --compiler-options -pthread --compiler-options -std=c++17 -m64 -I/cvmfs/cms.cern.ch/slc7_amd64_gcc900/cms/cmssw/CMSSW_11_2_0_pre5/external/slc7_amd64_gcc900/bin/../../../../../../../slc7_amd64_gcc900/lcg/root/6.20.06-ghbfee3/include
+CXX                  = g++
+CXXFLAGS             = -g -Wall -Wshadow -std=c++17 -Woverloaded-virtual -fPIC -fopenmp -I..
+CXXFLAGS_CUDA        =  -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_70 -I/mnt/data1/dsr/cub --use_fast_math --default-stream per-thread -I..
+ROOTCFLAGS           = -pthread -std=c++17 -m64 -I/cvmfs/cms.cern.ch/slc7_amd64_gcc900/cms/cmssw/CMSSW_11_2_0_pre5/external/slc7_amd64_gcc900/bin/../../../../../../../slc7_amd64_gcc900/lcg/root/6.20.06-ghbfee3/include
+ALPAKAINCLUDE        = -I${ALPAKA_ROOT}/include -I/${BOOST_ROOT}/include --std=c++17
+ALPAKASERIAL         = -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
+ALPAKACUDA           = -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_ACC_GPU_CUDA_ONLY --expt-relaxed-constexpr
```
VourMa, line 25: For simplicity, maybe get rid of the recurring `--std=c++17` flag, by putting it in the most general context (I guess `ALPAKAINCLUDE` works?)?

VourMa (commented):
Second set of comments - the next one should be the last, aiming to review all of the SDL files.

I guess there is one major comment that deserves some thinking: how we don't break the CMSSW configuration.

SDL/Constants.cuh
```diff
@@ -1,78 +1,122 @@
...
 
-#ifdef FP16_Base //This changes pT5 and pT3  and T3 completely. T5 for non regression parameters
-#define __F2H __float2half  
-#define __H2F __half2float  
-typedef __half FPX;
+// Half precision wrapper functions, turned off.
+#if defined(FP16_Base)
+#define __F2H //__float2half
+#define __H2F //__half2float
+typedef /*__half*/ float FPX;
```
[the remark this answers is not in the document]
GNiendorf: Those extra flags were not being used anywhere in the code. Also yeah good point, I'll just remove the flag from the makefile like you said.

SDL/Constants.cuh
GNiendorf: This is only needed because we are compiling LST.cc and sdl.cc with g++, and it includes some of the .cu files and .cuh files. uint4 is defined by CUDA, so I needed to have a generic implementation to not through compilation errors from the g++ stuff. This will become more important when we try to run the code on the CPU, because we need a uint4 implementation that mimics what our code does now on the GPU or we need to make changes and remove uint4 from the algo.

bin/sdl.cc
GNiendorf: Well for both I use shared ptr's now which are freed when they go out of scope. Since these two things are both in global scope right now they get freed at the very end automatically. Although the modules have buffers within the object, and those also get freed at the end when the shared ptr gets freed.

SDL/Makefile
GNiendorf: Kk will do.

SDL/LST.cc
GNiendorf: I see. Yeah that sounds reasonable to have a second constructor.

setup_cgpu.sh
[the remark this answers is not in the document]
GNiendorf: @VourMa Yes the cornell and ucsd setup files should be good to go. I've never used the setup_hpg file though so I'm unsure about that one.

code/alpaka_interface/AllocatorConfig.h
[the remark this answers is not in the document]
GNiendorf: @VourMa I added a README to the directory instead of file comments to make updating the allocator easier in the future.

SDL/Constants.cuh
VourMa: So, this is some kind of placeholder for the CPU implementation then? I am fine with it staying in with a comment added, thanks!

bin/sdl.cc
VourMa: Got it, thanks! I still don't like the global scope but let's deal with it in another PR.

SDL/Constants.cuh
GNiendorf: Yup, pretty much.

SDL/PixelTriplet.cuh
```diff
@@ -51,83 +41,1398 @@ namespace SDL
...
+
+    struct createPixelTripletsInGPUFromMapv2
+    {
+        template<typename TAcc>
+        ALPAKA_FN_ACC void operator()(
+                TAcc const & acc,
+                struct SDL::modules& modulesInGPU,
+                struct SDL::objectRanges& rangesInGPU,
+                struct SDL::miniDoublets& mdsInGPU,
+                struct segments& segmentsInGPU,
```
VourMa, line 855: Why, on the other hand, have we removed the namespace from this one?

setup_hpg.sh
```diff
@@ -40,4 +40,8 @@ export LSTPERFORMANCEWEBDIR=/home/users/phchang/public_html/LSTPerformanceWeb
 # List of benchmark efficiencies are set as an environment variable
 export LATEST_CPU_BENCHMARK_EFF_MUONGUN=
 export LATEST_CPU_BENCHMARK_EFF_PU200=
+
+source /cvmfs/cms.cern.ch/el8_amd64_gcc10/external/alpaka/develop-20220902-e80d13b043e1608b43d2007d06ad7e2f/etc/profile.d/init.sh
+export BOOST_ROOT="/cvmfs/cms.cern.ch/el8_amd64_gcc10/external/boost/1.78.0-12075919175e8d078539685f9234134a"
+export ALPAKA_ROOT="/cvmfs/cms.cern.ch/el8_amd64_gcc10/external/alpaka/develop-20220902-e80d13b043e1608b43d2007d06ad7e2f"
```
VourMa, line 46: I think there is an inconsistency with the `SCRAM_ARCH` used in this setup?

setup_lnx7188.sh
VourMa, line 1: I miss the essential difference between this setup file and the one used in the UCSD machines. Can we merge them?

VourMa (commented):
This should be the last set of comments on the actual PR (maybe we will have another iteration after the final rebase). I didn't see any major problems and I hope I didn't miss anything... On top of the mostly trivial, targeted comments I left, I have two more general questions:

1. Given that so many memory-related things were modified, did you run `nvprof` to verify that everything checks out in terms of host/device allocate/free? I guess this is especially relevant since we now use more `new` in the code, which have to be explicitly deleted.
2. Then, I made a couple of comments but I notice a very weird pattern of using the `SDL` namespace, specifically connected to the `segments` struct. Is there some peculiarity with that?

setup_lnx7188.sh
GNiendorf: It's this line: export LSTPERFORMANCEWEBDIR="/cdat/tem/${USER}/LSTPerformanceWeb"
Compared to this: export LSTPERFORMANCEWEBDIR="/home/users/phchang/public_html/LSTPerformanceWeb"

On the Cornell machine you have to make your own LSTPerformanceWeb so that your performance plots can be displayed at all.

SDL/MiniDoublet.cuh
```diff
@@ -73,61 +53,901 @@ namespace SDL
...
 
+    struct addMiniDoubletRangesToEventExplicit
+    {
+        template<typename TAcc>
+        ALPAKA_FN_ACC void operator()(
+                TAcc const & acc,
+                struct modules& modulesInGPU,
+                struct miniDoublets& mdsInGPU,
+                struct objectRanges& rangesInGPU,
+                struct SDL::hits& hitsInGPU) const
```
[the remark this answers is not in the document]
GNiendorf: Yeah this was just an inconsistency in the formatting when I was making a change. Good catch, I will normalize the formatting for all of these.

SDL/PixelTriplet.cuh
GNiendorf: See above, just a formatting inconsistency. I'll normalize it.

setup_lnx7188.sh
VourMa: I see. Since this is a minor difference in only one place, can we fix it with something similar to these lines and still merge the files?
```
if [[ "$(hostname)" == *"cgpu"* ]]; then
    python3 efficiency/python/lst_plot_performance.py ...
elif [[ "$(hostname)" == *"phi3"* ]]; then
    python3 efficiency/python/lst_plot_performance.py ...
else
    python3 efficiency/python/lst_plot_performance.py ...
```

GNiendorf: Hmm I'll give it a shot, thank you for the suggestion.

SDL/Event.cu
```diff
@@ -301,724 +73,503 @@ void SDL::Event::resetEvent()
...
 void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::vector<float> z, std::vector<unsigned int> detId, std::vector<unsigned int> idxInNtuple)
 {
     // Use the actual number of hits instead of a max.
     const int nHits = x.size();
 
-    // Get current device for future use.
-    cudaGetDevice(&dev);
+    // Needed for the memcpy to hitsInGPU below. Will be replaced with a View.
+    auto nHits_buf = allocBufWrapper<unsigned int>(devHost, 1, queue);
+    *alpaka::getPtrNative(nHits_buf) = nHits;
```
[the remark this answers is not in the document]
GNiendorf: Oh I forgot about this, thank you. This was when I was on an older version of Alpaka where you were only allowed to copy from one buffer to another buffer (with some exceptions). Now I can use views to make this cleaner.

SDL/MiniDoublet.cuh
```diff
@@ -73,61 +53,901 @@ namespace SDL
         float* outerLowEdgeX;
         float* outerLowEdgeY;
 
-#ifdef CUT_VALUE_DEBUG
-        //CUT VALUES
-        float* dzCuts;
-        float* drtCuts;
-        float* drts;
-        float* miniCuts;
-#endif
```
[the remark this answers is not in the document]
GNiendorf: It's been a while since I moved this file over, but looking at the code I don't think any of these variables are ever used in the code. I checked for occurrences of each but can't find anything, so these must be left over on the master branch from a previous PR. I think that's why I removed them from here. @YonsiG are these variables okay to remove? It seems like you commented out their definitions in Minidoublet.cu around a year ago but left them intact here.

SDL/Event.cu
```diff
@@ -1702,507 +1465,427 @@ unsigned int SDL::Event::getNumberOfQuintupletsByLayerEndcap(unsigned int layer)
...
-
-        trackCandidatesInCPU->trackCandidateType = new short[nTrackCandidates];
-        trackCandidatesInCPU->hitIndices = new unsigned int[14 * nTrackCandidates];
-        trackCandidatesInCPU->pixelSeedIndex = new int[nTrackCandidates];
-
-        cudaMemcpyAsync(trackCandidatesInCPU->hitIndices, trackCandidatesInGPU->hitIndices, 14 * nTrackCandidates * sizeof(unsigned int), cudaMemcpyDeviceToHost, stream);
-        cudaMemcpyAsync(trackCandidatesInCPU->pixelSeedIndex, trackCandidatesInGPU->pixelSeedIndex, nTrackCandidates * sizeof(int), cudaMemcpyDeviceToHost, stream);
-        cudaMemcpyAsync(trackCandidatesInCPU->trackCandidateType, trackCandidatesInGPU->trackCandidateType, nTrackCandidates * sizeof(short), cudaMemcpyDeviceToHost,stream);
-        cudaStreamSynchronize(stream);
+        // Get nTrackLocal parameter to initialize host based trackCandidatesInCPU
```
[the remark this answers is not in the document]
GNiendorf: Just local to the function is all. Do you think I should rename or change them to avoid confusion?

setup_hpg.sh
GNiendorf: Oh good catch, I will update.

SDL/Event.cu
```diff
@@ -1,19 +1,13 @@
...
-struct SDL::modules* SDL::modulesInGPU = nullptr;
-struct SDL::pixelMap* SDL::pixelMapping = nullptr;
+std::shared_ptr<SDL::modules> SDL::modulesInGPU = std::make_shared<modules>();
+std::shared_ptr<SDL::modulesBuffer<Acc>> SDL::modulesBuffers = std::make_shared<modulesBuffer<Acc>>(devAcc);
+std::shared_ptr<SDL::pixelMap> SDL::pixelMapping = std::make_shared<pixelMap>();
 uint16_t SDL::nModules;
 uint16_t SDL::nLowerModules;
 
-SDL::Event::Event(cudaStream_t estream,bool verbose)
+SDL::Event::Event(bool verbose): queue(alpaka::getDevByIdx<Acc>(0u))
```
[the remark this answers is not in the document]
GNiendorf: I'll give it a shot and see if it works.

SDL/Module.cu
```diff
@@ -1,1013 +1,9 @@
...
-#include "allocate.h"
-#include <unordered_map>
+
+// TODO: Change this to remove it from global scope.
 std::map <unsigned int, uint16_t> *SDL::detIdToIndex;
 std::map <unsigned int, float> *SDL::module_x;
 std::map <unsigned int, float> *SDL::module_y;
 std::map <unsigned int, float> *SDL::module_z;
 std::map <unsigned int, unsigned int> *SDL::module_type; // 23 : Ph2PSP, 24 : Ph2PSS, 25 : Ph2SS
 // https://github.com/cms-sw/cmssw/blob/5e809e8e0a625578aa265dc4b128a93830cb5429/Geometry/TrackerGeometryBuilder/interface/TrackerGeometry.h#L29
```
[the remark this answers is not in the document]
GNiendorf: Fixed this and removed Module.cu, thank you.

SDL/MiniDoublet.cuh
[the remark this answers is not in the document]
GNiendorf: The change log seems to be glitching on this PR, but this is fixed by https://github.com/SegmentLinking/TrackLooper/commit/082fb3d8fca6674f08d7a0a5e43f66856bc2a1a5

SDL/EndcapGeometry.cu
```diff
@@ -58,30 +62,23 @@ void SDL::EndcapGeometry::load(std::string filename)
...
     fillGeoMapArraysExplicit();
 }
 
-void SDL::freeEndCapMapMemory()
+void SDL::EndcapGeometry::fillGeoMapArraysExplicit()
 {
-    cudaFree(SDL::endcapGeometry.geoMapPhi);
-    cudaFree(SDL::endcapGeometry.geoMapDetId);
-}
+    QueueAcc queue(devAcc);
```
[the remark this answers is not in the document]
GNiendorf: I create an independent stream for global variables when they require one, like in this case. I'll post a screenshot from the timeline soon.

SDL/Constants.cuh
```diff
@@ -1,78 +1,124 @@
...
+const unsigned int N_MAX_TRACK_CANDIDATES = 1000;
+const unsigned int N_MAX_PIXEL_TRACK_CANDIDATES = 4000;
+
+const unsigned int N_MAX_TRACK_CANDIDATE_EXTENSIONS = 200000;
+const unsigned int N_MAX_TRACK_EXTENSIONS_PER_TC = 30;
+const unsigned int N_MAX_T3T3_TRACK_EXTENSIONS = 40000;
+
+const unsigned int size_superbins = 45000;
+
+// Temporary fix for endcap buffer allocation.
```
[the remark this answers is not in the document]
GNiendorf: To remove these variables from the global scope in a later PR hopefully. And to do it in a way that also works for the cmssw integration. That reminds me though, let me put some relevant checks in the code for now so that this doesn't cause problems in the meantime.

GNiendorf:
@VourMa nsys profile with 8 streams showing no memory leaks.

Screenshot 2023-07-21 at 10 09 32 AM

GNiendorf:
Alpaka timeline overview for 8 streams. You can see the 8 streams which do most of the work, and two additional streams. One stream is for the endcap maps and the other is for the modules initialization. Streams are only needed for the data transfer like mentioned above because these variables are not using the caching allocator.

Screenshot 2023-07-21 at 10 16 17 AM

GNiendorf:
@VourMa Last thing left now is the CMSSW integration work. I'll work on it today and hopefully push a fix today or tomorrow based on your suggestions.

SDL/Event.cu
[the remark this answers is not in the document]
VourMa: I guess it confused me, especially since we don't write things in this way for other similar functions. I would rename to omit "Local" from the names, if you agree.

SDL/MiniDoublet.cuh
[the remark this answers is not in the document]
YonsiG: I'm not sure if those variables are valid for use now. Last time, when I was trying to use those variables and compile with "-d", which means defining CUT_VALUE_DEBUG, it couldn't function correctly at least in the Minidoublet.cu, so I comment out that part. If this is not used anywhere else, I believe we can remove it.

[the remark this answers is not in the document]
GNiendorf: Sounds good, thank you Yonsi

SDL/Constants.cuh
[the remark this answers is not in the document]
VourMa: In the past, printout statements were causing an increase in the number of registers for some kernels, most notoriously for the T5 one. Can you check that this is not the case here?

In any case however, I would wrap those within the `Warnings` `ifdef` and maybe agree with people to run with that enabled when checking PRs? I would go ahead with wrapping the printouts no matter what and we can discuss next week on the validation procedure. I trust that these checks don't output anything for now, right?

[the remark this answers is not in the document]
GNiendorf: @VourMa These checks only occur on the CPU, so their cost should only be checking one condition each (the one condition inside the if statement). All of this is outside of the main timers as well (map loading, etc.) so I don't think this is an issue. My concern with wrapping these in the Warnings ifdef is that if these were to fail, it's possible they could fail silently and cause issues.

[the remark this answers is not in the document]
VourMa: I agree that this could potentially be an issue and that's why I proposed to define a workflow for checking it. However, I get your point about these being only on the CPU (sorry, I had not noticed), so they will most likely cause no issue. Let's not do anything about them for now but keep an eye for increased registers in the profiling report you are going to produce, just to make sure that there isn't anything we have overlooked.

SDL/Event.cu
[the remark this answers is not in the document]
GNiendorf: Fixed, thank you.

GNiendorf:
@VourMa I gave the CMSSW integration problems a shot. The second constructor method was the only one that I could get to work. When you get a chance can you see if it works? Thank you!

GNiendorf:
Alpaka does not have a log10 function yet, so I had to create a temporary one. I checked and it doesn't impact the timing or the performance plots. I'll post updated timing/plots after I rebase when the DNN PR is merged in.

GNiendorf:
@VourMa I got the algo to run on the CPU finally. I'm still cleaning things up though, so I'll probably not push anything until a couple of days from now. If everything works in regards to the cmssw integration though let's not wait on the CPU backend, I can always make a new PR for it afterwards.

VourMa:
> I can always make a new PR for it afterwards.

Yep, let's not overload this PR. It's already hard to test/follow.

VourMa:
@GNiendorf Please go ahead with the rebase. Try to not push anymore new features to this branch, so that we can be done with the review today and check the status with tests from other people during today's meeting.

I will try the CMSSW workflows today and let you know how it goes.

code/alpaka_interface/AllocatorConfig.h
```diff
@@ -0,0 +1,31 @@
+#ifndef HeterogeneousCore_AlpakaInterface_interface_AllocatorConfig_h
+#define HeterogeneousCore_AlpakaInterface_interface_AllocatorConfig_h
+
+#include <limits>
+
+namespace cms::alpakatools {
```
VourMa, line 6: This namespace won't work - it conflicts with the CMSSW one... Can we switch to `lst::`?

VourMa: > Can we switch to `lst::`?

I tried this but it didn't solve the initial error. I am sending the initial error [0]. Just to make it clear, the only thing I get out of LST for CMSSW is the `libsdl.so`. The include paths seem to be ok.

[0]:
```
In file included from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:9,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:6,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Constants.h:5,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Hit.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Event.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/LST.h:11,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/RecoTracker/LST/plugins/alpaka/LSTProducer.cc:22:
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachingAllocator.h:87:59: error: 'is_device_v' is not a member of 'cms::alpakatools'
   87 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                           ^~~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachingAllocator.h:87:75: error: template argument 1 is invalid
   87 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                           ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachingAllocator.h:87:77: error: expected '>' before 'and' token
   87 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                             ^~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachingAllocator.h:87:118: error: expected unqualified-id before '>' token
   87 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                                                      ^
In file included from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:6,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Constants.h:5,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Hit.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Event.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/LST.h:11,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/RecoTracker/LST/plugins/alpaka/LSTProducer.cc:22:
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:19:61: error: 'is_device_v' is not a member of 'cms::alpakatools'
   19 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                             ^~~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:19:77: error: template argument 1 is invalid
   19 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                             ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:19:79: error: expected '>' before 'and' token
   19 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                               ^~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:19:120: error: expected unqualified-id before '>' token
   19 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                                                        ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:75:59: error: 'is_device_v' is not a member of 'cms::alpakatools'
   75 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                           ^~~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:75:75: error: template argument 1 is invalid
   75 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                           ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:75:77: error: expected '>' before 'and' token
   75 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                             ^~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getDeviceCachingAllocator.h:75:118: error: expected unqualified-id before '>' token
   75 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                                                      ^
In file included from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:7,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Constants.h:5,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Hit.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Event.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/LST.h:11,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/RecoTracker/LST/plugins/alpaka/LSTProducer.cc:22:
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getHostCachingAllocator.h:13:76: error: 'is_queue_v' is not a member of 'cms::alpakatools'
   13 |   template <typename TQueue, typename = std::enable_if_t<cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                            ^~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getHostCachingAllocator.h:13:87: error: template argument 1 is invalid
   13 |   template <typename TQueue, typename = std::enable_if_t<cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                       ^~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/getHostCachingAllocator.h:13:95: error: expected unqualified-id before '>' token
   13 |   template <typename TQueue, typename = std::enable_if_t<cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                               ^
In file included from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Constants.h:5,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Hit.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Event.h:4,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/LST.h:11,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/RecoTracker/LST/plugins/alpaka/LSTProducer.cc:22:
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:21:61: error: 'is_device_v' is not a member of 'cms::alpakatools'
   21 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                             ^~~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:21:77: error: template argument 1 is invalid
   21 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                             ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:21:79: error: expected '>' before 'and' token
   21 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                               ^~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:21:120: error: expected unqualified-id before '>' token
   21 |               typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                                                        ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:28:12: error: 'CachedBufAlloc' is not a class template
   28 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, TQueue, void> {
      |            ^~~~~~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:41:95: error: wrong number of template arguments (6, should be 4)
   41 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, alpaka::QueueCudaRtBlocking, void> {
      |                                                                                               ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:28:74: note: provided for 'template<class TElem, class TDim, class TIdx, class TQueue> struct cms::alpakatools::traits::CachedBufAlloc'
   28 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, TQueue, void> {
      |                                                                          ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:64:98: error: wrong number of template arguments (6, should be 4)
   64 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, alpaka::QueueCudaRtNonBlocking, void> {
      |                                                                                                  ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:28:74: note: provided for 'template<class TElem, class TDim, class TIdx, class TQueue> struct cms::alpakatools::traits::CachedBufAlloc'
   28 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, TQueue, void> {
      |                                                                          ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:87:77: error: wrong number of template arguments (6, should be 4)
   87 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCudaRt, TQueue, void> {
      |                                                                             ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:28:74: note: provided for 'template<class TElem, class TDim, class TIdx, class TQueue> struct cms::alpakatools::traits::CachedBufAlloc'
   28 |     struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, TQueue, void> {
      |                                                                          ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:200:59: error: 'is_device_v' is not a member of 'cms::alpakatools'
  200 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                           ^~~~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:200:75: error: template argument 1 is invalid
  200 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                           ^
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:200:77: error: expected '>' before 'and' token
  200 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                             ^~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/../code/alpaka_interface/CachedBufAlloc.h:200:118: error: expected unqualified-id before '>' token
  200 |             typename = std::enable_if_t<cms::alpakatools::is_device_v<TDev> and cms::alpakatools::is_queue_v<TQueue>>>
      |                                                                                                                      ^
In file included from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Event.h:6,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/LST.h:11,
                 from /home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/RecoTracker/LST/plugins/alpaka/LSTProducer.cc:22:
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h: In function 'float SDL::isTighterTiltedModules_seg(SDL::modules&, unsigned int)':
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h:187:33: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
  187 |                         | (side == NegZ & (((layer == 2) & (rod > 5)) | ((layer == 1) & (rod > 9))))
      |                            ~~~~~^~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h:188:33: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
  188 |                         | (side == PosZ & (((layer == 2) & (rod < 8)) | ((layer == 1) & (rod < 4)))) );
      |                            ~~~~~^~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h:186:23: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
  186 |         return subdet == Barrel & ( ((side != Center) & (layer== 3))
      |                ~~~~~~~^~~~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h: In function 'float SDL::isTighterTiltedModules_seg(short int, short int, short int, short int)':
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h:197:33: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
  197 |                         | (side == NegZ & (((layer == 2) & (rod > 5)) | ((layer == 1) & (rod > 9))))
      |                            ~~~~~^~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h:198:33: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
  198 |                         | (side == PosZ & (((layer == 2) & (rod < 8)) | ((layer == 1) & (rod < 4)))) );
      |                            ~~~~~^~~~~~~
/home/users/evourlio/LSTinCMSSW/cgpu-1/CMSSW_13_0_0_pre4/src/../../../TrackLooper/SDL/Segment.h:196:23: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
  196 |         return subdet == Barrel & ( ((side != Center) & (layer== 3))
      |                ~~~~~~~^~~~~~~~~
gmake: *** [config/SCRAM/GMake/Makefile.rules:1740: tmp/el8_amd64_gcc11/src/RecoTracker/LST/plugins/RecoTrackerLSTPluginsPortableCudaAsync/alpaka/LSTProducer.cc.o] Error 1
gmake: *** [There are compilation/build errors. Please see the detail log above.] Error 2

```

GNiendorf: From the error, it doesn’t seem like you changed all of the cms::alpakatools in the directory to lst::, I’ll push a fix

SDL/Makefile
```diff
@@ -3,13 +3,12 @@
 # stuff to make
 #
 
-CCSOURCES=$(wildcard *.cc)
-CCOBJECTS=$(CCSOURCES:.cc=_cpu.o)
-CCHEADERS=$(CCSOURCES:.cc=.h)
+CCSOURCES=$(filter-out LST.cc, $(wildcard *.cc))
+CCOBJECTS=$(CCSOURCES:.cc=_cuda.o)
```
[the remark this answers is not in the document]
GNiendorf: Yup, this will change in the next PR.

GNiendorf:
I'm going to start posting validation plots, timing, and profiler differences between the two branches now that we have all of the issues worked out:

Validation Plots (No NN) - [Here](https://www.classe.cornell.edu/~gsn27/www/www/PRAlpakaRocks/mywork_162bc9D-PU200/summary/)
Master Plots (No NN, no Half-Precision) - [Here ](https://www.classe.cornell.edu/~gsn27/www/www/PRAlpakaRocks/mywork_a8c352D-PU200/summary/)
Comparison of Above - [Here](https://www.classe.cornell.edu/~gsn27/www/www/PRAlpakaRocks/mywork_a8c352D-PU200_162bc9D-PU200/summary/)

VourMa:
This PR is good to go from my side. Before merging, I have two comments so that we don't forget:
- A lower fake rate was observed in the transition region wrt. the master:
https://www.classe.cornell.edu/~gsn27/www/www/PRAlpakaRocks/mywork_a8c352D-PU200_162bc9D-PU200/summary/TC_fakerate_etacoarsezoom.html
- The caching version causes a crash when exiting the job in CMSSW:
`Fatal system signal has occurred during exit
Aborted (core dumped)`
Despite this, the output files seem to be fine.

GNiendorf (approved):
Best PR I've ever seen. Beautiful Code. A+ Job.

slava77:
> `Fatal system signal has occurred during exit Aborted (core dumped)`

is there a stack trace?

PR #210, opened 2022-10-21
Move CUDA Functions to Alpaka

GNiendorf:
Draft PR for moving CUDA functions to Alpaka. 

I have the initial setup working, so I just need to move all of the functions over now.

Current timing with a few functions moved over with this commit:
Screen Shot 2022-10-21 at 5 58 44 PM


GNiendorf:
Going to open a new PR.

PR #211, opened 2022-10-22
Cgpu migration

bsathian:
Generic Makefile and a new setup.sh file for the cgpu machine


sgnoohc:
The changes already pulled into PR212

PR #212, opened 2022-10-22
PR 202 + PR 211 + New plotting scheme and various cleaning of the output and performance plot making workflow

sgnoohc:
This already merges changes in the following two PR:
 - PR 202
 - PR 211

So those two don't need to be pulled and instead this should be fine.

To check quickly:

e.g. on cgpu-1
git clone --branch cleaning_v0 git@github.com:SegmentLinking/TrackLooper.git
cd TrackLooper/
source setup_cgpu.sh
sdl_make_tracklooper -mc8
sdl -i PU200 -o LSTNtuple.root
createPerfNumDenHists -i LSTNtuple.root -o LSTNumDen.root
python3 efficiency/python/lst_plot_performance.py LSTNumDen.root -t "myTag" # if you are on cgpu-1


sgnoohc:
Performance plot
http://uaf-10.t2.ucsd.edu/~phchang/performance/myTag_b60f2b-PU200/summary/

efficiency/Makefile
```diff
@@ -35,10 +38,22 @@ src/efficiency.o: src/efficiency.cc src/efficiency.h src/sdl_types.h
...
 
 clean:
 	rm -f $(OBJECTS) $(EXE)
 
+summary: 
+	wget http://uaf-10.t2.ucsd.edu/~phchang/lst/summary.tar.gz && tar xf summary.tar.gz && rm summary.tar.gz
+
+compare: 
+	wget http://uaf-10.t2.ucsd.edu/~phchang/lst/compare.tar.gz && tar xf compare.tar.gz && rm compare.tar.gz
+
```
VourMa, line 58: What are these exactly?

efficiency/misc/make_html.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
VourMa, line 1: Do we need this file? I just see personal directories, so I am not sure we want it if it's only possible to run it on specific machines.

efficiency/misc/make_lstpage.py
```diff
@@ -0,0 +1,344 @@
+#!/bin/env python
+
+import os
+import glob
+import sys
+
+#___________________________________________________________________________________________________
+def write_pages_v2(directory, objecttypes):
```
VourMa, line 8: Same comment as before: any usage for `v1`?

efficiency/src/helper_v2.cc
```diff
@@ -0,0 +1,434 @@
...
+    {
+        for (int i = 0; i < 41; ++i)
+        {
+            pt_boundaries.push_back(pow(10., -1. + 4. * i/40.));
+        }
+    }
+    return pt_boundaries;
+}
+
+//__________________________________________________________________________________________
```
VourMa, line 308: Do we need all of the following commented out lines?

VourMa (commented):
Thanks a lot for the PR! I think it's great and will help a lot to increase our productivity. I specifically appreciated the general clean up and the tidy and explanatory functions for parsing the objects.

I have some comments on specific lines and some others more general.

VourMa:
> Performance plot http://uaf-10.t2.ucsd.edu/~phchang/performance/myTag_b60f2b-PU200/summary/

The link does not work for me. I get a "403 Forbidden" error:

"You don't have permission to access /~phchang/performance/myTag_b60f2b-PU200/summary/ on this server."

VourMa:
I know this PR focused a lot on cleaning up the code but I think that a clean up of files is also needed. Since you spent some time on understanding how we make plots, efficiencies, etc., I was thinking that it would make sense to delete some of the files in the efficiency directory. For example, `efficiency/bin` and `efficiency/python` seem to have a lot of scripts that seem to be doing kinda of the same thing but maybe not? However, I am not sure that we use all of them.

Maybe people (@bsathian @tresreid @YonsiG) can comment on what they have been using and delete the rest? Or, even, maybe your PR replaces all of these files?

Related to this, do we need a `v2` of the `helper` code?

VourMa:
Finally, could you update the PR title/description to be a bit more descriptive? For example, I think you removed the track candidate extensions but this is not mentioned anywhere. Having a good description of what went in this PR would greatly help looking back at PRs and understanding changes. Thanks!

bin/sdl.cc
```diff
@@ -283,325 +281,218 @@ int main(int argc, char** argv)
...
+    std::vector<std::vector<short>> out_isQuad_vec;
+    std::vector<int> evt_num;
+    std::vector<TString> file_name;
+
     // Looping input file
     while (ana.looper.nextEvent())
     {
 
         // if (ana.looper.getCurrentEventIndex() ==49) {continue;}
         std::cout << "PreLoading event number = " << ana.looper.getCurrentEventIndex() << std::endl;
```
VourMa, line 326: I would put this as well under `verbosity >= 1`. Or maybe even drop it all together, since this step happens really fast and, in >99% of the cases, without any issues.

README.md
```diff
@@ -17,23 +29,48 @@ Once every new shell, source the setup script to initilaize the enviornment.
...
+    -o: provide an output root file name (e.g. num_den_hist.root)
+    -n: (optional) number of events
+    -g: (optional) comma separated pdgids to add more efficiency plots with different sim particle slices
     
+Plotting performance plots
+
+    lst_plot_performance.py num_den_hist.root -t "mywork"
+
+When running on ```cgpu-1``` remember to specify python3 as there is no python.
+The shebang on the ```lst_plot_performance.py``` is not updated as ```lnx7188``` works with python2....
```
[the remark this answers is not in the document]
VourMa: I checked and, even though `python3` works on `lnx7188`, the command `python3 efficiency/python/lst_plot_performance.py LSTNumDen.root -t "myTag"` doesn't.

I checked the same thing on `phi3` the latter command again doesn't work, even though the former does.

The error is the following:
```
Traceback (most recent call last):
  File "efficiency/python/lst_plot_performance.py", line 4, in <module>
    import ROOT as r
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc900/lcg/root/6.20.06-ghbfee3/lib/ROOT.py", line 24, in <module>
    import cppyy
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc900/lcg/root/6.20.06-ghbfee3/lib/cppyy.py", line 61, in <module>
    import libPyROOT as _backend
ImportError: dynamic module does not define module export function (PyInit_libPyROOT)
```

Therefore, I suspect that this is an issue with the (older) version of ROOT we are using with our `setup.sh` on those machines not supporting `python3` rather than an actual issue with `python3`.

README.md
```diff
@@ -1,6 +1,18 @@
...
+## Quick Start
+
+    git clone git@github.com:SegmentLinking/TrackLooper.git
+    cd TrackLooper/
+    source setup.sh
+    # source setup_cgpu.sh # if you are on cgpu-1
+    sdl_make_tracklooper -mc8
+    sdl -i PU200 -o LSTNtuple.root
+    createPerfNumDenHists -i LSTNtuple.root -o LSTNumDen.root
+    lst_plot_performance.py LSTNumDen.root -t "myTag"
```
VourMa, line 12: Can we suppress the output of this command. It is really long and probably no one will read it.

VourMa:
> python3 efficiency/python/lst_plot_performance.py LSTNumDen.root -t "myTag" # if you are on cgpu-1

After I have run, it's not clear to me how I get the webpage. I used to just copy the folder to my public_html. Now that doesn't work as expected:
https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/myTag_b60f2b-PU200/

What am I missing?

README.md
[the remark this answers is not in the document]
bsathian: I think this is some pesky CMSSW stuff. Probably that ROOT libraries are only available for python2 in the older versions (CMSSW 11), and only in python3 in the newer (CMSSW 12) versions

bsathian:
To summarize:-

1. cgpu-1 uses brand new versions of things, so a new `setup_cgpu.sh` was created, with new paths. A consequence of the `CMSSW_12` versions is that `python2` is not a thing anymore, and the command `python` doesn't exist. Hence, things were moved to `python3`
2. The main LST code itself runs well on all  machines, what I didn't notice was that the efficiency plotter will have issues in the older machines
3. The `CMSSW_11_2_0_pre5` version is quite old (source : Slava) and uses ROOT 6.20, which is probably why ROOT python3 libraries don't exist for this version.

I'm looking at whether we can move to a more recent CMSSW version with the slc7 architecture. After all, we don't critically depend on CMSSW at this juncture, we use it only so that we don't need to install root and python and all that stuff locally. So moving to a different CMSSW version should not be an issue

sgnoohc:
OK pulled the new environment setup script

VourMa:
I am trying to run the timing on this branch on the Cornell machine with:
```
source setup.sh
sdl_timing PU200 explicit_cache 200
```
However I get the following crash:
``` 
*** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================

Thread 3 (Thread 0x7fc27c571700 (LWP 57837) "cuda-EvtHandlr"):
#0  0x00007fc29d487ccd in poll () from /lib64/libc.so.6
#1  0x00007fc284780b11 in ?? () from /lib64/libcuda.so.1
#2  0x00007fc28480ab5a in ?? () from /lib64/libcuda.so.1
#3  0x00007fc28474ff46 in ?? () from /lib64/libcuda.so.1
#4  0x00007fc29d769ea5 in start_thread () from /lib64/libpthread.so.0
#5  0x00007fc29d4929fd in clone () from /lib64/libc.so.6

Thread 2 (Thread 0x7fc27cd72700 (LWP 57836) "sdl"):
#0  0x00007fc29d493e8f in accept4 () from /lib64/libc.so.6
#1  0x00007fc284780243 in ?? () from /lib64/libcuda.so.1
#2  0x00007fc28474ff46 in ?? () from /lib64/libcuda.so.1
#3  0x00007fc29d769ea5 in start_thread () from /lib64/libpthread.so.0
#4  0x00007fc29d4929fd in clone () from /lib64/libc.so.6

Thread 1 (Thread 0x7fc29aa33000 (LWP 57658) "sdl"):
#0  0x00007fc29d459549 in waitpid () from /lib64/libc.so.6
#1  0x00007fc29d3d6fb2 in do_system () from /lib64/libc.so.6
#2  0x00007fc29d3d7361 in system () from /lib64/libc.so.6
#3  0x00007fc2a02daf9b in TUnixSystem::StackTrace() () from /cvmfs/cms.cern.ch/slc7_amd64_gcc10/cms/cmssw/CMSSW_12_2_0/external/slc7_amd64_gcc10/lib/libCore.so
#4  0x00007fc2a02d8555 in TUnixSystem::DispatchSignals(ESignals) () from /cvmfs/cms.cern.ch/slc7_amd64_gcc10/cms/cmssw/CMSSW_12_2_0/external/slc7_amd64_gcc10/lib/libCore.so
#5  <signal handler called>
#6  0x00007fc2a04a4b08 in SDL::EndcapGeometry::fillGeoMapArraysExplicit (this=0x7fc2a0b20440 <SDL::endcapGeometry>) at EndcapGeometry.cu:92
#7  0x00007fc2a04a4905 in SDL::EndcapGeometry::load (this=0x7fc2a0b20440 <SDL::endcapGeometry>, filename=...) at EndcapGeometry.cu:62
#8  0x00000000004824cc in loadMaps () at code/core/trkCore.cc:23
#9  0x0000000000439fa9 in run_sdl () at bin/sdl.cc:289
#10 0x0000000000434659 in main (argc=1, argv=0x7ffec7b2fb38) at bin/sdl.cc:279
===========================================================


The lines below might hint at the cause of the crash.
You may get help by asking at the ROOT forum https://root.cern.ch/forum
Only if you are really convinced it is a bug in ROOT then please submit a
report at https://root.cern.ch/bugs Please post the ENTIRE stack trace
from above as an attachment in addition to anything else
that might help us fixing this issue.
===========================================================
#6  0x00007fc2a04a4b08 in SDL::EndcapGeometry::fillGeoMapArraysExplicit (this=0x7fc2a0b20440 <SDL::endcapGeometry>) at EndcapGeometry.cu:92
#7  0x00007fc2a04a4905 in SDL::EndcapGeometry::load (this=0x7fc2a0b20440 <SDL::endcapGeometry>, filename=...) at EndcapGeometry.cu:62
#8  0x00000000004824cc in loadMaps () at code/core/trkCore.cc:23
#9  0x0000000000439fa9 in run_sdl () at bin/sdl.cc:289
#10 0x0000000000434659 in main (argc=1, argv=0x7ffec7b2fb38) at bin/sdl.cc:279
===========================================================
```

@sgnoohc @bsathian Could you take a look?

README.md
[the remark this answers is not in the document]
sgnoohc: I wanted to separate them because in case we wish to create a few set of plots, i don't want to re-run the whole thing.
The code running to produce the entire set of histograms for all possibilities takes 9 seconds.
The code running the plotter takes 1-2 minutes for a standard set of plots
The code if you specify to generate just a few set of plots, can run in <few seconds.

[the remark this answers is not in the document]
sgnoohc: added.

bin/sdl.cc
```diff
@@ -79,22 +77,25 @@ int main(int argc, char** argv)
...
+        ana.input_file_list_tstring = TString::Format("%s/trackingNtuple_10mu_pt_0p5_50.root", TrackingNtupleDir.Data());
     else if (ana.input_raw_string.EqualTo("pionGun"))
-        ana.input_file_list_tstring = "/data2/segmentlinking/CMSSW_12_2_0_pre2/trackingNtuple_6pion_1k_pt_0p5_50.root";
+        ana.input_file_list_tstring = TString::Format("%s/trackingNtuple_6pion_1k_pt_0p5_50.root", TrackingNtupleDir.Data());
     else if (ana.input_raw_string.EqualTo("PU200"))
-        ana.input_file_list_tstring = "/data2/segmentlinking/CMSSW_12_2_0_pre2/trackingNtuple_ttbar_PU200.root";
+        ana.input_file_list_tstring = TString::Format("%s/trackingNtuple_ttbar_PU200.root", TrackingNtupleDir.Data());
     else if (ana.input_raw_string.EqualTo("PU200RelVal"))
-        ana.input_file_list_tstring = "/data2/segmentlinking/CMSSW_12_5_0_pre3/RelValTTbar_14TeV_CMSSW_12_5_0_pre3/";
+        ana.input_file_list_tstring = TString::Format("%s/RelValTTbar_14TeV_CMSSW_12_5_0_pre3/", TrackingNtupleDir.Data());
```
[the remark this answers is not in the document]
sgnoohc: ah. um. i suppose we can keep the code, but do a better orgnaization of the actual directory where the "TrackingNtupleDir" points to.
This is done via setup.sh commands.
Would that be OK?

bin/sdl.cc
```diff
@@ -230,9 +231,6 @@ int main(int argc, char** argv)
 
     //_______________________________________________________________________________
     // --lower_level
-#ifdef CUT_VALUE_DEBUG
```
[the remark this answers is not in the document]
sgnoohc: @YonsiG Is this OK?

bin/sdl.cc
```diff
@@ -23,42 +22,41 @@ int main(int argc, char** argv)
...
+        ("o,output"          , "Output file name", cxxopts::value<std::string>())
+        ("N,nmatch"          , "N match for MTV-like matching", cxxopts::value<int>()->default_value("9"))
+        ("n,nevents"         , "N events to loop over", cxxopts::value<int>()->default_value("-1"))
+        ("x,event_index"     , "specific event index to process", cxxopts::value<int>()->default_value("-1"))
+        ("g,pdg_id"          , "The simhit pdgId match option (default = 0)", cxxopts::value<int>()->default_value("0"))
+        ("v,verbose"         , "Verbose mode (0: no print, 1: only final timing, 2: object multiplitcity", cxxopts::value<int>()->default_value("0"))
+        ("w,write_ntuple"    , "Write Ntuple", cxxopts::value<int>()->default_value("1"))
+        ("s,streams"         , "Set number of streams (default=1)", cxxopts::value<int>()->default_value("1"))
+        ("d,debug"           , "Run debug job. i.e. overrides output option to 'debug.root' and 'recreate's the file.")
+        ("c,cpu"             , "Run CPU version of the code.")
```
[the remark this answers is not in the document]
sgnoohc: cpu i removed, streams, debug and write_ntuple are still there.

bin/sdl.cc
```diff
@@ -283,325 +281,218 @@ int main(int argc, char** argv)
...
-    
-    cudaStreamCreateWithFlags(&streams[s],cudaStreamNonBlocking);
-    SDL::Event* event = new SDL::Event(streams[s]);;//(streams[omp_get_thread_num()]);
-    events.push_back(event);
+    for (int s = 0; s < ana.streams; s++)
+    {
+
+        cudaStreamCreateWithFlags(&streams[s], cudaStreamNonBlocking);
+        SDL::Event *event = new SDL::Event(streams[s]);
+        ; //(streams[omp_get_thread_num()]);
```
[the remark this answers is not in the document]
sgnoohc: I am not sure... what this line is. have I just simply removed white spaces? perhaps?

code/core/trkCore.cc
```diff
@@ -1278,43 +869,334 @@ std::vector<std::vector<short>>&    out_isQuad_vec)
...
+
+
+
+
+
+
+
+
+
+// DEPRECATED FUNCTIONS
```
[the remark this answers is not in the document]
sgnoohc: I was just a bit worried that I may not want to delete them right away... but I am OK with nuking.

code/core/write_sdl_ntuple.cc
```diff
@@ -6,8 +6,544 @@
...
 void createOutputBranches()
+{
+    createOutputBranches_v2();
+}
+
+//________________________________________________________________________________________________________________________________
+void fillOutputBranches(SDL::Event* event)
+{
+    fillOutputBranches_v2(event);
+}
```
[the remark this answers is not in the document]
sgnoohc: ... so idk. probably not...? but i wasn't sure... similar to previous comment.

efficiency/Makefile
sgnoohc: this is to download the web pages html files.
The way I do it is I generate .md files then i use pandoc to turn them into html.
but lnx7188 does not have pandoc installed.
So it can't generate on the fly, so I decided to save them in my area. so people can download them web format.
Given that this won't change very often, I thought it'd be OK.
and I didn't want to save a gazillion number of .html files.

I am not 100% satisfied with this honestly.

efficiency/misc/make_html.sh
sgnoohc: this is the issue with the pandoc that is not installed on lnx7188.
Perhaps what we can do is to have a separate git repo for "LSTPerformanceWebpage" (or some better name)
and have them as a submodule.
And I can update the repo separate from our code.

efficiency/misc/make_lstpage.py
sgnoohc: i can  nuke them..

efficiency/src/helper_v2.cc
sgnoohc: probablly not

bin/sdl.cc
sgnoohc: i am not quite sure i follow which line needs to be put under verbosity >=1.
Do you mean evt_num and file_name ? this is necessary for writing out ntuple.

bin/sdl.cc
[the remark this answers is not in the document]
VourMa: @bsathian Does this line do anything? Is there a reason why it exists or should it be removed?

bin/sdl.cc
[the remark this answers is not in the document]
VourMa: Ok, thanks for the clarification. Can we remove the `-c` flag then?

bin/sdl.cc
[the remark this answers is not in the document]
VourMa: Yes, that would be the best.

code/core/trkCore.cc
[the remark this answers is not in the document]
VourMa: In the past, these functions have only confused us, at least in my experience. With the clarification you made, maybe that's not the case now. Let's not nuke for now.

code/core/write_sdl_ntuple.cc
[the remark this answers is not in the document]
VourMa: Contrary to my reply in the previous comment, I would be in favor of nuking those. These seem simple enough to reconstruct their pieces, if we need them in the future (which I doubt, since your updates are both fast and complete). And if the removal happens in a single comment, it would be easy to just check that and understand what we were doing.

efficiency/Makefile
VourMa: Sure, makes sense.

Concerning the location of these files, would it work better to add them to our /data2 directory of both phi3 and lnx7188, so that we don't depend on uaf-10 being available (which is increasingly doubtful recently)? These are small, tarred files, so it should be fine.

efficiency/misc/make_html.sh
VourMa: I think I got the whole concept of the html creation. I don't have any specific preference, so I leave it up to you to handle it as you prefer. Resolving the comment, thanks for the explanations.

efficiency/src/helper_v2.cc
VourMa: Nuke them as well then?

bin/sdl.cc
VourMa: I think the confusion comes from the GitHub interface. When it shows a code block like this, the comment only refers to the last line shown - the others are for reference. In this comment, I was only referring to the last line:
```
std::cout << "PreLoading event number = " << ana.looper.getCurrentEventIndex() << std::endl;
```
This line "spams" us for no reason when we run the code.

bsathian:
Strangely, I'm able to run on phi3

bsathian:
Update : If I use the old ROOT versions and the old config (setup.sh file here : https://www.classe.cornell.edu/~bsathian/setup_temp.sh), this works on lnx7188

bsathian:
Summary of (new) issues :-
1. Some weird CUDA issue that props up only in lnx7188
2. cuda-gdb does not run due to python mismatch 
3. Reverting to the age old CMSSW for the time being, and replacing python3 with python runs perfectly

YonsiG:
> Regarding the redundancy under efficiency/bin, I usually use the sdl_plot_multiple_efficiencies, sdl_timing, sdl_validate_efficiency and I haven't used the other files yet. For the efficiency/python, I think the SDL.cc and SDL.h are used to specify branches? or I'm wrong probably, but I never used those tbh

bin/sdl.cc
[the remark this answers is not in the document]
YonsiG: Hi Philip, I suggest we keep this flag without removing it. Because once we use the CUT_VALUE_DEBUG, we create more branches than the usual output, we save the cut values and a few chi2s which is not needed for the usual output but still under tuning.

efficiency/python/lst_plot_performance.py
```diff
@@ -0,0 +1,816 @@
...
+    f2 = r.TFile(file2)
+
+    h1 = f1.Get("githash").GetTitle()
+    h2 = f2.Get("githash").GetTitle()
+
+    i1 = f1.Get("input").GetTitle()
+    i2 = f2.Get("input").GetTitle()
+
+    params = {}
+    params["output_dir"] = "comparison/{}_v_{}".format(h1, h2)
```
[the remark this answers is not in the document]
sgnoohc: Could you elaborate a bit more..?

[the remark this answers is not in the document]
VourMa: Let's imagine that I have three efficiencies with the tags "ref", "test1" and "test2". Suppose that I want to run the comparison "test1" vs. "ref" and "test2" vs. "ref". Do I understand correctly that for both comparisons the plots will end up in a folder named "comparison", i.e. the comparison folder has no info on the tags that I compared? If that's the case, I was thinking that it would be better to include the info on the tags compared in the folder name. That could spare us from mistakenly overwriting comparison folders as well, because, e.g. we forgot to manually rename the file before running another comparison.

Or maybe I didn't read the code well and folder names of comparisons are clearly distinguished by some identifier?

README.md
```diff
@@ -1,6 +1,18 @@
 # TrackLooper
 
-## Quick start guide
+## Quick Start
+
+    git clone git@github.com:SegmentLinking/TrackLooper.git
+    cd TrackLooper/
+    source setup.sh
+    # source setup_cgpu.sh # if you are on cgpu-1
+    sdl_make_tracklooper -mc8
```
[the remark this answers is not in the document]
sgnoohc: done

efficiency/python/lst_plot_performance.py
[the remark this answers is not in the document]
sgnoohc: Actually this function is not used.
And the actualy compare option will put things in something like

performance/(tag)_githash-sample_githash-sample/

So it should be alright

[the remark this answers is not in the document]
VourMa: Since we compare, shouldn't there be two tags in the names? And if the function is not used, should we consider removing it to avoid confusion? In any case, I am resolving this comment, thanks for the clarifications.

README.md
```diff
@@ -1,6 +1,18 @@
...
+    sdl -i PU200 -o LSTNtuple.root
+    createPerfNumDenHists -i LSTNtuple.root -o LSTNumDen.root
+    lst_plot_performance.py LSTNumDen.root -t "myTag"
+    # python3 efficiency/python/lst_plot_performance.py LSTNumDen.root -t "myTag" # if you are on cgpu-1
```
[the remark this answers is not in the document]
sgnoohc: done

efficiency/Makefile
sgnoohc: Put them in a separate repository that one has do a recursive checkout

README.md
sgnoohc: with bin/sdl_run these are addressed.

sgnoohc:
Currently when I run the sdl_timing I get

```Total Timing Summary
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       TCE      Event      Short           Rate
   avg      3.3      2.7      1.7      4.0      3.8      1.4      1.5      1.2      2.6      22.3      17.6+/-  3.6      23.7   explicit_cache[s=1]
   avg      4.5      4.0      2.3      5.6      6.2      2.3      2.7      2.0      4.0      33.7      26.8+/-  4.7      17.7   explicit_cache[s=2]
   avg      6.8      5.8      3.9      9.7     10.7      3.2      6.6      3.6      6.7      57.0      47.0+/-  9.0      15.0   explicit_cache[s=4]
   avg     12.9      6.6      4.8     13.1     15.4      3.6     10.4      5.4     10.1      82.3      65.9+/- 11.9      14.2   explicit_cache[s=6]
   avg     23.9      7.6      5.3     14.6     18.6      3.7     13.2      7.5     14.0     108.3      80.8+/- 18.8      14.1   explicit_cache[s=8]```

... how do I understand the numbers increasing as i go to larger s?

(ran on cgpu-1)

File: bin/sdl.cc (hunk 9 above)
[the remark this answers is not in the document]
Comment (VourMa) [reply bin/sdl.cc:233]: @sgnoohc, Yanxi  asked that we keep this flag but I believe this has not been reverted. Is it in the plans?

I saw a lot of function deletions in the writeNTuple. I hope they are not deleting functions related to this?

In any case, before merging, I would ask @YonsiG and @bsathian, who are working on the tuning of more low level stuff to run this branch to check that everything they need for their studies is still included.

File: bin/sdl.cc (hunk 11 above)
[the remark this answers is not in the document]
Comment (VourMa) [reply bin/sdl.cc:364]: I guess delete this? To my naive eyes, it does nothing (but deleting will tell).

File: bin/sdl.cc (hunk 10 above)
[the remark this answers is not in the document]
Comment (VourMa) [reply bin/sdl.cc:55]: Is this removed in the latest version?

File: bin/sdl_run [hunk 17]
```diff
@@ -0,0 +1,94 @@
...
+echo "  TAG               : ${TAG}"
+echo ""
+echo "  (cf. Run > sh $(basename $0) -h to see all options)"
+echo ""
+
+JOBTAG=${TAG}_${SAMPLE}_NEVT${NEVENTS}
+
+rm -f ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log
+echo "Compiling code..."
+sdl_make_tracklooper ${FLAGS} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log
```
VourMa, line 83: On lnx7188:
```
source setup_lnx7188.sh
sdl_run -f -mc -s PU200 -n 100 -t testRun
```
gives:
```
Compiling code...
/mnt/data1/evourlio/TrackLooper/bin/sdl_run: line 83: /data2/segmentlinking/evourlio/testRun_PU200_NEVT100__LSTRun.log: No such file or directory
Running LST code...
/mnt/data1/evourlio/TrackLooper/bin/sdl_run: line 85: /data2/segmentlinking/evourlio/testRun_PU200_NEVT100__LSTRun.log: No such file or directory
ERROR: sdl command failed!
```

Why does this point to `/data2/segmentlinking/evourlio`? Issue in the setup files (the one for `lnx7188` is different in the `LSTOUTPUTDIR` variable wrt. the other for some reason)?

bin/sdl_run
```diff
@@ -0,0 +1,94 @@
...
+echo "Running LST code..."
+sdl -i ${SAMPLE} -o ${LSTOUTPUTDIR}/${JOBTAG}__LSTNtuple.root -n ${NEVENTS} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: sdl command failed!' ; exit 1; }
+echo "Creating performance histograms..."
+createPerfNumDenHists -i ${LSTOUTPUTDIR}/${JOBTAG}__LSTNtuple.root -o ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: createPerfNumDenHists command failed!' ; exit 1; }
+echo "Creating plots..."
+if [[ "$(hostname)" == *"cgpu"* ]]; then
+    python3 efficiency/python/lst_plot_performance.py ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root -t ${LSTOUTPUTDIR}/${JOBTAG} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: lst_plot_performance.py command failed!' ; exit 1; }
+else
+    lst_plot_performance.py ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root -t ${LSTOUTPUTDIR}/${JOBTAG} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: lst_plot_performance.py command failed!' ; exit 1; }
```
VourMa, line 92: I think the proper directory is missed here?
```suggestion
    efficiency/python/lst_plot_performance.py ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root -t ${LSTOUTPUTDIR}/${JOBTAG} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: lst_plot_performance.py command failed!' ; exit 1; }
```

bin/sdl_run
```diff
@@ -0,0 +1,94 @@
...
+echo "  SAMPLE            : ${SAMPLE}"
+echo "  NEVENTS           : ${NEVENTS}"
+echo "  TAG               : ${TAG}"
+echo ""
+echo "  (cf. Run > sh $(basename $0) -h to see all options)"
+echo ""
+
+JOBTAG=${TAG}_${SAMPLE}_NEVT${NEVENTS}
+
+rm -f ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log
```
VourMa, line 81: Running the same command for this script repeatedly fails because all of the scripts used within it stop running if their output files previously exist. Can we add a flag, say `-d` that would take care of deleting all these files, such as `rm ./testRun_PU200_NEVT100__LSTNtuple.root ./testRun_PU200_NEVT100__LSTNumDen.root`

By the way, is it easy to put the outputs in a folder called `${JOBTAG}`? Otherwise the top directory gets very quickly cluttered...

bin/sdl_run
VourMa: Changing `LSTOUTPUTDIR` to `.` in the `setup_ln7188.sh` file works.

bin/sdl_run
sgnoohc: The output dir can be changed via,

export LSTOUTPUTDIR=/my/favorite/output/dir

I will think about the existing output... thingy..
I usually don't like overwriting something that's already written.
But even if that's the case I guess a better "fail exit messagE" would be better.

bin/sdl_run
```diff
@@ -0,0 +1,94 @@
...
+sdl -i ${SAMPLE} -o ${LSTOUTPUTDIR}/${JOBTAG}__LSTNtuple.root -n ${NEVENTS} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: sdl command failed!' ; exit 1; }
+echo "Creating performance histograms..."
+createPerfNumDenHists -i ${LSTOUTPUTDIR}/${JOBTAG}__LSTNtuple.root -o ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: createPerfNumDenHists command failed!' ; exit 1; }
+echo "Creating plots..."
+if [[ "$(hostname)" == *"cgpu"* ]]; then
+    python3 efficiency/python/lst_plot_performance.py ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root -t ${LSTOUTPUTDIR}/${JOBTAG} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: lst_plot_performance.py command failed!' ; exit 1; }
+else
+    lst_plot_performance.py ${LSTOUTPUTDIR}/${JOBTAG}__LSTNumDen.root -t ${LSTOUTPUTDIR}/${JOBTAG} >> ${LSTOUTPUTDIR}/${JOBTAG}__LSTRun.log 2>&1 || { echo 'ERROR: lst_plot_performance.py command failed!' ; exit 1; }
+fi
+echo "Done!"
```
VourMa, line 94: When I am done, I get a folder, which I copy to my web direectory. However, it doesn't work properly:
https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/testRun_PU200_NEVT100_19f50aD-PU200/

Not sure what I am doing wrong...

bin/sdl_run
VourMa: > The output dir can be changed via,
> `export LSTOUTPUTDIR=/my/favorite/output/dir`

Sure. But let's have the default something that works.

VourMa: > I will think about the existing output... thingy..
> I usually don't like overwriting something that's already written.
> But even if that's the case I guess a better "fail exit messagE" would be better.

I agree and that's why I propose a flag that the user would consciously put there. The issue with the error messages is that the failure happens in multiple stages.

In my case, I had to do
```
rm ./testRun_PU200_NEVT100__LSTNtuple.root
```
only to get another fail for the existence of `./testRun_PU200_NEVT100__LSTNumDen.root`, so did
```
rm ./testRun_PU200_NEVT100__LSTNumDen.root
```
only to get another error, because I needed to do both removals at once. I imagine this could get very tiring, very quickly...

bin/sdl_run
sgnoohc: Have you done git submodule update --init --recursive?
i.e is LSTPerformanceWeb/ empty or it has content?

sgnoohc: basically the directory is missing "summary/" and "compare/" directory
And that should be in the LSTPerformanceWeb/
And it should be "soft-linked" (like this: https://github.com/SegmentLinking/TrackLooper/blob/cleaning_v0/efficiency/python/lst_plot_performance.py#L153)

bin/sdl_run
sgnoohc: this shouldn't be necessary as PATH includes efficiency/python directory.

bin/sdl_run
sgnoohc: I added mkdir -p 
and also added file override features

bin/sdl_run
sgnoohc: I added mkdir -p 
and also added file override features

bin/sdl_run
VourMa: > Have you done git submodule update --init --recursive?

You're right, I hadn't run this, my bad. Can you add this note to the PR description, so that other people notice as well?

VourMa: Still, it is not clear what one needs to do to show this on the web. After the output directory is created (with proper soft-links now), I used to just copy it to my public directory. This doesn't seem to work (see updated https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/testRun_PU200_NEVT100_19f50aD-PU200/) and I couldn't find any instruction of what I should do differently.

bin/sdl_run
VourMa: However, I still get
```
/mnt/data1/evourlio/TrackLooper/bin/sdl_run: line 131: lst_plot_performance.py: command not found
```
so I believe we need to change it

bin/sdl_run
sgnoohc: I believe I have addressed them now where the performanceweb is now stored in some common space so that the link will work

VourMa:
Finally indeed! Thanks a lot, @sgnoohc!

PR #216, opened 2022-10-25
Object accessors

sgnoohc:
Added another set of help functions to help access important informations from the types of objects we produce from the LST.


slava77:
it looks like this needs a rebase after #212 was merged

sgnoohc:
Ready to be merged.
Review requested.

code/core/write_sdl_ntuple.cc
```diff
@@ -199,114 +198,46 @@ std::tuple<float, float, float, vector<unsigned int>, vector<unsigned int>> pars
...
-    const float dr_out = sqrt(pow(hitsInGPU.xs[i5_idx] - hitsInGPU.xs[i3_idx], 2) + pow(hitsInGPU.ys[i5_idx] - hitsInGPU.ys[i3_idx], 2));
-    float betaIn_in   = __H2F(tripletsInGPU.betaIn[T3Idx1]);
-    float betaOut_in  = __H2F(tripletsInGPU.betaOut[T3Idx1]);
-    float betaIn_out  = __H2F(tripletsInGPU.betaIn[T3Idx2]);
-    float betaOut_out = __H2F(tripletsInGPU.betaOut[T3Idx2]);
+    unsigned int pT5 = trackCandidatesInGPU.directObjectIndices[idx];
+    std::vector<unsigned int> Hits = getHitsFrompT5(event, pT5);
+    unsigned int Hit_0 = Hits[0];
+    unsigned int Hit_4 = Hits[4];
+    unsigned int Hit_8 = Hits[8];
```
VourMa, line 210: This PR is very helpful in making the code more understandable but I feel some points may still be a bit cryptic. In this instance for example, why take the 9th hit of a pT5? Can there a comment next to each of the hits to explain what it corresponds to? There are more such instances below.

I add a suggestion below but I am not sure it is the correct one...
```suggestion
    unsigned int Hit_0 = Hits[0]; // First hit of pLS
    unsigned int Hit_4 = Hits[4]; // First hit of 1st OT LS
    unsigned int Hit_8 = Hits[8]; // First hit of 2nd OT LS
```

code/core/write_sdl_ntuple.cc
```diff
@@ -199,114 +198,46 @@ std::tuple<float, float, float, vector<unsigned int>, vector<unsigned int>> pars
...
+    std::vector<unsigned int> T3s = getT3sFrompT5(event, pT5);
+    unsigned int T3_0 = T3s[0];
+    unsigned int T3_1 = T3s[1];
+
+    unsigned int pLS = getPixelLSFrompT5(event, pT5);
+
     const float kRinv1GeVf = (2.99792458e-3 * 3.8);
     const float k2Rinv1GeVf = kRinv1GeVf / 2.;
+    const float dr_in = sqrt(pow(hitsInGPU.xs[Hit_4] - hitsInGPU.xs[Hit_0], 2) + pow(hitsInGPU.ys[Hit_4] - hitsInGPU.ys[Hit_0], 2));
+    const float dr_out = sqrt(pow(hitsInGPU.xs[Hit_8] - hitsInGPU.xs[Hit_4], 2) + pow(hitsInGPU.ys[Hit_8] - hitsInGPU.ys[Hit_4], 2));
```
VourMa, line 221: Related to the comment above, what are we computing here exactly? If I understand correctly, these hits are not on the same segment, or am I mistaken?

In any case, I think this shows that there are still some points that are not clear in the code. And I know this change is not introduced by this PR and it is an opportunity for others to clarify what we are doing.

code/core/write_sdl_ntuple.cc
```diff
@@ -321,131 +252,64 @@ std::tuple<float, float, float, vector<unsigned int>, vector<unsigned int>> pars
...
 
     // pLS  1    2    3
     // **** o -- o -- o
-    unsigned int i1_idx = T3_hit1;
-    unsigned int i3_idx = T3_hit5;
-    const float dr = sqrt(pow(hitsInGPU.xs[i3_idx] - hitsInGPU.xs[i1_idx], 2) + pow(hitsInGPU.ys[i3_idx] - hitsInGPU.ys[i1_idx], 2));
+    unsigned int pT3 = trackCandidatesInGPU.directObjectIndices[idx];
+    std::vector<unsigned int> Hits = getHitsFrompT3(event, pT3);
+    unsigned int Hit_0 = Hits[0];
+    unsigned int Hit_4 = Hits[4];
```
VourMa, line 261: If I read the code correctly, the hit array for pixel hits is of variable length, so it may have either 3 or 4 hits. If that's the case, I am not sure that the 4th hit is always the right choice here (and possibly in other places).

Trying to go back on the function calls:
1. L259 of `write_sdl_ntuple.cc`
2. L229 of `AccessHelper.cc`
3. L233 of `AccessHelper.cc`
4. L8 of `AccessHelper.cc`
5. L21-L24 of `AccessHelper.cc` which gives a variable length array.

If the logic above is indeed what's happening, we would need to understand how to resolve this. I see two ways:
1. We add a "fake" fourth hit, as in the previous logic.
2. We choose the 3rd or 4th hit based on the length of the pLS/pTX object (probably more natural for anything that comes downstream?).

Thinking a bit more about it, maybe we don't care that much, given that in any case the hit we take will be in the same MD, however I find this to be confusing and inconsistent.

code/core/write_sdl_ntuple.cc
```diff
@@ -459,88 +323,39 @@ std::tuple<float, float, float, vector<unsigned int>, vector<unsigned int>> pars
     const float pt = (ptAv_in + ptAv_out) / 2.;
 
     // T5 eta and phi are computed using outer and innermost hits
-    SDL::CPU::Hit hitA(trk.ph2_x()[i1_idx], trk.ph2_y()[i1_idx], trk.ph2_z()[i1_idx]);
-    SDL::CPU::Hit hitB(trk.ph2_x()[i5_idx], trk.ph2_y()[i5_idx], trk.ph2_z()[i5_idx]);
+    SDL::CPU::Hit hitA(trk.ph2_x()[Hit_0], trk.ph2_y()[Hit_0], trk.ph2_z()[Hit_0]);
+    SDL::CPU::Hit hitB(trk.ph2_x()[Hit_8], trk.ph2_y()[Hit_8], trk.ph2_z()[Hit_8]);
```
VourMa, line 327: The comment says that we are to compute η based on the outermost hit. But doesn't a T5 have 10 hits always? If so, why do we take the 9th hit? Do we need the lower one from the outer MD? If so, I think this is again an inconsistency wrt. how we take hits for pTXs...

SDL/TrackCandidate.cu
```diff
@@ -339,7 +347,7 @@ __global__ void SDL::addpT3asTrackCandidatesInGPU(struct SDL::pixelTriplets& pix
         atomicAdd(trackCandidatesInGPU.nTrackCandidatespT3,1);
     
         float radius = 0.5f * (__H2F(pixelTripletsInGPU.pixelRadius[pixelTripletIndex]) + __H2F(pixelTripletsInGPU.tripletRadius[pixelTripletIndex]));
-        addTrackCandidateToMemory(trackCandidatesInGPU, 5/*track candidate type pT3=5*/, pixelTripletIndex, pixelTripletIndex, &pixelTripletsInGPU.logicalLayers[5 * pixelTripletIndex], &pixelTripletsInGPU.lowerModuleIndices[5 * pixelTripletIndex], &pixelTripletsInGPU.hitIndices[10 * pixelTripletIndex], __H2F(pixelTripletsInGPU.centerX[pixelTripletIndex]), __H2F(pixelTripletsInGPU.centerY[pixelTripletIndex]),radius,trackCandidateIdx);
+        addTrackCandidateToMemory(trackCandidatesInGPU, 5/*track candidate type pT3=5*/, pixelTripletIndex, pixelTripletIndex, &pixelTripletsInGPU.logicalLayers[5 * pixelTripletIndex], &pixelTripletsInGPU.lowerModuleIndices[5 * pixelTripletIndex], &pixelTripletsInGPU.hitIndices[10 * pixelTripletIndex], __H2F(pixelTripletsInGPU.centerX[pixelTripletIndex]), __H2F(pixelTripletsInGPU.centerY[pixelTripletIndex]),radius,trackCandidateIdx, pixelTripletIndex);
```
VourMa, line 350: I can see that, in most cases, the additional argument is also the second argument in the same function call, so we could perhaps save ourselves from adding a new argument?

The above doesn't seem to be the case only for the pT5s but this looks like an inconsistency to me that may need addressing.

VourMa:
In general, I think that the PR is a major improvement in terms of code simplicity and clarity wrt. what we have now. However, a few points are not clear to me still and I have the suspicion that a few inconsistencies may have been introduced in the meanwhile.

Due to the fact that the specific part of the code that is being changed is complicated and sometimes confusing (at least to me), I would invite @bsathian and/or @tresreid to take a look, especially at the `AccessHelper.cc` file and the comments I left, since they know the code much better than me.

Finally, I noticed that the functions are written in the "CPU" part of the code, returning vectors. We had discussed in the past the possibility to make them return arrays (as in the current implementation, where we handle arrays) and to move it to the "SDL" part of the code. For the former, I don't have a strong opinion and others can comment, but I think that the latter may be a problem concerning the CMSSW integration. Right now, everything that is used in CMSSW must be in the "SDL" library, so these functions would have to be copied/rewritten to be used for the integration. I guess this can be discussed together with the review of PR #224 and understand what's the best way to proceed.

code/core/write_sdl_ntuple.cc
sgnoohc: The above "cartoon" comment was make this clear but clearly it was not sufficient!

VourMa: But `Hits` also include pLS hits, no? L320 of AccessHelper.cc seems to indicate so but maybe I am missing something?

VourMa: The cartoon is very helpful by the way, I would keep it, thanks!

code/core/write_sdl_ntuple.cc
sgnoohc: I believe it is possible that a T5 be made out of 5 PS modules, (if it's very forward for example.) in which case, the position of the last hit is best known via a pixel hit, instead of the strip hit.
And the ordering of the hits is made always such taht the 9th hit is the "anchor hit" i.e. always a pixel hit if the last two hits are PS module.
Hence, I picked 9th hit which is the anchor hit.

in other words, we always have the hits in the following order.
1. anchor hit
2. other hit
3. anchor hit
4. other hit
...
...
9. anchor hit
10. other hit

And so I chose the "anchor hit" of the last mini-doublet to represent the hit position.

code/core/write_sdl_ntuple.cc
sgnoohc: I added some more information... with some history

sgnoohc:
There was one comment about variable pixel hits that somehow I must have clicked something wrong and deleted?? I get easily lost with github GUI... Sorry..

Anyways, there seems to be a bug!!
The function call should have tried to access "getOuterHitsFrompT3"!!!

Fixing.....

code/core/write_sdl_ntuple.cc
sgnoohc: Ah found this comment now.
This seems like a bug!!!! 
Fixed it with c3f39ff

SDL/TrackCandidate.cu
sgnoohc: I must have not noticed that that variable was already passed on...
It doesn't help the function call is already reallly long...
Let me see if i can rever some of these.

sgnoohc: addTrackCandidateToMemory(trackCandidatesInGPU, 7/*track candidate type pT5=7*/, pixelQuintupletsInGPU.pixelIndices[pixelQuintupletIndex], pixelQuintupletsInGPU.T5Indices[pixelQuintupletIndex], &pixelQuintupletsInGPU.logicalLayers[7 * pixelQuintupletIndex], &pixelQuintupletsInGPU.lowerModuleIndices[7 * pixelQuintupletIndex], &pixelQuintupletsInGPU.hitIndices[14 * pixelQuintupletIndex], __H2F(pixelQuintupletsInGPU.centerX[pixelQuintupletIndex]),
                            __H2F(pixelQuintupletsInGPU.centerY[pixelQuintupletIndex]),radius , trackCandidateIdx, pixelQuintupletIndex);

sgnoohc: Ah I remember now.
So L194 and L350 needs to be called through a same function, and in one case, it is a duplicate information, but in the other it is not.

Fundamentally the problem is due to the fact that there is some sort of harmonization needed between how we treat pT3 and pT5.

VourMa: Is there an action item on this? Something concrete that we could make an issue of and fix it in a new PR?

In any case, I don't think the questions is related to this PR anymore.

code/core/write_sdl_ntuple.cc
VourMa: Thanks for the summary, it was useful.

In any case, my questions were originating from the meaning of these variables before the latest bug fix. After that, I think that everything makes sense.

VourMa (approved):
I think this looks good now, thanks to @sgnoohc!

Unless someone wants to check it as well, I would merge it soon.

PR #217, opened 2022-10-28
Move all CUDA Functions to Alpaka

GNiendorf:
Timing:
Screen Shot 2022-10-28 at 1 03 28 AM

Note: Still need to move a few CUDA_CONST_VAR's (constants defined on the device) over in subsequent PR.

edit: Clarified my Note.


VourMa:
Is this indeed ready to merge? I see a note in the description about moving more cuda stuff. Is this something to be dealt with in this PR, a leftover or mention for a future PR? I haven't checked the whole PR thoroughly but I see some inconsistencies in the changes, examples of which I will leave in separate comments below.

A more general comment: I can see the alpaka header included in a whole bunch of files, while the same files have includes that again include the alpaka header. Can we perhaps simplify on this point?

Finally, do we want to merge this before we merge #212, which will be the basis for us moving to alpaka 0.7, using a branch I have already developed? Apart from the fact that you were worried about conflicts, I am wondering if moving to the new version affects how you have implemented everything in this PR.

SDL/Hit.cuh
```diff
@@ -1,15 +1,10 @@
 #ifndef Hit_cuh
 #define Hit_cuh
 
+#include <alpaka/alpaka.hpp>
 #ifdef __CUDACC__
-#define CUDA_HOSTDEV  __host__ __device__
-#define CUDA_DEV __device__
-#define CUDA_CONST_VAR __device__
 #define CUDA_G __global__
```
VourMa, line 6: Why does this stay here while it is removed from the other files?

SDL/Kernels.cuh
```diff
@@ -26,25 +26,22 @@
 #include "PixelTriplet.cuh"
 #include "TrackExtensions.cuh"
 #include "Constants.cuh"
+#include <alpaka/alpaka.hpp>
 
+#include <alpaka/alpaka.hpp>
```
VourMa, line 31: Typo? (Header added twice)

SDL/PixelTriplet.cuh
```diff
@@ -1,12 +1,9 @@
...
 #define PixelTriplet_cuh
 
+#include <alpaka/alpaka.hpp>
 #ifdef __CUDACC__
-#define CUDA_HOSTDEV  __host__ __device__
-#define CUDA_DEV __device__
 #else
-#define CUDA_HOSTDEV
-#define CUDA_DEV
 #endif
```
VourMa, line 7: Incomplete removal?

SDL/Hit.cuh
GNiendorf: CUDA_G is defined for __global__ functions (i.e. kernels), this just focuses on moving external functions (__device__, __host__, or __device__ __host__) over.

SDL/Kernels.cuh
GNiendorf: Good catch, will fix in a cleanup commit now.

SDL/PixelTriplet.cuh
GNiendorf: Good catch, will fix in a cleanup commit now.

GNiendorf:
> Is this indeed ready to merge? I see a note in the description about moving more cuda stuff. Is this something to be dealt with in this PR, a leftover or mention for a future PR? I haven't checked the whole PR thoroughly but I see some inconsistencies in the changes, examples of which I will leave in separate comments below.

Yes, those will be dealt with in a subsequent PR. Those are constants defined with the device prefix, so they don't fall under functions that should be moved in this PR. I made the note since there are only a few of them, so we don't forget about it.

> A more general comment: I can see the alpaka header included in a whole bunch of files, while the same files have includes that again include the alpaka header. Can we perhaps simplify on this point?

This is a good point, but I'm not sure what best practice is here. If a file used Alpaka functions, I figured it was best to explicitly include alpaka in that file (even if it included a different file from SDL that also had the alpaka include). If you have a suggestion for a better approach though, I can change that in another commit.

> Finally, do we want to merge this before we merge #212, which will be the basis for us moving to alpaka 0.7, using a branch I have already developed? Apart from the fact that you were worried about conflicts, I am wondering if moving to the new version affects how you have implemented everything in this PR.

The prefixes haven't changed from 0.5 to 0.7, so everything in this PR should be the same (I think).

GNiendorf:
Timing - 
Screen Shot 2022-10-28 at 11 46 17 AM

I will update this comment with validation plots soon.

SDL/Event.cuh
```diff
@@ -28,11 +28,7 @@
 #include "Constants.h"
 
 #include "cuda_profiler_api.h"
-#ifdef __CUDACC__
-#define CUDA_G __global__
```
VourMa, line 32: Is this not needed here anymore?

SDL/Hit.cuh
VourMa: I see. I guess part of my confusion then comes from the fact that the same line is removed from other files, see [this comment](https://github.com/SegmentLinking/TrackLooper/pull/217#discussion_r1008675383) for example.

SDL/Event.cuh
VourMa: Looking at the code a bit better, is this removed for the same reason that the alpaka header was removed? Because it is included in the `Hit.cuh` file, which is then included in all of the files that need it?

GNiendorf: Yes, also because the "CUDA_G" definition is only used in Hit.cuh

PR #218, opened 2022-10-31
Move to Alpaka 0.7.0

GNiendorf:
This PR moves the current code from Alpaka 0.5 to 0.7. This is an important thing to update before we move the remaining kernels to Alpaka since there were many namespace simplifications between 0.5 and 0.7.

Timing
Screen Shot 2022-10-31 at 5 13 46 PM


GNiendorf:
Performance plots:

Current: https://www.classe.cornell.edu/~gsn27/www/www/PR217
This PR: https://www.classe.cornell.edu/~gsn27/www/www/PR218

PR #220, opened 2022-11-11
Move Hit Loading Kernels to Alpaka

GNiendorf:
Draft PR to move hit loading kernels to Alpaka. Defining a new queue in each scope as a quick workaround for now.

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -620,10 +620,10 @@ public:
         int const & nLowerModules) const
     {
 
-        int const threadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
-        int const threadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0];
+        int const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        int const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0];
 
-        for(int lowerIndex = threadIdx; lowerIndex < nLowerModules; lowerIndex += threadExtent)
+        for(int lowerIndex = globalThreadIdx; lowerIndex < nLowerModules; lowerIndex += gridThreadExtent)
         {
             uint16_t upperIndex = modulesInGPU->partnerModuleIndices[lowerIndex];
             if (hitsInGPU->hitRanges[lowerIndex * 2] != -1 && hitsInGPU->hitRanges[upperIndex * 2] != -1)
@@ -655,10 +655,10 @@ public:
         int const & nHits) const // Total number of hits in event
     {
 
-        int const threadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
-        int const threadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0];
+        int const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        int const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0];
 
-        for(int ihit = threadIdx; ihit < nHits; ihit += threadExtent)
+        for(int ihit = globalThreadIdx; ihit < nHits; ihit += gridThreadExtent)
         {
             float ihit_x = hitsInGPU->xs[ihit];
             float ihit_y = hitsInGPU->ys[ihit];
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -620,25 +620,18 @@ public:
         int const & nLowerModules) const
     {
 
-        int const gridThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u];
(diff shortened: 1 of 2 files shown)


GNiendorf:
Timing
timing
Will post validation plots soon.

SDL/Event.cu
```diff
@@ -640,57 +647,82 @@ public:
...
 
-        int found_index = binary_search(modulesInGPU->mapdetId, iDetId, nModules);
-        uint16_t lastModuleIndex = modulesInGPU->mapIdx[found_index];
+        if(threadFirstElemIdx < nHits)
+        {
+            TIdx const threadLastElemIdx(threadFirstElemIdx+threadElemExtent);
+            TIdx const threadLastElemIdxClipped((nHits > threadLastElemIdx) ? threadLastElemIdx : nHits);
 
-        hitsInGPU->moduleIndices[ihit] = lastModuleIndex;
+            for(TIdx ihit(threadFirstElemIdx); ihit<threadLastElemIdxClipped; ++ihit)
```
VourMa, line 680: As per this comment [here](https://github.com/SegmentLinking/TrackLooper/pull/209#discussion_r996465189), is there any practical reason to complicate the code like this, instead of writing cuda-like loops? Especially because on the GPUs we don't work on the element-level objects.

SDL/Event.cu
```diff
@@ -640,57 +647,82 @@ public:
...
+            {
+                // Is there a better way to do this?
+                int ihit_int = ihit;
```
VourMa, line 683: I guess the issue is `TIdx` vs. `int`? Do we have to use `TIdx`? I think everything should work even if we define `ihit` and `nHits` as `int`s from the beginning.

SDL/Event.cu
```diff
@@ -722,228 +754,226 @@ void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::
...
+        pixelModuleIndex,
+        *modulesInGPU,
+        *rangesInGPU,
+        *hitsInGPU,
+        *mdsInGPU,
+        *segmentsInGPU,
+        size_elements));
+
+    alpaka::enqueue(queue, pixel_segments_task);
+    alpaka::wait(queue);
```
VourMa, line 963: Out of curiosity, does this replace a `cudaStreamSynchronize(stream);`?

- If yes, then that's fine, it closely follows the previous implementation. However, it's a bit weird that later (L970) we also use `cudaStreamSynchronize(stream);`. Is there some detail here (synchronize stream vs. wait for queue) that we need to pay more attention to?
- If no, is there a reason we dropped the previous `cudaStreamSynchronize(stream);` from this implementation?

SDL/Event.cu
```diff
@@ -640,57 +647,82 @@ public:
...
-        if (old != -1)
-            atomicMin(&hitsInGPU->hitRanges[lastModuleIndex * 2], ihit);
+                float ihit_x = hitsInGPU->xs[ihit];
+                float ihit_y = hitsInGPU->ys[ihit];
+                float ihit_z = hitsInGPU->zs[ihit];
+                int iDetId = hitsInGPU->detid[ihit];
+        
+                hitsInGPU->rts[ihit] = alpaka::math::sqrt(acc, ihit_x*ihit_x + ihit_y*ihit_y);
+                // This needs to get moved over after all kernels that use it have been moved.
+                hitsInGPU->phis[ihit] = SDL::phi(ihit_x,ihit_y,ihit_z);
```
VourMa, line 692: Probably unrelated to this PR but why do we need `z` in the calculation of `phi`?

SDL/Event.cu
GNiendorf: Isn't the hope that we'll be able to eventually run this algo on CPU's as well? It would be a pain to have to generalize the code in the future if we find that we need to take advantage of elements. Although it's unclear how useful elements are. See this issue for example: https://github.com/alpaka-group/alpaka/issues/1824

SDL/Event.cu
GNiendorf: Yes, wait is equivalent to stream synchronize. This PR only replaces the kernel code though. All other code still uses CUDA streams, which is why the Synchronizes are necessary.

SDL/Event.cu
GNiendorf: I don't think it's necessary, but this is probably something that should be fixed on master.

GNiendorf:
I removed the pixel segment kernel from this PR since more work should be done on that area of the code on master before moving it to Alpaka.

SDL/Event.cu
GNiendorf: Thanks for the suggestion, just pushed a commit to fix this.

GNiendorf:
Timing
Screenshot 2022-11-22 at 12 47 31 PM
Validation plots soon.

SDL/Event.cu
GNiendorf: Isn't the hope that we'll be able to eventually run this algo on CPU's as well? It would be a pain to have to generalize the code in the future if we find that we need to take advantage of elements. Although it's unclear how useful elements are.

GNiendorf:
Timing
Screenshot 2022-11-22 at 5 14 42 PM
Validation Plots
https://www.classe.cornell.edu/~gsn27/www/www/PR220_bf3cc2fDIRTY/#1.1

VourMa:
Thanks for the updates, @GNiendorf! It looks great now, merging!

PR #222, opened 2022-11-15
Clean up N_MAX memory

YonsiG:
as mentioned by title. There are some remaining codes regarding the N_MAX_MD, N_MAX_SG, N_MAX_T3, N_MAX_T5. These memories are already allocated dynamically, so these fixed value are no longer useful and create confusions.


YonsiG:
Timing performance
Screen Shot 2022-11-15 at 7 27 15 PM

YonsiG:
efficiency comparison https://www.classe.cornell.edu/~ygu/www/clean_up_N_MAX_memory/eff_comparison_plots__PU200_5d1c67a_explicit_cache_PU200_751bb75_explicit_cache/

SDL/Constants.h
```diff
@@ -52,16 +52,11 @@ typedef float FPX_seg;
 
 const unsigned int MAX_BLOCKS = 80;
 const unsigned int MAX_CONNECTED_MODULES = 40;
-const unsigned int N_MAX_MD_PER_MODULES = 89;
-const unsigned int N_MAX_SEGMENTS_PER_MODULE = 537; //Also used in (pixel)trackletDefualtAlgo in header files-> change there as well (hard coded)
 const unsigned int N_MAX_PIXEL_MD_PER_MODULES = 100000;
```
VourMa, line 55: Are these still needed?

SDL/Event.cu
```diff
@@ -762,9 +761,8 @@ void SDL::Event::addPixelSegmentToEvent(std::vector<unsigned int> hitIndices0,st
         segmentsInGPU = (SDL::segments*)cms::cuda::allocate_host(sizeof(SDL::segments), stream);
         //hardcoded range numbers for this will come from studies!
         unsigned int nTotalSegments;
-        createSegmentArrayRanges(*modulesInGPU, *rangesInGPU, *mdsInGPU, nLowerModules, nTotalSegments, stream, N_MAX_SEGMENTS_PER_MODULE, N_MAX_PIXEL_SEGMENTS_PER_MODULE);
-//        cout<<"nTotalSegments: "<<nTotalSegments<<std::endl; // for memory usage
-
+        //problem here: didn't distinguish pixel segments and outtracker segments. so they use the same memory index, which should be different and allocate dynamically
```
VourMa, line 764: Is this a problem of this PR or something general?

It's fine to leave a comment here for now but I think that creating an issue with more details would be more helpful in the long run.

SDL/Event.cu
```diff
@@ -1128,13 +1126,14 @@ void SDL::Event::createSegmentsWithModuleMap()
     if(segmentsInGPU == nullptr)
     {
         segmentsInGPU = (SDL::segments*)cms::cuda::allocate_host(sizeof(SDL::segments), stream);
-        createSegmentsInExplicitMemory(*segmentsInGPU, N_MAX_SEGMENTS_PER_MODULE, nLowerModules, N_MAX_PIXEL_SEGMENTS_PER_MODULE,stream);
+        unsigned int nTotalSegments;
+        createSegmentArrayRanges(*modulesInGPU, *rangesInGPU, *mdsInGPU, nLowerModules, nTotalSegments, stream, N_MAX_PIXEL_SEGMENTS_PER_MODULE);
```
VourMa, line 1130: This line was not here in the previous version of the code. Why did it have to appear now?

code/core/write_sdl_ntuple.cc
```diff
@@ -6070,20 +6057,14 @@ void printMDs(SDL::Event* event)
...
-            unsigned int UpperHitIndex = miniDoubletsInGPU.outerHitIndices[mdIdx];
-            unsigned int hit0 = hitsInGPU.idxs[LowerHitIndex];
-            unsigned int hit1 = hitsInGPU.idxs[UpperHitIndex];
-            std::cout <<  "VALIDATION 'MD': " << "MD" <<  " hit0: " << hit0 <<  " hit1: " << hit1 <<  std::endl;
-        }
-        for (unsigned int jdx = 0; jdx < miniDoubletsInGPU.nMDs[2*idx+1]; jdx++)
+        for (unsigned int iMD = 0; iMD < miniDoubletsInGPU.nMDs[idx]; iMD++)
         {
-            unsigned int mdIdx = (2*idx+1) * N_MAX_MD_PER_MODULES + jdx;
+            unsigned int mdIdx = rangesInGPU.miniDoubletModuleIndices[idx] + iMD;
```
VourMa, line 6067: Does this change mean that we were getting the MDs wrong in the output NTuple until now?

SDL/Constants.h
YonsiG: No, these two are not used anymore

SDL/Event.cu
YonsiG: I think it's the way that we define our objects, as we put the segments including both pixel segments and also outer segments. I'm thinking that it's not minor work if we want to separate the pixel segments and outer segments, so we may not want to do it in action term right now and put this into issue that we want to solve now. I think it should be a matter of redundant memory but not serious to algorithm. Maybe we just leave a comment here and if someone want to optimize it they can do later?

SDL/Event.cu
YonsiG: Because we want to get the nTotalSegments from this function. Then we can put this into createSegmentsInExplicitMemory

code/core/write_sdl_ntuple.cc
YonsiG: I'm not sure about do we get it correctly before... This is changing in a function called "printMDs", but we never use this function for a long time actually, I'm not sure even it's still callable. I'm fixing here just to make the index be consistent with the dynamic allocation index.

SDL/Constants.h
VourMa: I am not talking about the ones you deleted. The comment is on the next line.

Is `N_MAX_PIXEL_MD_PER_MODULES` and `N_MAX_PIXEL_SEGMENTS_PER_MODULE` still needed?

SDL/Event.cu
VourMa: So it's not a problem but rather an optimization? I don't remember discussing it recently, so I am a bit surprised it just came up in a comment at a random place in the code. If anything, I would bring it up to be discussed with everyone and understand if this is something to aim for the future (in which case we should make it an issue).

SDL/Event.cu
VourMa: But now I see that we also run this function in another place in the code (L765 of this file). Does it really make sense to "`createSegmentArrayRanges`" twice? Isn't there a way to get the information from the first running of the function?

YonsiG:
also, cuda-memcheck and nvprof works without issues

YonsiG:
these variables are cleaned for N_MAX md, sg, t3, t5 in the code. checked by using grep.

SDL/Constants.h
YonsiG: the N_MAX_Pixel is still needed, because memory related to pixel is not optimized and still used as static allocation

code/core/write_sdl_ntuple.cc
VourMa: Ok, I see, thanks!

Checking a bit, I understand this function is used for debugging. Do you know if it's a good state now, giving correct results?

slava77:
it looks like some recent updates triggered conflicts 
```
Conflicting files
code/core/write_sdl_ntuple.cc 
```

is this PR waiting for something else, like #224 ? (IIRC, it was waiting for #212 and #216 at some point earlier)

YonsiG:
No, this PR was only waiting for 212 and 216, not waiting for others. I'll fix the conflict and then update this. Thanks!

VourMa:
Hi Yanxi,
I am afraid that, with the latest update, you re-pushed a lot for previous commits, messing up the history and the changes of this PR. I would start from the master branch, `cherry-pick` the commits relevant only to this PR and then either make an n the PR or force-push the changes to this one. As it is, it's a bit hard to keep track of what has happened with your changes.

YonsiG:
Thanks Manos! I think you are right, I think I can close this PR and only cherry-pick the one I changed and make a new PR on this:)

SDL/Event.cu
YonsiG: Hi Manos, I think regarding this comment I put here, there's no need for us to bother now about further optimization if we don't have clear goal on how much memory we want to fit in. If we separate the pixel segments and outer tracker segments, we can further dynamically allocate memories of the two and save more. But I think only ~100MB was saved last time when I do cleanings on segments, so I'm just leaving a comment here, but not wishing to solve it in a short time :)

code/core/write_sdl_ntuple.cc
YonsiG: I also check this function a bit, it's called in bin/sdl.cc. In principal, if verbose==4, this function should be called. But now someone marked it as // TODO BROKEN // and commented it out for now (and I tested it's indeed broken). I'm afraid if we want to use that, a bit more study and fix need to be done...

YonsiG:
This PR is closed and another one replacing it will be pushed up soon

SDL/Event.cu
VourMa: I see. Then let's not call it "problem" then. Maybe optimization? Still, I would say that it makes more sense to not have comments on optimization at random points in the code but rather make an issue explaining with a few more words what the optimization can be and we can still have pointers to the code.

(By the way, please do "resolve" the conversations you reply to, without an indication that the other person has read them. GitHub collapses the comment threads and they can be easily missed. Apart from that, I still think that the conversation is not "resolved", since the comment, as it is written, is misleading and it is there also in the new PR.)

code/core/write_sdl_ntuple.cc
VourMa: Ok, then let's not worry about this. I see also that a related issue exists, so that's enough for us to have this in mind for the future.

SDL/Event.cu
YonsiG: Ok, I'll explain a bit in the code there and make an issue pointing to here. and sorry for resolving and collapsing the threads, won't do that again...

PR #224, opened 2022-12-02
LST in CMSSW

VourMa:
Zeroth version of a class, called LST, which can be called by CMSSW producer and run the full algorithm.

The input expected by the class is all the info we are getting from the tracking NTuple but per event, i.e. no preloading. The output that it spits out are the TC pT, η, φ, length and hit indices, to be consumed by downstream producers in the CMSSW configuration and create CMSSW track candidates.

The methods in the LST class are simplified versions of the functions that we run in the standalone scripts. For the class to run within CMSSW, some of these functions needed to be copied and possibly rewrittten, since they need to be included in `libsdl`. Starting from the needs of this PR, there is an addition to the Hits.cuh functions and a simplification of some other functions there.

Right now, the LST geometry is set up in each event within CMSSW. This is suboptimal and is expected to be fixed in the near future. However, I envision this development to go in a next PR.

Related CMSSW PR: SegmentLinking/cmssw#2

Related slides: https://cernbox.cern.ch/index.php/s/t7M0BfsyiEVN7eu

### DIFF
--- a/SDL/Hit.cuh
+++ b/SDL/Hit.cuh
@@ -66,60 +66,40 @@ namespace SDL
     CUDA_G void addHitToMemoryGPU(struct hits& hitsInCPU,struct modules& modulesInGPU,float x, float y, float z, unsigned int detId, unsigned int idxInNtuple,unsigned int moduleIndex, float phis,struct objectRanges& rangesInGPU);
     
     CUDA_HOSTDEV inline float ATan2(float y, float x) {
-    //if (x != 0) return  x * (float(-0xf.8eed2p-4) + x * x * float(0x3.1238p-4)); // degree 3 7 bit accuracy//atan2f(y, x);
-    if (x != 0){ 
-      //float a = y/x;
-      //float z = a*a;  
-      float test = atan2f(y, x);
-      //float test1 = a * (float(-0xf.8eed2p-4) + z * float(0x3.1238p-4)); 
-      //float test2 = a * (float(-0xf.ecfc8p-4) + z * (float(0x4.9e79dp-4) + z * float(-0x1.44f924p-4)));
-      //float test3 =  a * (float(-0xf.fcc7ap-4) + z * (float(0x5.23886p-4) + z * (float(-0x2.571968p-4) + z * float(0x9.fb05p-8))));
-      //float test4 = a * (float(-0xf.ff73ep-4) +
-      //        z * (float(0x5.48ee1p-4) +
-      //             z * (float(-0x2.e1efe8p-4) + z * (float(0x1.5cce54p-4) + z * float(-0x5.56245p-8)))));
-      //float test5 = a * (float(-0xf.ffff4p-4) +
-      //        z * (float(0x5.552f9p-4 + z * (float(-0x3.30f728p-4) +
-      //                                       z * (float(0x2.39826p-4) +
-      //                                            z * (float(-0x1.8a880cp-4) +
-      //                                                 z * (float(0xe.484d6p-8) +
-      //                                                      z * (float(-0x5.93d5p-8) + z * float(0x1.0875dcp-8)))))))));
-      //printf("%f %f %f %f %f %f %f %f\n",y,x,test,test1,test2,test3,test4,test5);
-      return test;
+      if (x != 0) return atan2f(y, x);
+      if (y == 0) return  0;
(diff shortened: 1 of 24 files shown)


SDL/LST.cc
```diff
@@ -0,0 +1,681 @@
...
+
+      const float ptAv_in = (trackCandidateType == 7 or trackCandidateType == 5) ? // pT3 or pT5
+        segmentsInGPU.ptIn[innerTrackletInnerSegmentIndex - rangesInGPU.segmentModuleIndices[*(modulesInGPU.nLowerModules)]] :
+        dr_in * k2Rinv1GeVf / fabs(sin((betaIn_in + betaOut_in) / 2.));
+
+      const float ptAv_out = dr_out * k2Rinv1GeVf / fabs(sin((betaIn_out + betaOut_out) / 2.));
+
+      float ptAv;
+      if(trackCandidateType == 7) { // pT5
+        float ptAv_outermost = dr_in * k2Rinv1GeVf / fabs(sin((betaIn_in + betaOut_in) / 2.));
```
VourMa, line 565: `sin`s in `fabs` to avoid negative pT.

If this fix is ok, it can be propagated to the sister function in the standalone.

SDL/LST.cc
```diff
@@ -0,0 +1,681 @@
...
+  int nSeeds=0;
+  for (auto &&[iSeed, _] : iter::enumerate(see_stateTrajGlbPx))
+  {
+      nSeeds++;
+      bool good_seed_type = false;
+      if (see_algo[iSeed] == 4) good_seed_type = true;
+      if (see_algo[iSeed] == 22) good_seed_type = true;
+      if (not good_seed_type) continue;
+
+      ROOT::Math::PxPyPzMVector p3LH(see_stateTrajGlbPx[iSeed], see_stateTrajGlbPy[iSeed], see_stateTrajGlbPz[iSeed],0);
```
VourMa, line 230: This function is rewritten using the latest physics vectors by ROOT. It can be used as an example to modernize the corresponding function in the standalone as well.

SDL/Hit.cuh
```diff
@@ -66,60 +66,40 @@ namespace SDL
...
+      if (y >  0) return  float(M_PI) / 2.f;
+      else        return -float(M_PI) / 2.f;
     }
-    if (y == 0) return  0;
-    if (y >  0) return  float(M_PI) / 2.f;
-    else        return -float(M_PI) / 2.f;
+    CUDA_HOSTDEV inline float eta(float x, float y, float z) {
+      float r3 = std::sqrt( x*x + y*y + z*z );
+      float rt = std::sqrt( x*x + y*y );
+      float eta = ((z > 0) - ( z < 0)) * acosh( r3 / rt );
```
slava77, line 77: `1/2 log( (p+pz)/(p-pz) )` as I recall we talked about missing hyperbolic functions in some (alpaka?) context

SDL/Hit.cuh
```diff
@@ -66,60 +66,40 @@ namespace SDL
...
-      //                                       z * (float(0x2.39826p-4) +
-      //                                            z * (float(-0x1.8a880cp-4) +
-      //                                                 z * (float(0xe.484d6p-8) +
-      //                                                      z * (float(-0x5.93d5p-8) + z * float(0x1.0875dcp-8)))))))));
-      //printf("%f %f %f %f %f %f %f %f\n",y,x,test,test1,test2,test3,test4,test5);
-      return test;
+      if (x != 0) return atan2f(y, x);
+      if (y == 0) return  0;
+      if (y >  0) return  float(M_PI) / 2.f;
+      else        return -float(M_PI) / 2.f;
```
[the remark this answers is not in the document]
VourMa: As mentioned in the context of the previous comment, in this PR I tried to only clean up some functions and not change their implementation. However, I agree that the part of code that you mention is needless. Checking the C++ `atan2f` implementation from `math.h`, it gives the same results as the behaviour we are trying to emulate here. Given this, I would just change `SDL::ATan2` to the standard `atan2f`. Maybe in another PR but it can be done in this one as well.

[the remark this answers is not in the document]
slava77: sure ,can be in a follow up pr

bsathian:
Merging this...

PR #227, opened 2022-12-30
Optional branches with CUT_VALUE_DEBUG

sgnoohc:


sgnoohc:
Created dummy variables for CUT_VALUE_DEBUG flags.
The flag can be enabled during `sdl_make_tracklooper` via the option `-d`.
Once enabled the LSTNtuple.root tree will contain some additional branches e.g. t5_dummy, tc_dummy, ...
This is just a template, so that each analyzers can add variables for these.
NOTE that the function to write these optional branches is separated from the main performance evaluating main output branches.
This is so that we don't interfere with the main function bloating it like last time with gazillion CUT_VALUE_DEBUG flags written all over the places.

VourMa (approved):
The PR looks good to me!

The only thing I don't understand is the mentions on parallelism in the comments of the `setOptionalOutputBranches` function but I guess it doesn't really matter.

Once @bsathian and/or @YonsiG verify that the way this is structured works for them, I would be happy to merge.

bsathian:
This is fine with me, I'll work on the efficiency part to add the lower object efficiencies and set up a PR for that

bsathian (approved):

PR #230, opened 2023-01-11
Lower object efficiency

bsathian:
Compiling with -d flag adds the relevant branches to the output ntuple and makes efficiency histograms and plots out of this ntuple. They won't be found in the summary page (as of now), and hence need to be accessed manually.
Examples:-
http://uaf-10.t2.ucsd.edu/~bsathian/temp_e297fbD-PU200/mtv/var/T5_lower_base_0_0_eff_pt.pdf
http://uaf-10.t2.ucsd.edu/~bsathian/temp_e297fbD-PU200/mtv/var/pT5_lower_base_0_0_eff_pt.pdf

http://uaf-10.t2.ucsd.edu/~bsathian/temp_e297fbD-PU200/mtv/var/T5_lower_duplrate_pt.pdf
http://uaf-10.t2.ucsd.edu/~bsathian/temp_e297fbD-PU200/mtv/var/pT5_lower_fakerate_pt.pdf


bsathian:
NOTE : The reason for the high T5 duplicate rate is that T5s before duplicate cleaning are filled in here. This decision was taken quite recently (in light of T5 studies) - (old cod ref : https://github.com/bsathian/TrackLooper-1/blame/T5_bugFixes/code/core/write_sdl_ntuple.cc#L1024). Hence, a comparison with the "ORIGINAL" might not be possible.

sgnoohc:
efficiency/bin/createEffNumDenHists binary file should not be included in PR

sgnoohc:
I do not like whatever this n_accepted_simtrk is doing.
this kind of prolific copy-paste all over the place is what's gonna cause a giant headache in the future.
Can you explain to me what this is?

sgnoohc:
Don't change the Makefile and propagate another flag to efficiency/Makefile but rather change the performance.cc to accept the LSTNtuple.root as is, and decide based on what branches are available.

sgnoohc:
[efficiency/python/make_classfiles.py.bak](https://github.com/SegmentLinking/TrackLooper/pull/230/files#diff-fc00d6de0ddb5a21fa15d7ea038e0cbc51b7da5564f9896d83a2b05cd52d5b14) should not be included.

VourMa:
Apart from the comments by Philip, with which I agree (especially the one about the `n_accepted_simtrk` variable - I didn't understand it either when I checked the PR), I think we should clean out the commented lines, e.g. in the Makefile or the lst_plot_performance.py. They turn out to be confusing in the long run, if the reason why they are needed is not stated.

bsathian:
Updated with Philip's stuff

VourMa (approved):
After the latest updates, the PR looks good. I am merging it, since it is needed for the incoming developments.

PR #233, opened 2023-01-20
Extension purge

bsathian:
Purging the Track Extension code


SDL/Event.cu
```diff
@@ -1227,19 +1183,16 @@ void SDL::Event::createTrackCandidates()
         createTrackCandidatesInExplicitMemory(*trackCandidatesInGPU, N_MAX_TRACK_CANDIDATES + N_MAX_PIXEL_TRACK_CANDIDATES,stream);
     }
 
-#ifdef FINAL_pT3
```
VourMa, line 1230: Apart from the extension purging, I see that you make some more cleanup of a bunch of compilation flags, which makes a lot of sense. I have some comments:
1. It seems like you removed every trace of `FINAL_pT3` from the code. However, the flag seems to be there in the `SDL/Makefile`. Can you remove it from there as well? Same for the rest of the compilation flags that are completely removed.
2. I can see an instance of `FINAL_pT5` that serves no real purpose. Can you remove that as well? Same for `FINAL_T5`.

bsathian:
Screenshot from 2023-01-27 09-53-41

bsathian:
Screenshot from 2023-01-27 10-13-11

VourMa (approved):
Even if not obvious (some of the changes are hidden within the merge commit), all of the comments are resolved, so this is good to merge.

PR #240, opened 2023-02-09
:city_sunrise::crocodile: GNN Ntupling (feat. setup script for UF HiPerGator)

sgnoohc:
Added the following:
MiniDoublets: dphichange is now being read out (changes in the SDL/Event.cu)
Hits: detid is now being read out (changes in the SDL/Event.cu)

bin/sdl:
added -G option to enable GNN input variable branches to the output

AccessHelper:
Added some helper function for hitTypes/Index retrieval in one go

AnalysisConfig.h:
added gnn_ntuple boolean flag

trkCore:
minor helper function

write_sdl_ntuple:
GnnNtuple create/set/fill branches

setup_hpg.sh:
University of Florida HiPerGator cluster setup

### DIFF
--- a/bin/sdl.cc
+++ b/bin/sdl.cc
@@ -54,6 +54,7 @@ int main(int argc, char** argv)
         ("d,debug"           , "Run debug job. i.e. overrides output option to 'debug.root' and 'recreate's the file.")
         ("c,cpu"             , "Run CPU version of the code.")
         ("l,lower_level"     , "write lower level objects ntuple results")
+        ("G,gnn_ntuple"      , "write gnn input variable ntuple")
         ("j,nsplit_jobs"     , "Enable splitting jobs by N blocks (--job_index must be set)", cxxopts::value<int>())
         ("I,job_index"       , "job_index of split jobs (--nsplit_jobs must be set. index starts from 0. i.e. 0, 1, 2, 3, etc...)", cxxopts::value<int>())
         ("h,help"            , "Print help");
@@ -240,6 +241,17 @@ int main(int argc, char** argv)
         ana.do_lower_level = false;
     }
 
+    //_______________________________________________________________________________
+    // --gnn_ntuple
+    if (result.count("gnn_ntuple"))
+    {
+        ana.gnn_ntuple = true;
+    }
+    else
+    {
+        ana.gnn_ntuple = false;
+    }
+
     // Printing out the option settings overview
     std::cout << "=========================================================" << std::endl;
     std::cout << " Setting of the analysis job based on provided arguments " << std::endl;
@@ -292,6 +304,11 @@ void run_sdl()
     {
         createOutputBranches();
     }
+    if (ana.gnn_ntuple)
+    {
+        createGnnNtupleBranches();
+    }
+
 
     std::vector<std::vector<float>> out_trkX;
     std::vector<std::vector<float>> out_trkY;
@@ -452,7 +469,7 @@ void run_sdl()
                 }
             }
 
-            if (ana.do_write_ntuple)
+            if (ana.do_write_ntuple or ana.gnn_ntuple)
             {
                 #pragma omp critical
                 {
@@ -482,7 +499,7 @@ void run_sdl()
 
     SDL::cleanModules();
 
-    if (ana.do_write_ntuple)
+    if (ana.do_write_ntuple or ana.gnn_ntuple)
     {
         // Writing ttree output to file
         ana.output_tfile->cd();
--- a/bin/sdl.cc
+++ b/bin/sdl.cc
@@ -246,6 +246,13 @@ int main(int argc, char** argv)
     if (result.count("gnn_ntuple"))
     {
         ana.gnn_ntuple = true;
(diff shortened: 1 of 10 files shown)


sgnoohc:
image

From A100 on HiPerGator

VourMa:
@sgnoohc Can you please also run the master on the same machine and post the timing for comparison?

Also, since this PR involves the addition of a new variable, can you please perform the checks documented [here](https://github.com/SegmentLinking/TrackLooper/wiki/Line-Segment-Tracking-Algorithm#debugging) to make sure that nothing has been missed?

I will take a more detailed look at the code by our meeting tomorrow.

bin/sdl.cc
```diff
@@ -452,7 +469,7 @@ void run_sdl()
                 }
             }
 
-            if (ana.do_write_ntuple)
+            if (ana.do_write_ntuple or ana.gnn_ntuple)
```
VourMa, line 472: I can see an "or" here, meaning that one can in principle write out an NTuple with only the GNN branches. The changes in lines L307-L311 of the new version of this file corroborate this assumption. However, when I look at L13 of the `write_sdl_ntuple.cc` file, it seems like the regular branches are filled in any case?

Maybe I misinterpret the code but have the different combination of the `do_write_ntuple` and `gnn_ntuple` options been tested and work fine? And, ultimately, what is the intended behavior? Do we want the option to write out only the GNN NTuple and not the rest?

code/core/AccessHelper.cc
```diff
@@ -1,5 +1,26 @@
 #include "AccessHelper.h"
 
+// ===============
+// ----* Hit *----
+// ===============
+
+//____________________________________________________________________________________________
+std::tuple<std::vector<unsigned int>, std::vector<unsigned int>> convertHitsToHitIdxsAndHitTypes(SDL::Event* event, std::vector<unsigned int> hits)
```
VourMa, line 8: Are these functions intended to run also as part of standalone code that produces an output to be used for CMSSW or they are just there for the purpose of the NTuple? If the former is the case, then we should consider adding them to the CMSSW class, with the proper adjustments to reduce host-device memory copies. Maybe it's too early to know that but, in any case, it's good to keep in mind that something additional may be needed here in the future.

bin/sdl.cc
sgnoohc: I added so that do_write_ntuple must be set to 1 when gnn_ntuple is to be set.
The "or" statements are no longer necessary because of it.

code/core/AccessHelper.cc
sgnoohc: This is only for ntupling. afaik none of the functions in AccessHelper are not used for main results. (are they..?)

sgnoohc:
> @sgnoohc Can you please also run the master on the same machine and post the timing for comparison?
> 
> Also, since this PR involves the addition of a new variable, can you please perform the checks documented [here](https://github.com/SegmentLinking/TrackLooper/wiki/Line-Segment-Tracking-Algorithm#debugging) to make sure that nothing has been missed?
> 
> I will take a more detailed look at the code by our meeting tomorrow.

Total Timing Summary
   avg      3.4      2.2      1.0      3.5      3.1      0.7      1.1      0.8      1.9      17.8      13.6+/-  2.9      18.9   explicit_cache[s=1]
   avg      4.8      4.3      2.2      5.2      4.7      1.6      2.8      2.3      4.4      32.5      26.0+/-  9.7      18.9   explicit_cache[s=2]
   avg     10.2      5.0      4.1     13.4     14.5      2.5      4.9      4.4      4.9      63.9      51.1+/- 20.6      19.0   explicit_cache[s=4]
   avg     15.5      6.7      4.9     20.1     21.7      5.1      6.8      2.2      8.3      91.4      70.8+/- 35.2      19.1   explicit_cache[s=6]
   avg     30.1     10.5      6.9     27.8     29.2      3.6     10.7      5.5     13.4     137.7     104.0+/- 49.5      19.3   explicit_cache[s=8]

The memcheck returned 0 error

code/core/AccessHelper.cc
VourMa: They are. If we want to get the output of hit indices, pt, eta, etc. for usage in CMSSW, a modified version of some of them is used. The modifications were necessary to reduce device-host memory transfers. You can take a look at [SDL::LST::getOutput](https://github.com/SegmentLinking/TrackLooper/blob/c4551a751b9fef5d6261ff6974a5910a5ace0e7c/SDL/LST.cc#L369) and [Object accessors](https://github.com/SegmentLinking/TrackLooper/blob/c4551a751b9fef5d6261ff6974a5910a5ace0e7c/SDL/LST.cc#L525).

But, in any case, if this meant on for NTupling for now, it is ok and we can revisit it in the future.

VourMa (approved):
Thanks for the changes and the checks. It all looks good now. I am merging.

Something that is probably worth looking into in the future is the behavior of multistreaming on the A100, which is different from what we had seen up till now on other GPUs. I think this is an interesting point to investigate and understand, so I created issue #241 to take a note of it.

PR #242, opened 2023-02-20
Gnn ntuple v2

sgnoohc:
- setup_hpg.sh was updated to point to /blue storage for common files
- ```tc_lsIdx``` was added to account for line segments used in track candidates.

The ```tc_lsIdx``` will be used for future studies where the duplicate removal is removed and then check for efficiency drops when applying GNN score cuts on the line segments.


VourMa (approved):
The PR looks good to me. If it has been tested to work for your use case, I would be happy to merge it.

sgnoohc:
Yes it works for me, please merge them for me.
(I like this added layer of scrutiny of not being able to merge by myself.)

Dr. P. Chang, Asst. Prof., University of Florida
***@***.*** ***@***.***)

PR #244, opened 2023-02-22
include <limits> to cover usage of std::numeric_limits

slava77:
without this the compilation failed on cgpu1 (el8) on top of CMSSW_13_0_0_pre4


slava77:
I was editing the file for this PR on the web and was still offered (by default) to push directly to `master`. I thought that this was supposed to be disabled.
@bsathian

tresreid (approved):

PR #248, opened 2023-02-27
bug fix in T3

YonsiG:
a stale bug in the T3 code, with incorrect assignment of inner module index.
performance plots doesn't have obvious changes after fixing
master: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/T3bug_fix/master_16dc15-PU200/
after fix: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/T3bug_fix/fix_16dc15D-PU200/
comparison to master: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/T3bug_fix/comparison_16dc15-PU200_16dc15D-PU200/

### DIFF
--- a/SDL/Triplet.cu
+++ b/SDL/Triplet.cu
@@ -532,7 +532,7 @@ __device__ bool SDL::passPointingConstraintBBE(struct SDL::modules& modulesInGPU
     if(not pass) return pass;
 
     float dLum = copysignf(SDL::deltaZLum, zIn);
-    bool isOutSgInnerMDPS = modulesInGPU.moduleType[outerOuterLowerModuleIndex] == SDL::PS;
+    bool isOutSgInnerMDPS = modulesInGPU.moduleType[middleLowerModuleIndex] == SDL::PS;
     float rtGeom1 = isOutSgInnerMDPS ? SDL::pixelPSZpitch : SDL::strip2SZpitch;
     float zGeom1 = copysignf(zGeom,zIn);
     float rtLo = rtIn * (1.f + (zOut - zIn - zGeom1) / (zIn + zGeom1 + dLum) / dzDrtScale) - rtGeom1; //slope correction only on the lower end
@@ -623,7 +623,7 @@ __device__ bool SDL::passPointingConstraintEEE(struct SDL::modules& modulesInGPU
     pass = pass and ((rtOut >= rtLo) & (rtOut <= rtHi));
     if(not pass) return pass;
     
-    bool isInSgOuterMDPS = modulesInGPU.moduleType[outerOuterLowerModuleIndex] == SDL::PS;
+    bool isInSgOuterMDPS = modulesInGPU.moduleType[middleLowerModuleIndex] == SDL::PS;
 
     float drOutIn = rtOut - rtIn;
     float drtSDIn = rtMid - rtIn;


YonsiG:
timing after this commit:
Screen Shot 2023-03-01 at 3 05 41 PM

timing for master
Screen Shot 2023-03-01 at 2 49 23 PM

YonsiG:
This bug is found in function passPointingConstraintBBE(also used for BEE T3) and passPointingConstraintEEE in SDL/triplet.cu.  The Boolean variable isOutSgInnerMDPS is used for judging whether the middle MD of a T3 is PS or 2S. In BBE function, this variable is then put into deciding rtGeom1 variable, which is a cut on rt of T3. In EEE function, it’s deciding whether we do an additional rt cut on this T3.

YonsiG:
verbose level2 print out how many T3 are created, (here layer1-2-3 doesn't mean the T3 is using layer1,2,3, it just means this T3 starts from layer 1)
This is T3 created from the the master
Screen Shot 2023-03-01 at 3 44 05 PM
This is after the PR
Screen Shot 2023-03-01 at 3 46 51 PM

VourMa:
I might have solved the mystery of this bug and its corresponding bug fix.

As Yanxi mentions, this bug shows up in the `passPointingConstraintBBE` and `passPointingConstraintEEE` functions. However, when these functions are run ([here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L422) and [here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L436)), also the functions `runTripletDefaultAlgoBBBB`, `runTripletDefaultAlgoBBEE` and `runTripletDefaultAlgoEEEE` are run ([here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1375), [here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1384) and [here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1393)).

Two of these functions, `runTripletDefaultAlgoBBEE` and `runTripletDefaultAlgoEEEE`, have redundant cuts wrt. `passPointingConstraintBBE` and `passPointingConstraintEEE`. As an example, compare [this cut in `runTripletDefaultAlgoBBEE`](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L963) with [the corresponding one in `passPointingConstraintBBE`](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L551) and [this cut in `runTripletDefaultAlgoEEEE`](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1177) with [the corresponding one in `passPointingConstraintEEE`](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L623).

Note that even though the bug fix changes whatever happens in `passPointingConstraintBBE` and `passPointingConstraintEEE`, it does nothing for the `runTripletDefaultAlgoBBEE` and `runTripletDefaultAlgoEEEE` functions ([here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L944) and [here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1161)). And the latter seem to be correct. As a result, I think that, before the bug fix were cutting looser at the first pass of the cut (as were using 2S pitch for PS modules) and things were properly taken care by the the second pass of the cut, while now we cut properly at the first pass already. This explanation fits the observation that the bug fix does nothing - we have redundant cuts in this part of the code.

However, I am not sure what happens in the `runTripletDefaultAlgoBBBB` case, which runs in the BBE case ([here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1375)), which means that it sometimes runs together with the `passPointingConstraintBBE` function ([here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L422)). To me, it seems like `runTripletDefaultAlgoBBBB` doesn't include the redundant cut, so the bug fix should have an effect in those cases. Maybe BBE with 2S module in the E layer and a PS module in the 2nd B layer are very rare, which I find very plausible, given the tracker geometry and the fact that we do not skip layers in the module map.

To sum up, based on the discussion above, I think it makes sense that we didn't see any change after the bug fix. I guess that the proper bug fix would be to get rid of the redundant cuts. A homogenization of the use cases of `passPointingConstraintXXX` and `runTripletDefaultAlgoXXXX` would also be useful. Finally, I would like to note that overloading the `runTripletDefaultAlgo` ([here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L309) and [here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Triplet.cu#L1338)) when it has so many arguments is highly confusing.

YonsiG:
Hi Manos, thanks a lot for these investigation about the two cuts! I read the parts in the passPointingConstraintXXX function and the runTripletDefaultAlgoXXXX function and agree that we can remove the redundant cuts. But I'm wondering why in both BBB and BBE case we run the runTripletDefaultAlgoBBBB? and also, the BBEE category seems redundant, if we require 2LS share the same MD, the innerSgOuterMD should be the outerSgInnerMD.

VourMa:
> But I'm wondering why in both BBB and BBE case we run the runTripletDefaultAlgoBBBB?

I am not sure about this one. Maybe @bsathian or @tresreid can comment? Otherwise we would need to understand the details of the code and compare with how the last layer is used in the EEEE case. And that is partially why I said that we should probably rationalize the use cases for these functions.

> and also, the BBEE category seems redundant, if we require 2LS share the same MD, the innerSgOuterMD should be the outerSgInnerMD.

You are right, this must be a remnant of the previous algorithm version, when no common MD was required. Are we ready to forgo it? I would say yes but others should comment as well. Good point for discussion.

sgnoohc:
It may have been some kind of study at the time that indicated that this worked better...
There were some subtle things like that where based on empirical evidence certain things worked out better.
It's unfortunate that the comments are not very well written as to understand why certain things are as is.

Philip


Dr. P. Chang, Asst. Prof., University of Florida
***@***.***

________________________________

Issue #249, opened 2023-03-01
pt estimate for T5s in parseT5 function is deprecated


jkguiang:
The pT for T5s in the output NTuple [are computed](https://github.com/jkguiang/TrackLooper/blob/master/code/core/write_sdl_ntuple.cc#L410) as follows:
```cpp
float pt = quintupletsInGPU.innerRadius[quintupletIndex] * kRinv1GeVf;
...
ana.tx->pushbackToBranch<float>("t5_pt", pt);
```
However, the [pT estimate](https://github.com/jkguiang/TrackLooper/blob/master/code/core/write_sdl_ntuple.cc#L1041-L1042) in ```parseT5``` is just the average of the T3 pT's (which is somewhat questionable per #214):
```cpp
// Compute pt estimates from inner and outer triplets
const float ptAv_in = abs(dr_in * k2Rinv1GeVf / sin((betaIn_in + betaOut_in) / 2.));
const float ptAv_out = abs(dr_out * k2Rinv1GeVf / sin((betaIn_out + betaOut_out) / 2.));

// T5 pt is average of the two pt estimates
const float pt = (ptAv_in + ptAv_out) / 2.;
```
It's my understanding that the pT as computed for the T5s in the output NTuple are "more correct," and that both the pTs from the output NTuple and ```parseT5``` are used and therefore giving inconsistent results. I am working on a PR (will include link here when available) that will add some additional branches for the GNN NTuple, so I can make the fix in that PR as well if desired.

VourMa:
@jkguiang Is this still an issue? Did you deal with it in any of your PRs,as mentioned in the description?

VourMa:
Closed by #388

PR #250, opened 2023-03-03
Added T3/T5 branches for GNN T5 classification

jkguiang:
Added GNN NTuple branches for T5 classification plus some minor fixes to Event.cu

### DIFF
--- a/code/core/write_sdl_ntuple.cc
+++ b/code/core/write_sdl_ntuple.cc
@@ -190,14 +190,17 @@ void createGnnNtupleBranches()
     ana.tx->createBranch<vector<float>>("t3_eta");
     ana.tx->createBranch<vector<float>>("t3_phi");
     ana.tx->createBranch<vector<float>>("t3_0_r");
+    ana.tx->createBranch<vector<float>>("t3_0_dr");
     ana.tx->createBranch<vector<float>>("t3_0_x");
     ana.tx->createBranch<vector<float>>("t3_0_y");
     ana.tx->createBranch<vector<float>>("t3_0_z");
     ana.tx->createBranch<vector<float>>("t3_2_r");
+    ana.tx->createBranch<vector<float>>("t3_2_dr");
     ana.tx->createBranch<vector<float>>("t3_2_x");
     ana.tx->createBranch<vector<float>>("t3_2_y");
     ana.tx->createBranch<vector<float>>("t3_2_z");
     ana.tx->createBranch<vector<float>>("t3_4_r");
+    ana.tx->createBranch<vector<float>>("t3_4_dr");
     ana.tx->createBranch<vector<float>>("t3_4_x");
     ana.tx->createBranch<vector<float>>("t3_4_y");
     ana.tx->createBranch<vector<float>>("t3_4_z");
@@ -729,20 +732,21 @@ void setGnnNtupleBranches(SDL::Event* event)
                                               outer T3
             */
             unsigned int T5_idx = rangesInGPU.quintupletModuleIndices[idx] + jdx;
+
             std::vector<unsigned int> T3s = getT3sFromT5(event, T5_idx);
             // Inner T3
             if (T3s_used_in_T5.find(T3s[0]) == T3s_used_in_T5.end())
             {
                 T3s_used_in_T5.insert(T3s[0]);
                 T3_index_map[T3s[0]] = T3s_used_in_T5.size() - 1;
-                setGnnNtupleTriplet(event, T3s[0]);
+                setGnnNtupleTriplet(event, T3s[0], rangesInGPU.tripletModuleIndices[idx]);
             }
             // Outer T3
             if (T3s_used_in_T5.find(T3s[1]) == T3s_used_in_T5.end())
             {
                 T3s_used_in_T5.insert(T3s[1]);
                 T3_index_map[T3s[1]] = T3s_used_in_T5.size() - 1;
-                setGnnNtupleTriplet(event, T3s[1]);
+                setGnnNtupleTriplet(event, T3s[1], rangesInGPU.tripletModuleIndices[idx]);
             }
 
             ana.tx->pushbackToBranch<int>("t5_t3_idx0", T3_index_map[T3s[0]]);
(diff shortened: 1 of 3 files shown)


jkguiang:
Let me know if I should address #249 in this PR as well. I think the fix is easy.

code/core/write_sdl_ntuple.cc
```diff
@@ -782,6 +843,195 @@ void setGnnNtupleMiniDoublet(SDL::Event* event, unsigned int MD)
...
+        short module_side = modulesInGPU.sides[module];
+        float module_drdz = modulesInGPU.drdzs[module];
+        float module_slope = modulesInGPU.slopes[module];
+        // Get deltas for sigma calculation
+        float delta1, delta2;
+        bool is_flat;
+        // Category 1: barrel PS flat
+        if (module_subdet == SDL::Barrel and module_type == SDL::PS and module_side == SDL::Center)
+        {
+            delta1 = inv1;//1.1111f;//0.01;
```
VourMa, line 916: Commented out numbers should be deleted? The first one seems to be the result of `inv1` (so why have it re-written), while the second one is random - at least I can't understand where it comes from...

This comment applies to similar instances below.

code/core/write_sdl_ntuple.cc
```diff
@@ -782,6 +843,195 @@ void setGnnNtupleMiniDoublet(SDL::Event* event, unsigned int MD)
...
+        }
+        // Category 4: endcap PS
+        else if (module_subdet == SDL::Endcap and module_type == SDL::PS)
+        {
+            delta1 = inv1;//1.1111f;//0.01;
+            is_flat = false;
+
+            /* Despite the type of the module layer of the lower module index,
+             * all anchor hits are on the pixel side and all non-anchor hits are
+             * on the strip side! */
```
VourMa, line 946: Is this comment relevant here? If it is, probably its connections should be made clearer?

code/core/write_sdl_ntuple.cc
```diff
@@ -782,6 +843,195 @@ void setGnnNtupleMiniDoublet(SDL::Event* event, unsigned int MD)
...
+        sigmas.push_back(2 * sqrtf((xPrime * delta1) * (xPrime * delta1) + (yPrime * delta2) * (yPrime * delta2)));
+    }
+
+    ana.tx->pushbackToBranch<float>("t3_0_dr", sigmas[0]);
+    ana.tx->pushbackToBranch<float>("t3_2_dr", sigmas[1]);
+    ana.tx->pushbackToBranch<float>("t3_4_dr", sigmas[2]);
+
+    // Constants
+    const float kRinv1GeVf = (2.99792458e-3 * 3.8);
+    const float k2Rinv1GeVf = kRinv1GeVf / 2.;
```
VourMa, line 1005: Should we just include the constants from [here](https://github.com/SegmentLinking/TrackLooper/blob/579b9d08bfb5d774c6ce3727e7625a4620faf1ac/SDL/Constants.h#L78)?

code/core/write_sdl_ntuple.cc
```diff
@@ -782,6 +843,195 @@ void setGnnNtupleMiniDoublet(SDL::Event* event, unsigned int MD)
...
+
+    // Radial distance
+    const float dr = sqrt(pow(hitsInGPU.xs[hit4] - hitsInGPU.xs[hit0], 2) + pow(hitsInGPU.ys[hit4] - hitsInGPU.ys[hit0], 2));
+
+    // Beta angles
+    float betaIn  = __H2F(tripletsInGPU.betaIn[T3]);
+    float betaOut = __H2F(tripletsInGPU.betaOut[T3]);
+
+    // Legacy T4 pt estimate
```
VourMa, line 1014: T4? I guess this is confusing, because we don't have T4s now?

code/core/write_sdl_ntuple.cc
```diff
@@ -1042,7 +1292,7 @@ std::tuple<float, float, float, vector<unsigned int>, vector<unsigned int>> pars
     const float ptAv_out = abs(dr_out * k2Rinv1GeVf / sin((betaIn_out + betaOut_out) / 2.));
 
     // T5 pt is average of the two pt estimates
-    const float pt = (ptAv_in + ptAv_out) / 2.;
+    const float pt = (ptAv_in + ptAv_out) / 2.;     // this is deprecated, c.f. setQuintupletOutputBranches
```
VourMa, line 1295: Not sure I understand the comment. Is this used? If not, maybe we should delete it and use the up-to-date version?

VourMa (commented):
Since two new variables are added (are they used anywhere, by the way?), could you please run the checks [here](https://github.com/SegmentLinking/TrackLooper/wiki/Line-Segment-Tracking-Algorithm#debugging) to be sure that everything checks out?

sgnoohc:
general comment.
Is there a reason why not to put idxs to the relevant LS in the t3 branches?
The flattening of the t3 hit positions are OK in principle.
But it puts a danger in creating a lot of branches and lots of duplicate information.
(e.g. if two T3's share a LS, it's going to copy paste same information twice.)

I would rather do something like
T3_LS_idx0
keeping the same branch philosophy as the LS.

code/core/write_sdl_ntuple.cc
```diff
@@ -181,6 +181,30 @@ void createGnnNtupleBranches()
     ana.tx->createBranch<vector<float>>("LS_sim_vz");
     ana.tx->createBranch<vector<int>>("LS_isInTrueTC");
 
+    // T3 branches
+    ana.tx->createBranch<vector<int>>("t5_t3_idx0");
+    ana.tx->createBranch<vector<int>>("t5_t3_idx1");
```
sgnoohc, line 186: parallelism is broken here.
Technically these are t5 branches.
The comment //T3 branches showing right above this is misleading

jkguiang:
> Is there a reason why not to put idxs to the relevant LS in the t3 branches?

We use the three hits stored in the t3 branches for computing Balaji's Chi^2. Storing them directly makes it very easy to use them as inputs to the GNN, rather than requiring nested lookups (first to the LS, then to the MD). I also figure that there is not much "danger" here for now, as this is just for training the GNN.

> I would rather do something like
T3_LS_idx0
keeping the same branch philosophy as the LS.

This is doable if necessary, but would be more of a headache as explained above.

code/core/write_sdl_ntuple.cc
jkguiang: What do you mean by "parallelism is broken"? The comment is a typo and will be fixed with the responses to Manos's comments.

PR #256, opened 2023-03-20
Remove Tracklet and Pixeltracklet Files

GNiendorf:
The functions in these two files were not used anywhere in the code. I moved one of the constants, pt_betaMax, to a better location in Constants.cuh since it is currently defined in one of the files I deleted.

No timing changes obviously since none of the functions were being used
Screenshot 2023-03-20 at 1 27 26 PM


VourMa (approved):
Thanks for the clean-up, @GNiendorf, I am merging this.

Since it may get messy this time round, let me know if you need a second pair of eyes when rebasing the `alpaka_move` to `master` and I can take a look.

Issue #223, opened 2022-11-22
Issues with Pixel Segment Code


GNiendorf:
1. Variables are not directly copied to the correct location. They are first allocated with temporary device variables, and then manually copied in the kernel (addPixelSegmentToMemory device function).
https://github.com/SegmentLinking/TrackLooper/blob/cff3eda1228855233f5be55620aee06c0d3464b5/SDL/Event.cu#L830-L846

2. SegmentsInGPU is allocated with a large constant amount of space, instead of using the correct "size" variable.
https://github.com/SegmentLinking/TrackLooper/blob/cff3eda1228855233f5be55620aee06c0d3464b5/SDL/Event.cu#L765-L768

3. createMDArrayRanges and createSegmentArrayRanges both make the mistake of pulling variables from the GPU, computing some quantities on the CPU, and then pushing the results back to the GPU instead of just calculating everything directly on the GPU.
https://github.com/SegmentLinking/TrackLooper/blob/4015cfcb67b215aee3eaf990e4ab4d3135392c1f/SDL/Segment.cu#L32-L103

GNiendorf:
Issue 3 is now fixed by #272, thank you @tresreid!

VourMa:
...And I believe that issue 2 is a necessary evil, as showcased in PR #342. Should we close this? @GNiendorf

PR #260, opened 2023-03-25
Move PixelSegmentToEvent Kernel to Alpaka

GNiendorf:
Keeping this as a draft pull request for now since I want to fix some of the issues mentioned in #223 before merging.


SDL/Event.cu
```diff
@@ -722,170 +719,151 @@ void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::
...
+        {
+            unsigned int innerMDIndex = rangesInGPU.miniDoubletModuleIndices[pixelModuleIndex] + 2*(tid);
+            unsigned int outerMDIndex = rangesInGPU.miniDoubletModuleIndices[pixelModuleIndex] + 2*(tid) +1;
+            unsigned int pixelSegmentIndex = rangesInGPU.segmentModuleIndices[pixelModuleIndex] + tid;
+
+            addMDToMemory(mdsInGPU, hitsInGPU, modulesInGPU, hitIndices0[tid], hitIndices1[tid], pixelModuleIndex, 0,0,0,0,0,0,0,0,0,innerMDIndex);
+            addMDToMemory(mdsInGPU, hitsInGPU, modulesInGPU, hitIndices2[tid], hitIndices3[tid], pixelModuleIndex, 0,0,0,0,0,0,0,0,0,outerMDIndex);
+
+            //in outer hits - pt, eta, phi
+            float slope = sinhf(hitsInGPU.ys[mdsInGPU.outerHitIndices[innerMDIndex]]);
```
VourMa, line 758: I think we had discussed that this function does not exist in Alpaka. Was there any development on this?

Still, if not, I think this should be made alpaka-compatible by expanding it and using the relevant alpaka versions of the involved functions.

SDL/Event.cu
```diff
@@ -722,170 +719,151 @@ void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::
...
 {
+    const int size = ptIn.size();
+    unsigned int mdSize = 2 * size;
+    uint16_t pixelModuleIndex = (*detIdToIndex)[1];
+
     if(mdsInGPU == nullptr)
     {
         mdsInGPU = (SDL::miniDoublets*)cms::cuda::allocate_host(sizeof(SDL::miniDoublets), stream);
-        //hardcoded range numbers for this will come from studies!
+        // hardcoded range numbers for this will come from studies!
```
VourMa, line 782: This has been done, right? @YonsiG ?

SDL/Event.cu
```diff
@@ -722,170 +719,151 @@ void SDL::Event::addHitToEvent(std::vector<float> x, std::vector<float> y, std::
...
 
         cudaMemcpyAsync(mdsInGPU->nMemoryLocations, &nTotalMDs, sizeof(unsigned int), cudaMemcpyHostToDevice, stream);
         cudaStreamSynchronize(stream);
-
     }
     if(segmentsInGPU == nullptr)
     {
         segmentsInGPU = (SDL::segments*)cms::cuda::allocate_host(sizeof(SDL::segments), stream);
-        //hardcoded range numbers for this will come from studies!
+        // hardcoded range numbers for this will come from studies!
```
VourMa, line 793: This has been done, right? @YonsiG ?

SDL/Event.cu
```diff
@@ -2162,7 +2140,7 @@ SDL::segments* SDL::Event::getSegments()
         segmentsInCPU->phi = new float[N_MAX_PIXEL_SEGMENTS_PER_MODULE];
         segmentsInCPU->seedIdx = new unsigned int[N_MAX_PIXEL_SEGMENTS_PER_MODULE];
         segmentsInCPU->isDup = new bool[N_MAX_PIXEL_SEGMENTS_PER_MODULE];
-        segmentsInCPU->isQuad = new bool[N_MAX_PIXEL_SEGMENTS_PER_MODULE];
+        segmentsInCPU->isQuad = new short[N_MAX_PIXEL_SEGMENTS_PER_MODULE];
```
VourMa, line 2143: Why the change of type here (and also below)? `bool` seems to make more sense, no?

SDL/Segment.cu
```diff
@@ -28,7 +27,8 @@ void SDL::segments::resetMemory(unsigned int nMemoryLocationsx, unsigned int nLo
 void SDL::createSegmentArrayRanges(struct modules& modulesInGPU, struct objectRanges& rangesInGPU, struct miniDoublets& mdsInGPU, uint16_t& nLowerModules, int& nTotalSegments, cudaStream_t stream, const uint16_t& maxPixelSegments)
 {
     /*
-        write code here that will deal with importing module parameters to CPU, and get the relevant occupancies for a given module!*/
+        write code here that will deal with importing module parameters to CPU, and get the relevant occupancies for a given module!
```
VourMa, line 30: What is this comment supposed to mean? @YonsiG or @tresreid know maybe?

SDL/Segment.cuh
```diff
@@ -324,9 +324,71 @@ namespace SDL
...
+        /*
+        The two anchor hits are r3PCA and r3LH. p3PCA pt, eta, phi is hitIndex1 x, y, z
+        */
+        float circleRadius = mdsInGPU.outerX[innerMDIndex] / (2 * k2Rinv1GeVf);
+        float circlePhi = mdsInGPU.outerZ[innerMDIndex];
+        float candidateCenterXs[] = {mdsInGPU.anchorX[innerMDIndex] + circleRadius * alpaka::math::sin(acc, circlePhi), mdsInGPU.anchorX[innerMDIndex] - circleRadius * alpaka::math::sin(acc, circlePhi)};
+        float candidateCenterYs[] = {mdsInGPU.anchorY[innerMDIndex] - circleRadius * alpaka::math::cos(acc, circlePhi), mdsInGPU.anchorY[innerMDIndex] + circleRadius * alpaka::math::cos(acc, circlePhi)};
+
+        //check which of the circles can accommodate r3LH better (we won't get perfect agreement)
+        float bestChiSquared = 123456789.f;
```
VourMa, line 376: This should probably be `SDL_INF`?
...which should probably move to the `SDL/Constants.cu` file, instead of the `MiniDoublet.cu`?

SDL/Event.cu
GNiendorf: Yes, they merged in a pull request recently that adds hyperbolic functions to Alpaka after I raised it in an issue. Since that is only in the latest version of Alpaka though, it may be better for now to make our own functions like you suggested. We can address this later on though after all the kernels are moved over.

VourMa: Ok, so do I understand correctly that you want this merged without changing this line?

SDL/Segment.cuh
VourMa: Sorry, I must have looked in a wrong branch? All good.

GNiendorf:
Unlinking the issue since I haven't addressed points 2 or 3 in this PR, so I don't want it to close just yet.

SDL/Event.cu
GNiendorf: Yup that's correct. I will fix it later on.

SDL/Event.cu
GNiendorf: This was causing a bug in the code. If you notice in master, the isQuad variable is created in the sdl.cc code as a vector of type short. Then after it gets moved through this kernel it gets manually copied to an array on the GPU which has type bool. So if you just try to replace the current manual copy with an automatic copy the data types don't match and it isn't copied correctly. The solution is to either change everything to type short or type bool. Initially I tried changing everything to bool type but it gave me an error saying I couldn't use the pointer of a "temporary rvalue" when copying data to the GPU. If you change everything to data type short though it works fine.

Let me try changing everything to bool again and see if I can get it to work. I agree that bool type makes more sense.

SDL/Event.cu
YonsiG: Yes!

VourMa: Thanks!

@GNiendorf Could you please remove the comment?

SDL/Segment.cu
VourMa: Thanks for the reply, Yanxi!

After investigating a bit, I understood this is a leftover comment for a development that is already done:
https://github.com/SegmentLinking/TrackLooper/blob/master/SDL/Segment.cu#L58-L91
I would remove it.

GNiendorf:
Timing
Screenshot 2023-04-04 at 3 25 51 PM

Validation plots: [Here](https://www.classe.cornell.edu/~gsn27/www/www/PR260/mywork_888722-PU200_5aacc9-PU200/summary/)

SDL/Event.cu
GNiendorf: Just to log this, I was not able to get the bool datatype working properly. As a better solution, I changed the datatype from short to char everywhere. Since char and bool have the same memory footprint this is a slight improvement over master which has isQuad as a short initially.

VourMa: For documentation, here is a link with pointers to the reason why this doesn't work and an alternative way to deal with it:
https://stackoverflow.com/questions/39701660/fast-way-of-conversion-of-stl-vectorbool-to-bool-array

VourMa (approved):
This looks great, merging... Thanks to @GNiendorf for all the investigations!

PR #261, opened 2023-03-28
T3 kernel rewriting

YonsiG:
following PR #248 , make a general rewriting of T3 building functions, remove/combine the similar function of passPointingXXX and runTripletDefaultAlgoXXXX. Originally create T3 out of T4, so in a T4, need to judge the alignment of MD1 and MD3. Now create T3, don't need to judge the alignment of MD1 and MD2 (already have requirements in sg functions)


VourMa:
Closing this, as it will need complete rework for the alpaka version.
@YonsiG

PR #262, opened 2023-03-31
README update (adding description for lower level object plotting)

sgnoohc:
Added description for lower level object plotting


VourMa (approved):
Thanks a lot, approved and merged!

PR #272, opened 2023-04-12
Move CPU Range functions to GPU

tresreid:
Moves CPU Range functions to GPU for T3s and LSs. Fixes bug within T5.
MDs introduce a bug in writing to the ntuple so that has been omitted for now.

General comment, these range functions originally loops through the modules, check if the module has any lower level objects (if not then skip) and appends the maximum number of objects that can be created at that time. Afterwards, the max occupancy for that module based on the location is added to the maximum number of objects and used in the next non-empty module. This range array is used in 2 ways:

To generate a unique ID for any object
To obtain the maximum occupancy for any model
The implementation of 2 works by finding the difference between this module and the next module.
What this means is that when running in parallel, the indices are no longer in order and thus the differences do not necessarily correspond to the intended occupancy number (this was the T5 bug). This issue is solved by introducing an additional small array of size nLowerModules to hold the actual occupancy and that is used when necessary. There is no real difference in the T5s fixing this bug. For the first point, this randomness in the indexing of the arrays does not matter as the IDs will still be unique.


tresreid:
Validation: https://www.classe.cornell.edu/~mgr85/SegmentLinking_keep/PR272/PR272_PU200_NEVT-1_42daf3-PU200/summary/
Comparison: https://www.classe.cornell.edu/~mgr85/SegmentLinking_keep/PR272/PR272comparison_398270-PU200_42daf3-PU200/summary/

Timing: 
image

tresreid:
image
Kernel times are quite small. 
I attribute the small slow down in multistreaming to the fact that there is more work done on the GPU and less on the CPU so pipelining has less of an effect here. I don't believe we see any slow down in the single stream for this reason.

GNiendorf:
Just a general comment, but I wonder if removing the use of shared memory so that you could have less threads per block and more total blocks (since you're maximizing that one block out at 1024 threads) would improve the scaling. I know because you're using shared memory that you're forced to use only one block.

tresreid:
Though I don't think it is necessary to move to multiple blocks, I think I see what you are saying wrt using less than the maximum number of threads per block to improve the scaling. It is a good idea and I will look into it.

tresreid:
API calls from this PR: 
image


API calls from master:
image

But note that these numbers vary a lot between runs. I'm only running 1 event here.

tresreid:
Memory transfers from this PR: 
image

Memory transfers from master: 
image

tresreid:
I'm not entirely sure about the nature of the minidoublet bug. But I will loop back to it. 
In terms of Gavin's comment, my understanding is that the point is that the scaling becomes more poor because we are asking for an additional block that will take up a full SM. However, given that the actual kernel runtime is fairly small, we can reduce the number of threads which might open up more space on an SM to run more work and improve the scaling. However, trying to reduce the number of threads launched for these kernels do not seem to give further improvement. As for removing shared memory, I do not see why this would be helpful. We don't need to speed up this kernel by running in multiple blocks and furthermore we would have to coordinate between different blocks to understand the state of the counters there (since each number in the filled array uses the counters to get a unique ID, we can't have duplicate numbers in the array). I would think the way to achieve this with multiple blocks is to run through once to fill the array and then add an offset depending on the block. And this offset would be the total counters in the other blocks. But then you would have to loop through the array once more. And this would be worse in terms of the timing. So I don't think getting rid of shared memory and running on multiple blocks would really help. In Gavin's PR where he sees the speedup, he only reduces the number of threads launched in a kernel that already runs on multiple blocks. Additionally, it seems the old thread numbers don't work at all. So probably something else has changed as well (I expect register usage as this is usually the limiting factor to the threads that can be run). In any case, I plan to leave the threads as is for now. 
My Plan is to let this PR get merged as is (with the additional cleanup) and do the addXXXtoEvent functions today as a separate PR.

VourMa:
From the numbers with the `nvprof` results, I understand that this was run with the `-w 2` option, is that correct?
If we don't write out the NTuple, i.e. using the `-w 0` option for the numbers, we should get O(100) DtoH transfers. If I take the decrease in the number of DtoH calls in your post to be exclusively from intermediate memory transfers, which should be the case, we save 14 of those, i.e. we have a ~15% decrease in those calls. Not bad.

VourMa (approved):
The PR look good to me. I am approving now but I will leave some time before merging, in case @GNiendorf has a strong opinion on [this one](https://github.com/SegmentLinking/TrackLooper/pull/272#discussion_r1165266579).

GNiendorf:
> The PR look good to me. I am approving now but I will leave some time before merging, in case @GNiendorf has a strong opinion on [this one](https://github.com/SegmentLinking/TrackLooper/pull/272#discussion_r1165266579).

I'm fine with merging as is. I think it's appropriate to keep them as unsigned ints for now and I'll move them over to ints on the alpaka branch. As for the timing, it might be good to explore a comparison between shared memory on 1 block vs. using multiple blocks and no shared memory to see how well each one scales, but that can be done later on.

PR #132, opened 2022-04-20
First Pass Cleanup

tresreid:
Cleans up many of the warnings printing in compilation. 
Moves max objects to constants.h from kernels.cuh
adds 1 more sigfig to timing printout.

PR #278, opened 2023-04-18
Clean Up Duplicate Legacy CPU Code

GNiendorf:
In PR #132 the legacy CPU code was supposed to be moved to the cpu/ directory, but was instead duplicated and also moved to the code/core directory. In the #240 GNN Pull Request, the old legacy functions that were duplicated in the code/core directory started being used in a couple of the GNN methods in write_sdl_ntuple.cc

This PR removes most of the duplicate CPU files from the code/core directory, while keeping the original files in the cpu/ directory intact. The two legacy files that are kept, Hit.cc/h and MathUtil.c/h, are required by the new GNN code. After the Alpaka branch is merged in though these two files should be deleted as well and the GNN code can use the corresponding files in the SDL directory.

I also moved Algo.h from SDL/ to the cpu directory which should have been moved in PR #132

Timing (unchanged since none of these files were used)
Screenshot 2023-04-18 at 7 04 27 PM


VourMa (approved):
On my side, the PR looks ok, hence I am approving it.

However, I have not worked with the CPU code at all, so I may be missing some conceptual problem with the clean up. As a result, I leave it to @sgnoohc to take a quick look and merge, if there is no problem indeed.

PR #279, opened 2023-04-19
Move Constants to Alpaka + General Alpaka Cleanup

GNiendorf:
This PR moves the device side constants in Constants.cuh over to Alpaka, creates temporary hyperbolic functions to replace CUDA equivalents (since Alpaka has only recently added support for them), and does general cleanup (moves a device side function in Minidoublets.cu to Alpaka, resolves a cyclical include statement in Modules.cu by moving the Alpaka include to Constants.cuh, replaces a couple instances of CUDA's sin() function being called instead of Alpaka's, and general formatting fixes). I made sure to check that the temporary alpaka hyperbolic functions produce outputs close to CUDA's built-in functions. The two functions only differ by around .00004% or so, so there shouldn't be any changes to the performance.

This PR also removes duplicate math functions in Hit.cuh that were not able to be removed until all kernels were moved to Alpaka.

edit: This PR also removes the following unused functions from modules.cu
1. bool SDL::modules::parseIsInverted(unsigned int index)
2. bool SDL::modules::parseIsLower(unsigned int index)
3. SDL::ModuleType SDL::modules::parseModuleType(short subdet, short layer, short ring)
4. SDL::ModuleType SDL::modules::parseModuleType(unsigned int index)
5. SDL::ModuleLayerType SDL::modules::parseModuleLayerType(ModuleType moduleTypex,bool isInvertedx, bool isLowerx)
6. SDL::ModuleLayerType SDL::modules::parseModuleLayerType(unsigned int index)

This PR Timing (outdated, see comment below)
Screenshot 2023-04-19 at 4 17 08 PM


GNiendorf:
Got a nice speedup from the last commit, [c31cd8f](https://github.com/SegmentLinking/TrackLooper/pull/279/commits/c31cd8fd21d9d80f8c874dfdc612896ff4ea76f3)
Screenshot 2023-04-19 at 8 41 29 PM

SDL/MiniDoublet.cu
```diff
@@ -1,86 +1,18 @@
...
-    mdsInGPU.outerHighEdgeY[idx] = hitsInGPU.highEdgeYs[outerHitIndex];
-    mdsInGPU.outerLowEdgeX[idx] = hitsInGPU.lowEdgeXs[outerHitIndex];
-    mdsInGPU.outerLowEdgeY[idx] = hitsInGPU.lowEdgeYs[outerHitIndex];
-}
-
 void SDL::createMDArrayRanges(struct modules& modulesInGPU, struct objectRanges& rangesInGPU, uint16_t& nLowerModules, unsigned int& nTotalMDs, cudaStream_t stream, const unsigned int& maxPixelMDs)
 {
     /*
-        write code here that will deal with importing module parameters to CPU, and get the relevant occupancies for a given module!*/
+        write code here that will deal with importing module parameters to CPU, and get the relevant occupancies for a given module!
```
VourMa, line 14: Isn't this done starting at line 35 of this file? If so, I would delete the comment (I think we have deleted similar ones previously).

SDL/MiniDoublet.cuh
```diff
@@ -248,7 +292,185 @@ namespace SDL
         }
     }
 
-    ALPAKA_FN_ACC void shiftStripHits(struct modules& modulesInGPU, uint16_t& lowerModuleIndex, uint16_t& upperModuleIndex, unsigned int lowerHitIndex, unsigned int upperHitIndex, float* shiftedCoords,float xLower,float yLower,float zLower,float rtLower,float xUpper,float yUpper,float zUpper,float rtUpper);
+    template<typename TAcc>
+    ALPAKA_FN_INLINE ALPAKA_FN_ACC void shiftStripHits(TAcc const & acc, struct modules& modulesInGPU, /*struct hits& hitsInGPU,*/ uint16_t& lowerModuleIndex, uint16_t& upperModuleIndex, unsigned int lowerHitIndex, unsigned int upperHitIndex, float* shiftedCoords, float xLower, float yLower, float zLower, float rtLower,float xUpper,float yUpper,float zUpper,float rtUpper)
```
VourMa, line 296: Shall we delete the unused/commented-out argument?

SDL/MiniDoublet.cuh
```diff
@@ -248,7 +292,185 @@ namespace SDL
...
+        drprime_x = drprime * alpaka::math::sin(acc, angleM);
+        drprime_y = drprime * alpaka::math::cos(acc, angleM);
+
+        // The new anchor position is
+        xa = xp + drprime_x;
+        ya = yp + drprime_y;
+
+        // The original strip hit position
+        //xo = hitsInGPU.xs[stripHitIndex];
+        //yo = hitsInGPU.ys[stripHitIndex];
```
VourMa, line 436: I guess these go unused, so we should probably delete them to avoid confusion?

SDL/Module.cuh
```diff
@@ -95,11 +92,11 @@ namespace SDL
...
-        ALPAKA_FN_HOST_ACC ModuleType parseModuleType(unsigned int index);
-        ALPAKA_FN_HOST_ACC ModuleType parseModuleType(short subdet, short layer, short ring);
-        ALPAKA_FN_HOST_ACC unsigned int parsePartnerModuleId(unsigned int detId, bool isLowerx, bool isInvertedx);
-        ALPAKA_FN_HOST_ACC ModuleLayerType parseModuleLayerType(unsigned int index);
-        ALPAKA_FN_HOST_ACC ModuleLayerType parseModuleLayerType(ModuleType moduleType, bool isInvertedx, bool isLowerx);
+        ModuleType parseModuleType(unsigned int index);
+        ModuleType parseModuleType(short subdet, short layer, short ring);
+        unsigned int parsePartnerModuleId(unsigned int detId, bool isLowerx, bool isInvertedx);
+        ModuleLayerType parseModuleLayerType(unsigned int index);
+        ModuleLayerType parseModuleLayerType(ModuleType moduleType, bool isInvertedx, bool isLowerx);
```
VourMa, line 99: Are we going a bit back and forth here? Why was the alpaka keywords, supposedly added in one of your previous PRs, removed here?

SDL/Quintuplet.cuh
```diff
@@ -998,7 +998,7 @@ namespace SDL
         float sdlPVoff = 0.1f/rt_OutLo;
         sdlCut = alpha1GeV_OutLo + alpaka::math::sqrt(acc, sdlMuls * sdlMuls + sdlPVoff * sdlPVoff);
 
-        deltaPhiPos = SDL::deltaPhi_alpaka(acc, mdsInGPU.anchorX[secondMDIndex], mdsInGPU.anchorY[secondMDIndex], mdsInGPU.anchorX[fourthMDIndex], mdsInGPU.anchorY[fourthMDIndex]);
+        deltaPhiPos = SDL::deltaPhi(acc, mdsInGPU.anchorX[secondMDIndex], mdsInGPU.anchorY[secondMDIndex], mdsInGPU.anchorX[fourthMDIndex], mdsInGPU.anchorY[fourthMDIndex]);
```
VourMa, line 1001: It seems to me that there is some inconsistency on how we use the `SDL` namespace for these functions. I guess it should always be explicitly there for the `copysignf` function, since there is a one with the same name in the standard `math.h` library but I understand that it doesn't have to be there for the `deltaPhi` function, since it isn't there in other files (contrary to what is done here). Should we try to homogenize?

For my own understanding, is it true that the `SDL` namespace is mandatory for the `copysignf`? Is this the case for other functions?

VourMa:
> @VourMa when you get a chance can you check that your sdl.cc in the SDL/ directory runs fine with this fix?

I guess that the test you wanted me to perform was to check whether the changes in the `LST.cc/h` break the CMSSW workflow? If so, that's what I did below. If not, please clarify.

Checks:
- I pulled this PR, compiled the standalone, then compiled CMSSW and finally ran the test configuration. All of the steps succeeded but the output was empty, i.e. the there were no CMSSW tracks produced.
- To verify that the current PR is the problem, I pulled the `alpaka_move` branch and followed the same steps. The result was the same, i.e. no CMSSW tracks produced. As a result, the failure is a feature of the alpaka migration in general and not of this specific PR.
- As a sanity, I pulled the latest `master ` branch and repeated the procedure. CMSSW tracks were produced - although not as many as I anticipated, so I need to investigate... This is not a showstopper for this PR though.

The way I see it, I think that we should continue with the alpaka migration as planned but, before merging to the master branch, we should revisit the integration to CMSSW, recheck if the output is produced and, if not, track down which part of the workflow is broken. What do people think?

SDL/Module.cuh
GNiendorf: This was carried over from master, where these functions have the hostdev prefix, but I realized these functions are only ever run on the host, so their prefixes are kind of pointless. Also, I realized a lot of these functions aren't even being used so I just removed most of them.

SDL/Event.cuh
```diff
@@ -10,6 +10,7 @@
 #include "MiniDoublet.cuh"
 #include "PixelTriplet.cuh"
 #include "TrackCandidate.cuh"
+#include "Constants.cuh"
```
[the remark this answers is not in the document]
GNiendorf: I guess this is just preference, but Event.cu uses alpaka keywords obviously, so I figured it would be good to explicitly include the Constants.cuh file where the alpaka include is called. I think this could be removed without any issues, but then we would be relying on some of these other files that are included also including Constants.cuh rather than just explicitly including it where it's needed.

SDL/Quintuplet.cuh
GNiendorf: Yeah that's a good point, I think best practice would be to add the SDL prefix to the relevant functions so that it's clear  where these functions are being called from. I agree that copysignf is the only function where it's really required, but good either way to homogenize everything like you said.

GNiendorf:
@VourMa More speed improvements from the last commit 👀. I guess clean code is fast code.

New Timing
Screenshot 2023-04-22 at 5 42 28 PM

Validation Plots: Soon
Comparison Plots: Soon

GNiendorf:
> The way I see it, I think that we should continue with the alpaka migration as planned but, before merging to the master branch, we should revisit the integration to CMSSW, recheck if the output is produced and, if not, track down which part of the workflow is broken. What do people think?

Thank you for testing this. I agree with you that this is something we should fix at the end before we merge the alpaka branch into master. Maybe it will work itself out as well.

SDL/Event.cuh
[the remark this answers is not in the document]
VourMa: Ok, my concern was that the re-inclusion of the same file through different `include`s could cause a problem in the long run, as we have seen it was the case for some other files we have been including everywhere, even though it was not needed. But, if you think it is only a matter of preference in this case and it won' cause problems, I am fine with including it.

SDL/Module.cu
VourMa, line 1: I can see that a few `SDL::modules` functions were removed in this commit. I guess these were functions that we were unused and I agree with removing them. However, they don't seem like trivial function (one would have to spend an afternoon recreating their logic from scratch) and I am not sure they will be totally useless in the future.
As a result, I think that it would be important to somehow keep track of them. Probably the best way to do so at this stage would be to mention their removal and even their names explicitly in the PR description. Hopefully, this could help in finding a reference to them in the future, if we need them again.

VourMa (approved):
Apart from a comment from the previous time which you can deal with in any way you like (even by just resolving its discussion) and a minor PR description modification requested now, the PR is good to go. I am approving it for now but I will delay merging it until:

1. You tell me that you dealt with the comment in your preferred way
2. Other people take a look at the comment I made about the CMSSW integration and comment or I discuss it in the internal UCSD meeting today. Either way, this will be done by tonight.

SDL/Event.cuh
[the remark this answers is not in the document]
GNiendorf: Yea but this case is different though because it's just a constants.cuh file that itself doesn't include any other files in the SDL directory, and every file that includes it actually uses it. The other files that were included everywhere (kerles.cuh for example) had a bunch of cyclical includes and it was never even used by the files that included it lol.

PR #175, opened 2022-06-07
Removal of tracking algo iterations

VourMa:
This PR removes all tracking algorithm iterations except initialStep and highPtTripletStep, to be compatible with what's being used for HLT Tracking @ Phase 2. The removal of the extra iterations (lowPtTripletStep, lowPtQuadStep, detachedTripletStep and detachedQuadStep) leads to small reduction of the efficiency (percent level reduction in some bins low/medium pT bins) but also cuts the fake rate at higher |η| almost by half. Validation plots can  be found here:
https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/trkIterAlgoComp/eff_comparison_plots__1d99f25noOtherSteps_explicit_cache_1d99f25allSteps_explicit_cache_PU200/

PR #280, opened 2023-04-25
Sync latest standalone developments with CMSSW class

VourMa:
This PR brings the CMSSW class up to speed with the latest standalone developments by properly initializing the `SDL::Event` object. It also simplifies the CMSSW input logic, dropping the selection based on the pixel seed iteration, as this is controlled by the module input tags in CMSSW.

The PR has been tested by running the CMSSW configuration on a few events and verifying that the TrackCandidates are the same.

Coupled with SegmentLinking/cmssw#9.


GNiendorf:
Is the code in code/core/trktree.cc relating to "see_algo" still necessary? I'm not sure what trktree.cc is used for to be honest, but I saw that it has references to the "see_algo" variable.

VourMa:
> Is the code in code/core/trktree.cc relating to "see_algo" still necessary?

Yes. When loading the pixel seeds from the tracking NTuple, pixel seeds from all iterations are included. Based on what Phase-2 HLT setup was using and some studies done in the past (PR #175 and [plots](https://www.classe.cornell.edu/~evourlio/www/SDL_GPU/trkIterAlgoComp/)), we settled (for now) to including only initialStep and highPtTripletStep pixel seeds. This selection is performed with the `see_algo` variable in `code/core/trktree.cc`. The latter file also prepares the LST input in the proper format.

For CMSSW, we control explicitly pixel seeds from which iterations to include ([here](https://github.com/SegmentLinking/cmssw/blob/52f4284ef0d81c39ca9a4339aa478a8498149004/RecoTracker/LST/plugins/LSTPixelSeedInputProducer.cc#L57-L58)), so no need select the pixel seeds also downstream, hence the simplification (which also simplifies the CMSSW side as well).

GNiendorf (approved):
Looks good to me. I'll approve the PR for now and let @tresreid look it over and merge.

tresreid:
Looks good to me.

PR #283, opened 2023-05-02
placed printf in warnings ifdef

tresreid:
By @VourMa's suggestion, placed all printf statements in quintuplets.cu around the warnings ifdef. This does seem to reduce the max registers in T5 creation from 255 to 167 (still the highest but better than before.). This doesn't seem to happen anywhere else.
image

No physics validation


VourMa (approved):
Thanks for the prompt follow-up! I think this is ok to be merged.

PR #284, opened 2023-05-03
Move New "Ranges" Kernels to Alpaka + Working Alpaka Multistreaming

GNiendorf:
This PR moves the recent kernels created by @tresreid over to Alpaka. These are the last kernels in the SDL directory that need to be moved over on the current master branch.

This PR Timing
Screenshot 2023-05-02 at 11 21 27 PM

This PR also introduces multistreaming with Alpaka. PR #286 resolved the issues I was having a while ago.

Timeline view with multistreaming (s=8). Notice that there are 8 streams which handle the kernel execution now through Alpaka, and the original 8 CUDA streams which only deal with memory allocation and transfers. Also an extra stream at the very bottom, but that's also present on the master branch. I think that's just the default stream or a stream used to set up the modules. In later PR's, after the alpaka caching allocator is used, I will remove the remaining CUDA streams so that it's equivalent to the master branch.
Screenshot 2023-05-11 at 0 36 17 PM

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -768,11 +768,30 @@ void SDL::Event::addPixelSegmentToEvent(std::vector<unsigned int> hitIndices0,st
         mdsInGPU = (SDL::miniDoublets*)cms::cuda::allocate_host(sizeof(SDL::miniDoublets), stream);
         unsigned int nTotalMDs;
         cudaMemsetAsync(&rangesInGPU->miniDoubletModuleOccupancy[nLowerModules],N_MAX_PIXEL_MD_PER_MODULES, sizeof(unsigned int),stream);
-        createMDArrayRangesGPU<<<1,1024,0,stream>>>(*modulesInGPU, *rangesInGPU);
+
+        // Temporary fix for queue initialization.
+        QueueAcc queue(devAcc);
+
+        Vec const threadsPerBlockCreateMD(static_cast<Idx>(1), static_cast<Idx>(1), static_cast<Idx>(1024));
+        Vec const blocksPerGridCreateMD(static_cast<Idx>(1), static_cast<Idx>(1), static_cast<Idx>(1));
+        WorkDiv const createMDArrayRangesGPU_workDiv(blocksPerGridCreateMD, threadsPerBlockCreateMD, elementsPerThread);
+
+        SDL::createMDArrayRangesGPU createMDArrayRangesGPU_kernel;
+        auto const createMDArrayRangesGPUTask(alpaka::createTaskKernel<Acc>(
+            createMDArrayRangesGPU_workDiv,
+            createMDArrayRangesGPU_kernel,
+            *modulesInGPU,
+            *rangesInGPU));
+
+        alpaka::enqueue(queue, createMDArrayRangesGPUTask);
+        alpaka::wait(queue);
+
         cudaMemcpyAsync(&nTotalMDs,rangesInGPU->device_nTotalMDs,sizeof(unsigned int),cudaMemcpyDeviceToHost,stream);
         cudaStreamSynchronize(stream);
         nTotalMDs+= N_MAX_PIXEL_MD_PER_MODULES;
+
         createMDsInExplicitMemory(*mdsInGPU, nTotalMDs, nLowerModules, N_MAX_PIXEL_MD_PER_MODULES,stream);
+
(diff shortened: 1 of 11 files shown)


GNiendorf:
Everything looks good so I'm marking this as ready for review.

This PR timing (basically unchanged from CUDA kernels)
Screenshot 2023-05-03 at 2 55 10 PM

Validation Plots: [Here](https://www.classe.cornell.edu/~gsn27/www/www/PR284/mywork_67f716-PU200/summary/)
Comparison to Master: [Here](https://www.classe.cornell.edu/~gsn27/www/www/PR284/mywork_ce2f36-PU200_67f716-PU200/summary/)

SDL/MiniDoublet.cuh
```diff
@@ -729,17 +719,114 @@ namespace SDL
...
+            using Dim = alpaka::Dim<TAcc>;
+            using Idx = alpaka::Idx<TAcc>;
+            using Vec = alpaka::Vec<Dim, Idx>;
 
+            Vec const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+            Vec const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
 
-}
+            // Initialize variables in shared memory and set to 0
+            int& nTotalMDs = alpaka::declareSharedVar<int, __COUNTER__>(acc); nTotalMDs = 0;
```
VourMa, line 744: I would do two lines here, just so that the initialization is clear.

If you agree with the proposal, this applies to more places below.

SDL/MiniDoublet.cuh
```diff
@@ -729,17 +719,114 @@ namespace SDL
...
+            alpaka::syncBlockThreads(acc);
 
-#endif
+            // Initialize variables outside of the for loop.
+            float module_eta;
+            unsigned int nTotMDs;
+            int occupancy, category_number, eta_number;
+            short module_subdets, module_layers, module_rings;
```
VourMa, line 751: Since there is some moving around of the declaration of the variables, are we sure that we want them defined outside of the loop, even if they are not used anywhere else? I would think that the better practice would be to declare them inside of the scope they are used.
I also search around the internet and a lot of the selected replies in stackoverflow, etc. questions, e.g. [here](https://stackoverflow.com/questions/7959573/declaring-variables-inside-loops-good-practice-or-bad-practice), seem to agree that this could be better for compiler optimizations. I guess this is not very relevant here but it would be good to be consistent with our preferred practice.

Depending on which practice we decide to follow, this could imply changes to more places below.

SDL/Event.cu
```diff
@@ -1000,14 +1050,28 @@ void SDL::Event::createMiniDoublets()
         *mdsInGPU,
         *rangesInGPU));
 
+    // No wait is needed here, runs concurrently with next kernel.
```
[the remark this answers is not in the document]
GNiendorf: Oh yeah, that's strange. @tresreid, Why didn't you include a stream sync between the two for most of these kernels? I agree that it doesn't seem like they should run concurrently.

SDL/MiniDoublet.cuh
GNiendorf: Thank you for looking into this question more. I think the comments about scope aren't relevant here because the for loop is wrapped in its own kernel already, so there's no chance of it causing issues downstream. However, since none of the others kernels do this, and there's probably no noticeable timing difference from it, I'll define them in the loop for now and we can keep this in the back of our minds.

GNiendorf:
This PR Timing:
Screenshot 2023-05-11 at 0 20 34 PM

SDL/Event.cuh
```diff
@@ -14,13 +14,27 @@
...
+
 namespace SDL
 {
     class Event
     {
     private:
         cudaStream_t stream;
         bool addObjects;
+
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
```
VourMa, line 32: Out of curiosity, is this needed in our case? Aren't we always compiling for the cuda backend? Or is it a guard for the future (that we didn't have before, as we were defining the accelerator without it previously)?

GNiendorf: It's actually needed because of the LST.cc that's present in the SDL directory now. The LST.cc is compiled with the host compiler of nvcc, and it throws an error without the ifdef here.

VourMa (approved):

PR #288, opened 2023-05-15
Harmonize the LST class indentation with the rest of the code

VourMa:
As per title. Purely "technical" PR.


GNiendorf (approved):
Thank you for fixing this Manos. Approving.

PR #291, opened 2023-05-23
Add T5 DNN

jkguiang:
## Summary
Added hard-coded T5 DNN to T5-building algorithm in place of the chi-squared cuts (see these [slides](http://uaf-10.t2.ucsd.edu/~jguiang/presentations/lst/lst_05-19-2023.pdf) for more info). The DNN gives a significant reduction in the T5 fake rate (particularly in the barrel) with only a slight change to the efficiency. This can be seen by comparing the efficiency plots in [slide 8](http://uaf-10.t2.ucsd.edu/~jguiang/presentations/lst/lst_05-19-2023.pdf#page=8) (original master) and [slide 9](http://uaf-10.t2.ucsd.edu/~jguiang/presentations/lst/lst_05-19-2023.pdf#page=9) (this PR) in the aforementioned slides. Finally, the current implementation of the DNN does significantly affect the overall runtime of the T5 building step (see the "Timing" section below).

#### This PR
```
Total Timing Summary
   avg      8.5      2.0      1.4      3.7      6.7      1.4      1.5      1.1      2.1      28.5      18.6+/-  3.9      32.1   explicit[s=1]
   avg     12.5      3.3      1.8      6.2     10.0      1.5      4.0      2.5      3.4      45.3      31.3+/-  6.6      27.2   explicit[s=2]
   avg     25.1      4.6      2.7     10.2     16.2      1.7      9.5      5.1      5.0      80.0      53.3+/- 10.3      28.8   explicit[s=4]
   avg     34.9      4.5      2.7     13.7     22.6      1.9     17.3      8.1      5.8     111.6      74.9+/- 18.3      27.7   explicit[s=6]
   avg     56.1      4.6      2.5     17.5     30.8      1.7     25.5     13.6      5.8     158.1     100.3+/- 27.9      29.1   explicit[s=8]
```
#### Original master
```
Total Timing Summary
   avg      8.9      2.0      1.4      3.7      4.1      1.4      1.5      1.1      2.2      26.2      15.9+/-  3.3      29.4   explicit[s=1]
   avg     13.6      3.5      1.9      6.3      6.3      1.5      3.6      2.5      3.7      43.0      27.9+/-  5.7      27.1   explicit[s=2]
   avg     20.2      3.7      2.3     10.0     12.2      1.5      8.8      5.1      4.6      68.4      46.6+/- 10.2      25.8   explicit[s=4]
   avg     31.1      5.1      2.9     12.9     15.6      1.8     14.1      7.8      5.9      97.2      64.3+/- 16.4      24.6   explicit[s=6]
   avg     49.5      4.1      2.5     16.5     26.5      1.7     23.3     12.9      5.6     142.5      91.3+/- 27.1      26.3   explicit[s=8]
```


VourMa:
Quick comment by just scrolling through the changes: could we add the hardcoded arrays in `SDL/Constants.cu`? I feel like it is much more appropriate to include them there rather than injecting them in the kernel.

SDL/Quintuplet.cu
```diff
@@ -428,6 +428,236 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
...
+    float bias_0[32] = {
+        -0.453155100345611572265625000000,-3.414395570755004882812500000000, 0.461477220058441162109375000000,-0.154097318649291992187500000000,-0.470577239990234375000000000000, 2.228851079940795898437500000000,-0.163815096020698547363281250000,-2.296737670898437500000000000000,-0.098085217177867889404296875000, 2.589857339859008789062500000000, 2.105398178100585937500000000000,-0.525056004524230957031250000000, 1.790417790412902832031250000000,-2.162935256958007812500000000000, 0.012643844820559024810791015625,-1.595005750656127929687500000000, 0.222240477800369262695312500000,-3.097779273986816406250000000000,-0.748212277889251708984375000000, 1.027869462966918945312500000000,-0.504925668239593505859375000000, 0.079979300498962402343750000000,-0.680445253849029541015625000000, 1.331984400749206542968750000000, 2.027832508087158203125000000000,-0.063643321394920349121093750000,-0.304876327514648437500000000000,-0.020376743748784065246582031250, 2.847964286804199218750000000000,-1.884062528610229492187500000000, 2.168398141860961914062500000000,-2.673130989074707031250000000000
```
slava77, line 485: for longer term, we should see how to load these from a file.
For this PR, I'd suggest to format the values in float precision to have a more compact code.

For the hackathon, I'm curious how much of this can be sped up by using half-precision and/or tensor operations

GNiendorf:
> Quick comment by just scrolling through the changes: could we add the hardcoded arrays in `SDL/Constants.cu`? I feel like it is much more appropriate to include them there rather than injecting them in the kernel.

Could we keep them in their own separate (new) file maybe? Would make my life a lot easier when rebasing, and would probably be cleaner <3

jkguiang:
Just saw Gavin's comment, I will add another .cu file to contain the hard-coded weights and biases.

SDL/NeuralNetworkWeights.cu
VourMa, line 1: Sorry to bug you again about this but the direction we are moving towards with Gavin while migrating to alpaka is to have only the `.cuh` file for files with just arrays/constants (e.g. check what happens with `Constants.cu` and `Constants.cuh` [here](https://github.com/SegmentLinking/TrackLooper/pull/209/files)). Could you unify this file with its `.cuh` partner?

SDL/Quintuplet.cu
```diff
@@ -428,6 +428,117 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
...
+    {
+        layer2_adjustment = 1; // get upper segment to be in second layer
+    }
+    bool is_endcap1 = (modulesInGPU.subdets[lowerModuleIndex1] == 4);
+    bool is_endcap2 = (modulesInGPU.subdets[lowerModuleIndex2] == 4);
+    bool is_endcap3 = (modulesInGPU.subdets[lowerModuleIndex3] == 4);
+    bool is_endcap4 = (modulesInGPU.subdets[lowerModuleIndex4] == 4);
+    bool is_endcap5 = (modulesInGPU.subdets[lowerModuleIndex5] == 4);
+    float x[38] = { 
+        log10(2.99792458e-3f*3.8f*innerRadius),                       // inner t3_pt
```
VourMa, line 442: We have a dedicated variable (`k2Rinv1GeVf`) to be used in such cases, defined in `Constants.cu`. Could you switch to that?

The same comment possibly applies to more cases below and for other constants, e.g. the magnetic field.

SDL/Quintuplet.cu
```diff
@@ -428,6 +428,117 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
...
+        mdsInGPU.anchorEta[firstMDIndex],                             // inner (hit 1) t3_0_eta
+        mdsInGPU.anchorPhi[firstMDIndex],                             // inner (hit 1) t3_0_phi
+        mdsInGPU.anchorZ[firstMDIndex],                               // inner (hit 1) t3_0_z
+        sqrtf(x1*x1 + y1*y1),                                         // inner (hit 1) t3_0_r
```
VourMa, line 446: Maybe this is not for this PR but would it make sense to use r^2, instead of r, to avoid extra, costly operations?

The same comment applies to more instances below.

SDL/Quintuplet.cu
```diff
@@ -428,6 +428,117 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
...
+        mdsInGPU.anchorPhi[fourthMDIndex],                            // outer (hit 4) t3_2_phi
+        mdsInGPU.anchorZ[fourthMDIndex],                              // outer (hit 4) t3_2_z
+        sqrtf(x4*x4 + y4*y4),                                         // outer (hit 4) t3_2_r
+        float(modulesInGPU.layers[lowerModuleIndex4] + 6*is_endcap4), // outer (hit 4) t3_2_layer
+        mdsInGPU.anchorEta[fifthMDIndex],                             // outer (hit 5) t3_4_eta
+        mdsInGPU.anchorPhi[fifthMDIndex],                             // outer (hit 5) t3_4_phi
+        mdsInGPU.anchorZ[fifthMDIndex],                               // outer (hit 5) t3_4_z
+        sqrtf(x5*x5 + y5*y5),                                         // outer (hit 5) t3_4_r
+        float(modulesInGPU.layers[lowerModuleIndex5] + 6*is_endcap5), // outer (hit 5) t3_4_layer
+        log10((innerRadius+outerRadius)*3.8f*1.602f/(2*100*5.39f)),   // t5_pt
```
VourMa, line 474: What are those numbers? If they are constants, could they be added in the proper file and included here? If they are some temporary, magic numbers, could they be added in this file with a descriptive name (and possibly a descriptive comment), so that one can follow the calculations?

SDL/Quintuplet.cu
```diff
@@ -428,6 +428,117 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
...
+            x_4[col] += x_3[inner]*T5DNN::wgtT_4[inner][col];
+        }
+        x_4[col] += T5DNN::bias_4[col];
+    }
+    
+    // (5): Sigmoid()
+    float x_5[1];
+    for (unsigned int col = 0; col < 1; ++col)
+    {
+        x_5[col] = exp(x_4[col])/(exp(x_4[col]) + 1);
```
VourMa, line 536: I was wondering whether 1/(1+exp(-x)) or exp(x)/(1+exp(x)) would be faster for the sigmoid. Maybe it's worth checking.
While searching about it, I also found this post:
https://stackoverflow.com/questions/10732027/fast-sigmoid-algorithm
Probably not something for this PR but could be tried?

code/core/write_sdl_ntuple.cc
```diff
@@ -703,7 +703,7 @@ void setGnnNtupleBranches(SDL::Event* event)
         ana.tx->pushbackToBranch<vector<int>>("tc_lsIdx", lsIdx);
     }
 
-    std::cout <<  " ana.tx->getBranch<vector<vector<int>>>('tc_lsIdx').size(): " << ana.tx->getBranch<vector<vector<int>>>("tc_lsIdx").size() <<  std::endl;
+    // std::cout <<  " ana.tx->getBranch<vector<vector<int>>>('tc_lsIdx').size(): " << ana.tx->getBranch<vector<vector<int>>>("tc_lsIdx").size() <<  std::endl;
```
VourMa, line 706: I would propose to delete this. No reason to leave comments in the code (especially if they crash the code when enabled).

SDL/NeuralNetworkWeights.cu
jkguiang: Just to be clear (to avoid even more commits on this same issue), I should do this for the neural network weights while Constants.cu/h is still separated? I guess to ease the rebase?

SDL/Quintuplet.cu
jkguiang: If we used r^2 instead, I think we'd anyway need to normalize the value like we do for e.g. the pTs. In any case, I expect the `sqrt` call is a small cost compared to the for loops, no?

SDL/Quintuplet.cu
jkguiang: Interesting, though I'm not sure if we are sensitive to any speedup this might give, i.e. the for loops take so long that a small speedup in changing the sigmoid around is not noticeable. I can try at some point anyway, though, and the suggestion is interesting--thanks!

SDL/NeuralNetworkWeights.cu
VourMa: Yes, `Constants.cu/h` stay as they are to facilitate rebasing while only one file, `SDL/NeuralNetworkWeights.cuh`, is created for including the NN-related hardcoded info.

SDL/Quintuplet.cu
VourMa: Why normalize the value if it is `r^2` and not when it is `r`?

Maybe I have missed the point but my understanding was that you are using `r` as an input variable in the training (and inference), is that right? If that's the case, what I was proposing was to use `r^2` instead of `r` in the training, thinking that it would have similar discriminatory power but saving us the `sqrt` downstream.

SDL/Quintuplet.cu
VourMa: Yes, `3.8f` should be the magnetic field, so we could at least change that. But, in any case, I think it would be good to understand what all the numerical factors are. Not sure who wrote that formula but @sgnoohc was taking a look at radius calculations at some point, so may know what's going on?

jkguiang: Ok, I will make the replacement. The line comes from elsewhere in Quintuplet.cu:
```
float pt = (innerRadius+outerRadius)*3.8f*1.602f/(2*100*5.39f);
```

SDL/Quintuplet.cu
jkguiang: Sorry, to be clear, we need to normalize DNN input features such that they are all O(1). For example, we take the log10 of the T3 pTs and T5 pT. Thus, if we were to give the DNN r^2 instead of r, it would be O(100), which we would need to normalize like we do for the pTs. Does that make sense?

VourMa: I see. But then r = sqrt( O(100) ) = O(10), which is still one order of magnitude larger than O(1). For my education, is this fine? Where do we set the limit for normalization? Is it an empirical limit by checking if the training does well or something else?

jkguiang: Yes, at least for what we're doing right now. I've seen for other problems that normalizing anything that's more flat (e.g. phi) such that the values are made to be between 0 and 1 also helps. However, since the DNN is converging, we can say "taking the log of the pTs is enough for now."

SDL/Quintuplet.cu
```diff
@@ -433,51 +433,52 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
...
+        mdsInGPU.anchorZ[fourthMDIndex],                                             // outer T3 anchor hit 4 z (t3_2_z)
+        sqrtf(x4*x4 + y4*y4),                                                        // outer T3 anchor hit 4 r (t3_2_r)
+        float(modulesInGPU.layers[lowerModuleIndex4] + 6*is_endcap4),                // outer T3 anchor hit 4 layer (t3_2_layer)
+        mdsInGPU.anchorEta[fifthMDIndex],                                            // outer T3 anchor hit 5 eta (t3_4_eta)
+        mdsInGPU.anchorPhi[fifthMDIndex],                                            // outer T3 anchor hit 5 phi (t3_4_phi)
+        mdsInGPU.anchorZ[fifthMDIndex],                                              // outer T3 anchor hit 5 z (t3_4_z)
+        sqrtf(x5*x5 + y5*y5),                                                        // outer T3 anchor hit 5 r (t3_4_r)
+        float(modulesInGPU.layers[lowerModuleIndex5] + 6*is_endcap5),                // outer T3 anchor hit 5 layer (t3_4_layer)
+        log10((innerRadius + outerRadius)*SDL::magnetic_field*1.602f/(2*100*5.39f)), // T5 pT (t5_pt)
```
jkguiang, line 476: I forgot to mention it in the commit message, but I have replaced `3.8f` with `SDL::magnetic_field`

VourMa (approved):
Thanks for the updates! PR looks good to me, so I am approving.

I saw that the validation is included in the slides but am I ask to add a summary comment about the performance in PR description, so that one doesn't have to have through the full set of slides to understand the changes?

Same goes for the timing, which could also be improved in the slides by removing the "explicit" only results (which we don't usually look at), mentioning which timing corresponds to this PR and which to the master and explaining which numbers are used the ΔΤ5 column. Thanks!

GNiendorf:
Is the SDL::passChiSquaredConstraint function no longer used now that this PR is merged in? If so, should we remove it from the code?

slava77:
> Is the SDL::passChiSquaredConstraint function no longer used now that this PR is merged in? If so, should we remove it from the code? Along those lines, is there any other code that can be removed from this (maybe some variables that no longer need to be calculated that were used for the previous code)?

there was also a point to get rid of all regression computation, since it's not really used (or not supposed to).

VourMa:
Timing on lnx7188 (V100) - DNN timing is not really worse here...

**Before DNN:**
```
Total Timing Summary
   avg      4.4      3.2      4.1      3.2      3.3      1.2      1.9      1.4      3.6      26.1      20.5+/-  3.3      29.5   explicit_cache[s=1]
   avg      7.5      5.3      4.7      6.0      5.1      1.5      3.7      2.4      7.1      43.3      34.3+/-  6.4      25.8   explicit_cache[s=2]
   avg     16.3      8.6      5.9     12.8     10.0      2.1      8.3      4.9     13.5      82.4      63.9+/-  9.9      22.0   explicit_cache[s=4]
   avg     24.7     11.5      7.9     17.8     14.5      2.7     12.4      7.2     18.2     116.9      89.6+/- 17.7      22.0   explicit_cache[s=6]
   avg     37.0     15.2     10.2     22.0     20.3      3.5     18.1     10.3     24.8     161.3     120.8+/- 21.8      22.2   explicit_cache[s=8]
```

**After DNN:**
```
Total Timing Summary
   avg      4.4      3.1      3.4      3.0      3.5      1.2      1.8      1.4      3.5      25.4      19.9+/-  3.5      29.2   explicit_cache[s=1]
   avg      7.0      4.8      4.0      5.6      4.9      1.4      3.4      2.1      6.1      39.3      30.9+/-  6.8      26.1   explicit_cache[s=2]
   avg     15.8      9.1      7.8     14.0     11.8      2.1      8.5      4.9     13.0      87.0      69.0+/- 11.0      23.5   explicit_cache[s=4]
   avg     24.7     11.8      8.0     17.5     14.1      2.7     13.5      7.6     17.5     117.5      90.1+/- 19.4      22.4   explicit_cache[s=6]
   avg     37.0     14.9      9.7     23.2     20.5      3.7     18.0     10.6     23.9     161.6     120.8+/- 22.0      21.9   explicit_cache[s=8]
```

VourMa:
DNN performance within CMSSW:
LST w/o DNN -> Blue
LST w/ DNN -> Red

PU200 sample: https://uaf-10.t2.ucsd.edu/~evourlio/SDL/DNNv1InCMSSW_LSTvsLSTDNNOnly/
Cube5 sample:  https://uaf-10.t2.ucsd.edu/~evourlio/SDL/DNNv1InCMSSW_cube5_LSTvsLSTDNNOnly/

slava77:
> Timing on lnx7188 (V100) - DNN timing is not really worse here...

could it be that the code updates during the review are also responsible?

jkguiang:
I re-ran `sdl_timing PU200` today on `cgpu-1` with a fresh clone of the current `master` branch:
```
Total Timing Summary
   avg      1.7      2.0      1.5      3.4      3.7      1.4      1.4      1.0      2.1      18.1      15.1+/-  3.5      19.7   explicit_cache[s=1]
   avg      2.6      3.5      1.8      4.9      5.0      1.5      2.8      1.6      3.5      27.2      23.1+/-  4.9      14.3   explicit_cache[s=2]
   avg      4.9      5.2      3.1      9.3      8.9      1.7      6.4      3.2      6.1      48.8      42.1+/-  7.6      12.7   explicit_cache[s=4]
   avg      9.7      6.7      4.3     12.8     12.8      2.2      9.6      4.8      9.2      72.1      60.1+/- 12.0      12.7   explicit_cache[s=6]
   avg     16.9      9.1      6.3     18.5     18.2      2.6     13.3      6.7     12.1     103.6      84.1+/- 16.6      13.7   explicit_cache[s=8]
```

These seem more consistent with what Manos sees on the Cornell machines. I'm not sure what is different from when I ran the timing last time, but I'm not complaining 😄

VourMa:
I don't see any smoking gun in the updates for the decrease in the timing. @jkguiang could it be that you measured the timing for the master and the DNN-version at very different times?

PR #294, opened 2023-06-22
MD kernel parallelization and shared memory cleanup

slava77:
This is a result of my "day 1" so far.

- `createMiniDoubletsInGPUv2` launch had a 16 kB shared memory allocation requested but not used. Apparently declaring shared mem and not using it comes at a cost. After removing the first event MD time goes down by about 19% using the profiler and 10% using a minimum over 10 runs.
image

- we were suggested to be more aggressive and schedule more blocks; I've increased the number of blocks (from 80 to `825=nLowerModules/16`) to have the outermost stride-loop become a no-op (one iteration). The MD time goes down a bit: by 8% in profiler and 5% using a minimum over 10 runs. However, this comes with a significant cost in memory transfers, especially the L2-to-device.
image
The last part better be tested on some less performing GPU to confirm that the update is not a loss.

~I'm leaving this as a draft for now; perhaps a better strategy is to just split off the first commit as that one is non-controversial.~

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -946,7 +946,7 @@ void SDL::Event::createMiniDoublets()
     //dim3 nThreads(64,16,1);
     dim3 nBlocks(1,MAX_BLOCKS,1);
 
-    SDL::createMiniDoubletsInGPUv2<<<nBlocks,nThreads,64*4*16*sizeof(float),stream>>>(*modulesInGPU,*hitsInGPU,*mdsInGPU,*rangesInGPU);
+    SDL::createMiniDoubletsInGPUv2<<<nBlocks,nThreads,0,stream>>>(*modulesInGPU,*hitsInGPU,*mdsInGPU,*rangesInGPU);
     addMiniDoubletRangesToEventExplicit<<<1,1024,0,stream>>>(*modulesInGPU,*mdsInGPU, *rangesInGPU,*hitsInGPU);
 
     cudaError_t cudaerr = cudaGetLastError(); 
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -944,7 +944,7 @@ void SDL::Event::createMiniDoublets()
 
     dim3 nThreads(32,16,1);
     //dim3 nThreads(64,16,1);
-    dim3 nBlocks(1,MAX_BLOCKS,1);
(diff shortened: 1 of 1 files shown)


VourMa:
Out of curiosity, could you run the timing for multiple streams, just to see the scaling. Because, maybe, with the increased resources in the terms memory, the plateau of timing comes for smaller number of streams/is higher.

Partially note to self and maybe idea for this PR: What happens if we run the profiling while multistreaming?

pwittich:
Re the unused shared memory, this is likely due to this:

> In Volta the L1 cache, texture cache, and shared memory are backed by a combined 128 KB data cache. As in previous architectures, the portion of the cache dedicated to shared memory (known as the carveout) can be selected at runtime using cudaFuncSetAttribute() with the attribute cudaFuncAttributePreferredSharedMemoryCarveout.

https://docs.nvidia.com/cuda/volta-tuning-guide/index.html

I'm sure @slava77 knows this already...

GNiendorf:
I forgot about this, but on the Alpaka branch I think the shared memory issue is fixed. I never explicitly specify the amount of shared memory for a kernel launch, so I think that it allocates it automatically based on any shared variables in the kernel (probably?).

slava77:
I did some more somewhat mindless scanning of the block size (instead of 32,16) with `grid.y=nLowerModules/block.y` . I found a minimum at (128, 4), which takes 0.73 ms from profiler and 1.14 ms in "MD" column (compare this to 1.61 ms in profiler for a version in the "master" (head of the PUHack branch))

Calls to the device main memory increased further by 50%
image

slava77:
Here are timing results with multistreaming. The picture is less clear; I'm not sure how much of this is fluctuations.

The last one has the best timing in MD 1-thread as expected from my single-event FOM used for tuning. The multistream, even starting from 2 are kind of all over the place.

image

VourMa:
@slava77 Just checking, what is this is PR missing to be made into a non-draft PR?

slava77:
> @slava77 Just checking, what is this is PR missing to be made into a non-draft PR?

initially there was some concern about the possible cost of main memory reads, but then later during the hackathon when we were revisiting the profiler reports it didn't seem to be a significant concern.
So, I've changed this to "ready for review"

VourMa:
Closing, as this was superseeded by #300.

PR #295, opened 2023-06-22
Alpaka Kernel Parameter Exploration

GNiendorf:
Making this draft PR to make some changes to the kernel launch parameters on the Alpaka branch. First change is createSegmentsInGPUv2, moving from (13200,1,1)x(64,1,1) to (512,1,1)x(32,1,1)

Current Timing
current

This PR Timing
alpaka_timing

You can see a significant reduction in the LS timing above for s=1 and 2. The total Event time decreases for all streams. If we look at the profiler we see a significant increase in the hit rate, as well as a large decrease in the long scoreboard stalls and LG Throttle stalls. (blue is new, green is current)

prof_timing

scoreboard

roofline

stats

Screenshot 2023-06-22 at 10 23 19 AM

### DIFF
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1157,7 +1157,7 @@ void SDL::Event::createTriplets()
     cudaStreamSynchronize(stream);
 
     Vec const threadsPerBlockCreateTrip(static_cast<Idx>(1), static_cast<Idx>(16), static_cast<Idx>(16));
-    Vec const blocksPerGridCreateTrip(static_cast<Idx>(MAX_BLOCKS), static_cast<Idx>(1), static_cast<Idx>(1));
+    Vec const blocksPerGridCreateTrip(static_cast<Idx>(2048), static_cast<Idx>(1), static_cast<Idx>(1));
     WorkDiv const createTripletsInGPUv2_workDiv(blocksPerGridCreateTrip, threadsPerBlockCreateTrip, elementsPerThread);
 
     SDL::createTripletsInGPUv2 createTripletsInGPUv2_kernel;
--- a/SDL/Event.cu
+++ b/SDL/Event.cu
@@ -1059,8 +1059,8 @@ void SDL::Event::createSegmentsWithModuleMap()
         createSegmentsInExplicitMemory(*segmentsInGPU, nTotalSegments, nLowerModules, N_MAX_PIXEL_SEGMENTS_PER_MODULE,stream);
     }
 
-    Vec const threadsPerBlockCreateSeg(static_cast<Idx>(1), static_cast<Idx>(1), static_cast<Idx>(64));
-    Vec const blocksPerGridCreateSeg(static_cast<Idx>(1), static_cast<Idx>(1), static_cast<Idx>(nLowerModules));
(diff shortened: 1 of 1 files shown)


GNiendorf:
Motivated by the alpaka auto scheduler, I reduced the block size for a small kernel (moduleRangesKernel) and saw a performance increase.

Screenshot 2023-06-22 at 11 22 22 AM
Screenshot 2023-06-22 at 11 22 43 AM

GNiendorf:
Since the outer loop has a max around 8000, and Manos observed improvement from increasing the block size, I found that going to 2048 blocks gave a good speed improvement. 2.4 to 1.1 for single stream. Only downside is that the stall no instruction goes up.

New Timing
Screenshot 2023-06-22 at 1 38 05 PM

Screenshot 2023-06-22 at 1 48 55 PM

Screenshot 2023-06-22 at 1 49 28 PM
Screenshot 2023-06-22 at 1 49 47 PM
Screenshot 2023-06-22 at 1 50 35 PM

GNiendorf:
Plot

For single stream it looks like things level off at 1024 blocks, but looking at the avg time over all 175 events with s=8 using nvprof I get an avg of around 6.3ms for 1024 blocks and 4.4ms for 2048 blocks.

YonsiG:
I am a bit confused for the performance here. Do you mean increase the grid size(to 1048) and decrease the grid size(from 80 to 5) both gains good performance? Or did I misunderstand..

GNiendorf:
> I am a bit confused for the performance here. Do you mean increase the grid size(to 1048) and decrease the grid size(from 80 to 5) both gains good performance? Or did I misunderstand..

Yes, that's correct. The kernel where I saw a gain going from 80 to 5 blocks though is very small, so I think that has something to do with it.

YonsiG:
> I think 1024 may be a better block size for multistreaming since we get diminishing returns after that point. Only downside is that the stalls go up slightly, but I don't think this matters. I think it may be smart to set the block size by the number of streams as well, since for single stream it doesn't make sense to have the blocks set less than 8000 to give maximum performance.
> 
> Plot
> 
> Here3
> 
> Screenshot 2023-06-23 at 10 51 33 AM

Is the table above also 8 streams or single stream?

GNiendorf:
It seems like what matters (at least for the segments kernel) is that threads x blocks = limit loop, and beyond that there is a sweet spot at 32 threads (1 full warp).

p2

This is at 32 threads below.

p1

This is controlling the ratio of blocks to thread such that Nblocks = Outer loop max / threads

p3

Here is the comparison to my current best baseline at the top of this PR.

Screenshot 2023-06-23 at 1 22 46 PM
Screenshot 2023-06-23 at 1 22 32 PM
Screenshot 2023-06-23 at 1 22 17 PM
Screenshot 2023-06-23 at 1 17 38 PM

GNiendorf:
> Is the table above also 8 streams or single stream?

The first plot is single stream, the second is for 8 streams.

GNiendorf:
I'm going to close this PR for now and reopen a relevant issue when the Alpaka branch gets pushed into master.

PR #296, opened 2023-06-22
avoid the dPhi calculation in createTriplet

YonsiG:
The dPhi calculation is very time consuming from the compiler. (Which is the largest warp stalls of the Triplet kernel). We store already the dPhi information in the mdsInGPU, so we can avoid some of the dPhi calculation in creating the T3 and use the variable directly.
After this change, the timing of T3 kernel is decreased. From single stream to multi stream, the timing decrease is visible.
As a validation check, using Phi instead of computed from XY has the same physics performance as before.
master: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/fix_dphi/master_again_again_PU200_NEVT-1_b498de-PU200/compare/TC_eff_base_0.html
after this PR: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/fix_dphi/T3T5_removedPhi_PU200_NEVT-1_61a75dD-PU200/compare/TC_eff_base_0.html
The master timing is 
Screen Shot 2023-06-22 at 7 00 53 PM
The timing after using anchorPhi in Triplet kernels
Screen Shot 2023-06-22 at 6 57 15 PM

Furthermore, apply this change to the Quintuplet.cu kernels. The timing after using anchorPhi in Quintuplet kernels:
Screen Shot 2023-06-22 at 9 03 58 PM

The Segments kernels do not have many usage of the deltaPhi, optimizing it does not give obvious performance gains. Timing after change: 
Screen Shot 2023-06-28 at 6 10 36 PM

### DIFF
--- a/SDL/Hit.cuh
+++ b/SDL/Hit.cuh
@@ -65,12 +65,6 @@ namespace SDL
     void addHitToMemory(struct hits& hitsInCPU,struct modules& modulesInGPU,float x, float y, float z, unsigned int detId, unsigned int idxInNtuple,cudaStream_t stream,struct objectRanges& rangesInGPU);
     CUDA_G void addHitToMemoryGPU(struct hits& hitsInCPU,struct modules& modulesInGPU,float x, float y, float z, unsigned int detId, unsigned int idxInNtuple,unsigned int moduleIndex, float phis,struct objectRanges& rangesInGPU);
     
-    CUDA_HOSTDEV inline float ATan2(float y, float x) {
-      if (x != 0) return atan2f(y, x);
-      if (y == 0) return  0;
-      if (y >  0) return  float(M_PI) / 2.f;
-      else        return -float(M_PI) / 2.f;
-    }
     CUDA_HOSTDEV inline float eta(float x, float y, float z) {
       float r3 = std::sqrt( x*x + y*y + z*z );
       float rt = std::sqrt( x*x + y*y );
@@ -91,7 +85,7 @@ namespace SDL
       return x - n * float(2.f * float(M_PI));
     }
     CUDA_HOSTDEV inline float phi(float x, float y) {
-      return phi_mpi_pi(float(M_PI) + ATan2(-y, -x));
+      return atan2f(y,x);
     }
     CUDA_HOSTDEV inline float deltaPhi(float x1, float y1, float x2, float y2) {
       float phi1 = phi(x1,y1);
--- a/SDL/Hit.cuh
+++ b/SDL/Hit.cuh
@@ -85,7 +85,7 @@ namespace SDL
       return x - n * float(2.f * float(M_PI));
     }
     CUDA_HOSTDEV inline float phi(float x, float y) {
-      return atan2f(y,x);
+      return phi_mpi_pi(float(M_PI) + atan2f(y,x));
     }
     CUDA_HOSTDEV inline float deltaPhi(float x1, float y1, float x2, float y2) {
       float phi1 = phi(x1,y1);
--- a/SDL/MiniDoublet.cu
+++ b/SDL/MiniDoublet.cu
@@ -86,7 +86,7 @@ void SDL::createMDsInExplicitMemory(struct miniDoublets& mdsInGPU, unsigned int 
     mdsInGPU.nMDs = (unsigned int*)cms::cuda::allocate_device(dev, (nLowerModules + 1) *sizeof(unsigned int),stream);
     mdsInGPU.totOccupancyMDs = (unsigned int*)cms::cuda::allocate_device(dev, (nLowerModules + 1) *sizeof(unsigned int),stream);
     mdsInGPU.anchorX = (float*)cms::cuda::allocate_device(dev, nMemoryLocations * 6 * sizeof(float), stream);
-    mdsInGPU.anchorHighEdgeX = (float*)cms::cuda::allocate_device(dev, nMemoryLocations * 4 * sizeof(float), stream);
+    mdsInGPU.anchorHighEdgeX = (float*)cms::cuda::allocate_device(dev, nMemoryLocations * 6 * sizeof(float), stream);
     mdsInGPU.outerX = (float*)cms::cuda::allocate_device(dev, nMemoryLocations * 6 * sizeof(float), stream);
     mdsInGPU.outerHighEdgeX = (float*)cms::cuda::allocate_device(dev, nMemoryLocations * 4 * sizeof(float), stream);
 
@@ -99,7 +99,7 @@ void SDL::createMDsInExplicitMemory(struct miniDoublets& mdsInGPU, unsigned int 
     cudaMalloc(&mdsInGPU.nMDs, (nLowerModules + 1) * sizeof(unsigned int)); 
     cudaMalloc(&mdsInGPU.totOccupancyMDs, (nLowerModules + 1) * sizeof(unsigned int)); 
     cudaMalloc(&mdsInGPU.anchorX, nMemoryLocations * 6 * sizeof(float));
-    cudaMalloc(&mdsInGPU.anchorHighEdgeX, nMemoryLocations * 4 * sizeof(float));
+    cudaMalloc(&mdsInGPU.anchorHighEdgeX, nMemoryLocations * 6 * sizeof(float));
     cudaMalloc(&mdsInGPU.outerX, nMemoryLocations * 6 * sizeof(float));
     cudaMalloc(&mdsInGPU.outerHighEdgeX, nMemoryLocations * 4 * sizeof(float));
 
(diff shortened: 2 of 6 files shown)


SDL/Hit.cuh
```diff
@@ -91,7 +85,7 @@ namespace SDL
       return x - n * float(2.f * float(M_PI));
     }
     CUDA_HOSTDEV inline float phi(float x, float y) {
-      return phi_mpi_pi(float(M_PI) + ATan2(-y, -x));
+      return atan2f(y,x);
```
slava77, line 88: is it safe to avoid nans? Most of `ATan2` implementation was meant to regularize the result if inputs are 0

YonsiG: I removed it because the Alpaka branch removes it and still runs fine. I think that the atan2f function by C++ considers the case on x=0 internally, so we don't need to do it extra. Except that if we put (x,y)=(0,0), then atan2 function will give issues. But I think we don't have this in our case.

YonsiG:
> @YonsiG
> please prepare/post the profiler reports
profiler results in ogpu
profiler results in simplify dPhi:
http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/fix_dphi/profiling_fixdPhi2.ncu-rep
profiler results in master:
http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/fix_dphi/profiling_master_cgpu2.ncu-rep

SDL/Quintuplet.cu
```diff
@@ -1799,11 +1799,11 @@ __device__ bool SDL::runQuintupletDefaultAlgoBBBB(struct SDL::modules& modulesIn
     float tl_axis_lowEdge_x = tl_axis_x;
     float tl_axis_lowEdge_y = tl_axis_y;
 
-    betaIn = alpha_InLo - SDL::deltaPhi(mdsInGPU.anchorX[firstMDIndex], mdsInGPU.anchorY[firstMDIndex], tl_axis_x, tl_axis_y);
+    betaIn = alpha_InLo - SDL::phi_mpi_pi(atan2f(tl_axis_y,tl_axis_x)-mdsInGPU.anchorPhi[firstMDIndex]);
```
GNiendorf, line 1802: If you still have the phi function in Hit.cuh, maybe you should use that here instead of just applying atan2f directly. Just so that the code is more readable. Same comment for a few other places in the code.

YonsiG: Thank you for the comments Gavin! I agree that since I still have the phis, it makes no difference whether I use the phi function or the atan directly. I'm now using the atan expression because I feel like there is no clear meaning for this variable atan(tl_axis_y, tl_axis_x). It's not a "phi" for something but more of a temporary calculation. Do you think it make sense to leave it here?

YonsiG:
Adding the highedge and lowedge phi does not seem help from sdl_timing
Screen Shot 2023-06-28 at 7 32 33 PM

But if we read the profiler reports, it is decreasing the LS kernel time by ~0.01ms, T3 for ~0.1ms, T5 for ~0.15ms, at the cost of increase MD kernel time by 0.1ms. Going at the correct direction at least.
http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/fix_dphi/profiling_precomputedPhi.ncu-rep

VourMa:
@YonsiG I am unclear on the timeline for this PR. What else needs to be done and when do you foresee it to be ready for review? Or is it ready already?

VourMa:
@YonsiG As I am starting to review this, could you post some screenshots from the profiler results, so that people have a general picture of what you improved and can know where to look for more in the profiler? For example, it would be interesting to compare how the stalls change in the lines you modified.

SDL/Quintuplet.cu
```diff
@@ -1786,7 +1786,7 @@ __device__ bool SDL::runQuintupletDefaultAlgoBBBB(struct SDL::modules& modulesIn
 
     float alpha_OutUp,alpha_OutUp_highEdge,alpha_OutUp_lowEdge;
 
-    alpha_OutUp = SDL::deltaPhi(mdsInGPU.anchorX[fourthMDIndex], mdsInGPU.anchorY[fourthMDIndex], mdsInGPU.anchorX[fourthMDIndex] - mdsInGPU.anchorX[thirdMDIndex], mdsInGPU.anchorY[fourthMDIndex] - mdsInGPU.anchorY[thirdMDIndex]);
+    alpha_OutUp = SDL::phi_mpi_pi(atan2f(mdsInGPU.anchorY[fourthMDIndex] - mdsInGPU.anchorY[thirdMDIndex],mdsInGPU.anchorX[fourthMDIndex] - mdsInGPU.anchorX[thirdMDIndex])-mdsInGPU.anchorPhi[fourthMDIndex]);
```
VourMa, line 1789: Maybe use the `phi` function here to understand better what's going on, instead of the `atan2`?

Only later I noticed that Gavin also had a similar comment below. I didn't really understand your arguments why not change it. If we are putting this variable in a function called `phi_mpi_pi`, we want to interepret that as φ and I believe it would be nice to make that clear.

I guess that part of the reason you don't want to change `atan2` to `phi` is because it is in so many places of the code. In this case, I don't have a strong opinion. However, if we decide that we want some kind of protection (as mentioned in the comment above), then a switch to `phi` would be necessary.

YonsiG:
> @YonsiG As I am starting to review this, could you post some screenshots from the profiler results, so that people have a general picture of what you improved and can know where to look for more in the profiler? For example, it would be interesting to compare how the stalls change in the lines you modified.

I can paste a few profiler reports here for quick reference. This is a comparison of the T5 in master and createT5 after change. The green is the baseline master while the blue is current after change. The "stalls wait" and "stalls no instructions" have been reduced a lot, while the third longest "stall long scoreboard" increased a bit.
Screen Shot 2023-07-04 at 9 19 53 PM

The overall memory throughput is increased
Screen Shot 2023-07-04 at 9 22 38 PM

YonsiG:
This is the atan line in Hit.cuh, it has warp stats and instructions executed in create T5 kernels. It's a bit hard to read from the lines, since those number doesn't precisely match the warp usage in each line and they move around up and down. But still can see that the numbers generally decreased after changing 
After the change
Screen Shot 2023-07-04 at 9 24 22 PM
In master
Screen Shot 2023-07-04 at 9 25 21 PM

SDL/Hit.cuh
YonsiG: I think the usage of wrap it by a phi_mpi_pi function is to constrain the output angle into the range of (-pi,pi)
But I googled the usage of the atan function and the default output is (-pi, pi), so I think another protection is not necessary

YonsiG: I think the usage of wrap it by a phi_mpi_pi function is to constrain the output angle into the range of (-pi,pi)
But I googled the usage of the atan function and the default output is (-pi, pi), so I think another protection is not necessary

VourMa: As the messages above your last replies say, the whole discussion is about the ATan2 usage and the x=0, y=0 case. Given this,  I am confused about how the (-π,π) range fits into that discussion (maybe I missed some comment?).

But, in any case, if you are sure that `atan2` will work for the (0,0) case, fine by me.

VourMa:
Thanks for the profiler screenshots, they are useful. Could you modify the comment with the one where you show specific lines, so that you explain the color code and the different numbers?

SDL/Quintuplet.cu
YonsiG: For the idea of protection, I still think that we don't need it... We are giving the atan2 function an input of (x,y) and want it to return an angle between (-pi, pi), which is exactly the definition of the atan2 function in C++... So I'm a bit confused of why can't we simplify and what else do we need the phi to return..
For Gavin's suggestion of renaming all of them to "phi", I can change that to phi function and make it easy for people to understand.

SDL/Hit.cuh
YonsiG: I checked that the atan2 function on the (0,0) case, and found that it returns pi. so I added a protection on the phi_mpi_pi function. Thank you for insisting!

VourMa (approved):
Thanks a lot for the updates, @YonsiG! Merging this...

SDL/Hit.cuh
```diff
@@ -91,7 +85,7 @@ namespace SDL
       return x - n * float(2.f * float(M_PI));
     }
     CUDA_HOSTDEV inline float phi(float x, float y) {
-      return phi_mpi_pi(float(M_PI) + ATan2(-y, -x));
+      return phi_mpi_pi(float(M_PI) + atan2f(y,x));
```
slava77, line 88: is this correct?
e.g. pi/4 is (1,1)
- old : `atan2(-1,-1)+M_PI = 0.78539816`
- new : `atan2(1,1)+M_PI = 3.9269908`;  `phi_mpi_pi(3.9269908) = -2.35619`; there is a shift by π

VourMa: I'll take care of this as well.

SDL/MiniDoublet.cu
```diff
@@ -196,7 +198,8 @@ __device__ void SDL::addMDToMemory(struct miniDoublets& mdsInGPU, struct hits& h
     mdsInGPU.anchorHighEdgeY[idx] = hitsInGPU.highEdgeYs[anchorHitIndex];
     mdsInGPU.anchorLowEdgeX[idx] = hitsInGPU.lowEdgeXs[anchorHitIndex];
     mdsInGPU.anchorLowEdgeY[idx] = hitsInGPU.lowEdgeYs[anchorHitIndex];
-
+    mdsInGPU.anchorHighEdgePhi[idx] = atan2f(mdsInGPU.anchorHighEdgeY[idx], mdsInGPU.anchorHighEdgeX[idx]);
+    mdsInGPU.anchorLowEdgePhi[idx] = atan2f(mdsInGPU.anchorLowEdgeY[idx], mdsInGPU.anchorLowEdgeX[idx]);
```
VourMa, line 202: Should we be applying the phi function here now?

YonsiG: I think we can, using phi(x,y)

PR #297, opened 2023-06-23
Precompute custom layer variable

VourMa:
As per title, we identified some variables that were computed per kernel, even though they are properties of the modules and, hence, they can be computed up front. These variables are used in the Triplet and Quintuplet kernels.

**On cgpu-1 (A30)**
Before:
image

After:
image

I guess the timing is within the usual variations.

**Stall reduction**
Before (T3):
image

After (T3):
image

Before (T5):
image

After (T5):
image

### DIFF
--- a/SDL/Quintuplet.cu
+++ b/SDL/Quintuplet.cu
@@ -566,11 +566,11 @@ __device__ bool SDL::runQuintupletDefaultAlgo(struct SDL::modules& modulesInGPU,
 __device__ bool SDL::passChiSquaredConstraint(struct SDL::modules& modulesInGPU, uint16_t& lowerModuleIndex1, uint16_t& lowerModuleIndex2, uint16_t& lowerModuleIndex3, uint16_t& lowerModuleIndex4, uint16_t& lowerModuleIndex5, float& chiSquared)
 {
     //following Philip's layer number prescription
-    const int layer1 = modulesInGPU.layers[lowerModuleIndex1] + 6 * (modulesInGPU.subdets[lowerModuleIndex1] == SDL::Endcap) + 5 * (modulesInGPU.subdets[lowerModuleIndex1] == SDL::Endcap and modulesInGPU.moduleType[lowerModuleIndex1] == SDL::TwoS);
-    const int layer2 = modulesInGPU.layers[lowerModuleIndex2] + 6 * (modulesInGPU.subdets[lowerModuleIndex2] == SDL::Endcap) + 5 * (modulesInGPU.subdets[lowerModuleIndex2] == SDL::Endcap and modulesInGPU.moduleType[lowerModuleIndex2] == SDL::TwoS);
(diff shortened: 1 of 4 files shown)


GNiendorf:
What's the argument for precomputing these? I forget, but the modules.cu stuff isn't included in the timing right? So are we expecting this quantity to be on the GPU already when we integrate into CMSSW?

GNiendorf:
Is there any benefit to the kernel timing from the reduced stalls btw?

VourMa:
> Is there any benefit to the kernel timing from the reduced stalls btw? From the profiler.

Yes, about 7% - 1.98 to 1.84 ms on the A100 of the NVIDIA cluster.

YonsiG:
Thank you for this checks Manos! Gavin has asked some good questions and I think the commit is clear for me to merge.

YonsiG (approved):

SDL/Quintuplet.cu
```diff
@@ -709,11 +709,11 @@ __device__ bool SDL::passT5RZConstraint(struct SDL::modules& modulesInGPU, struc
...
-    const int layer2 = modulesInGPU.layers[lowerModuleIndex2] + 6 * (modulesInGPU.subdets[lowerModuleIndex2] == SDL::Endcap) + 5 * (modulesInGPU.subdets[lowerModuleIndex2] == SDL::Endcap and modulesInGPU.moduleType[lowerModuleIndex2] == SDL::TwoS);
-    const int layer3 = modulesInGPU.layers[lowerModuleIndex3] + 6 * (modulesInGPU.subdets[lowerModuleIndex3] == SDL::Endcap) + 5 * (modulesInGPU.subdets[lowerModuleIndex3] == SDL::Endcap and modulesInGPU.moduleType[lowerModuleIndex3] == SDL::TwoS);
-    const int layer4 = modulesInGPU.layers[lowerModuleIndex4] + 6 * (modulesInGPU.subdets[lowerModuleIndex4] == SDL::Endcap) + 5 * (modulesInGPU.subdets[lowerModuleIndex4] == SDL::Endcap and modulesInGPU.moduleType[lowerModuleIndex4] == SDL::TwoS);
-    const int layer5 = modulesInGPU.layers[lowerModuleIndex5] + 6 * (modulesInGPU.subdets[lowerModuleIndex5] == SDL::Endcap) + 5 * (modulesInGPU.subdets[lowerModuleIndex5] == SDL::Endcap and modulesInGPU.moduleType[lowerModuleIndex5] == SDL::TwoS);
+    const int layer1 = modulesInGPU.layers[lowerModuleIndex1];
```
slava77, line 712: is this the bug?

VourMa: Indeed, this must have been `sdlLayers`. I'm on it.

@GNiendorf I am fixing it on the master and the alpaka branch.

PR #298, opened 2023-06-23
LS optimization: early stop, const arrays in moduleGapSize, simpler logic in isTighterTiltedModules 

slava77:
- check inner module occupancy before looping over the segments: small speedup in one event test from 0.95 to 0.90 ms
- const arrays in moduleGapSize; provides most of the gain
-  logic in isTighterTiltedModules adds a few %

Summary:
- createSegmentsInGPUv2:
    - overall profiler for the first event: 0.76 ms -> 0.42 ms
    - register usage: 80 -> 72
- standard timing 
image

There was someone else on the 8-head A100 node during the tests above; so, fluctuations are possible.

### DIFF
--- a/SDL/Segment.cu
+++ b/SDL/Segment.cu
@@ -733,10 +733,10 @@ __device__ float SDL::moduleGapSize_seg(short layer, short ring, short subdet, s
 
 __device__ float SDL::moduleGapSize_seg(struct modules& modulesInGPU, unsigned int moduleIndex)
 {
-    float miniDeltaTilted[3] = {0.26f, 0.26f, 0.26f};
-    float miniDeltaFlat[6] ={0.26f, 0.16f, 0.16f, 0.18f, 0.18f, 0.18f};
-    float miniDeltaLooseTilted[3] = {0.4f,0.4f,0.4f};
-    float miniDeltaEndcap[5][15] = {
+    constexpr float miniDeltaTilted[3] = {0.26f, 0.26f, 0.26f};
+    constexpr float miniDeltaFlat[6] ={0.26f, 0.16f, 0.16f, 0.18f, 0.18f, 0.18f};
+    constexpr float miniDeltaLooseTilted[3] = {0.4f,0.4f,0.4f};
+    constexpr float miniDeltaEndcap[5][15] = {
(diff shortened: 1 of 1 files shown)


slava77:
I've updated the PR description with notes on the latest commits

slava77:
Summary using the nvidia cluster:
- createSegmentsInGPUv2:
    - overall profiler for the first event: 0.76 ms -> 0.42 ms
    - register usage: 80 -> 72
- standard timing 
image

There was someone else on the 8-head A100 node during the tests above; so, fluctuations are possible.

slava77:
@VourMa 
the memory use goes down indeed
image

SDL/Segment.cu
```diff
@@ -625,84 +625,34 @@ __device__ inline float SDL::isTighterTiltedModules_seg(struct modules& modulesI
...
-
+    return subdet == Barrel & ( ((side != Center) & (layer== 3))
+		| (side == NegZ & (((layer == 2) & (rod > 5)) | ((layer == 1) & (rod > 9))))
+		| (side == PosZ & (((layer == 2) & (rod < 8)) | ((layer == 1) & (rod < 4)))) );
 }
 
 //__device__ float SDL::moduleGapSize_seg(struct modules& modulesInGPU, unsigned int moduleIndex)
 __device__ float SDL::moduleGapSize_seg(short layer, short ring, short subdet, short side, short rod)
 {
-    float miniDeltaTilted[3] = {0.26f, 0.26f, 0.26f};
```
slava77, line 662: the same can be applied in MiniDoublet.cu as well (or consider using the same underlying function)

VourMa: Indeed. Also for the `isTighterTiltedModules`. Is this something you wanted to follow up on, or are you proposing that someone else does it?

slava77: at this point "for someone else"

slava77:
superseded by #301

SDL/Segment.cu
VourMa: Sure. I'll make an issue out of it and may follow-up in the next days, if I find some time.

PR #308, opened 2023-07-26
Fixing hackathon bugs

VourMa:
Validation that this worked (everything on non-DNN version):

- Current master (buggy): https://uaf-10.t2.ucsd.edu/~evourlio/SDL/hackathonFixes/beforeFixes_PU200_NEVT175_a8c352-PU200/mtv/var/TC_fakerate_etazoom.pdf
- Before bugs were introduced (as it should be) - took as reference commit 1b12ee5b8be13f9ebf6c9530873c3635eaf039f3, which has no DNN and no other performance changes in between: https://uaf-10.t2.ucsd.edu/~evourlio/SDL/hackathonFixes/beforeHackathonPRsAndDNN_PU200_NEVT175_1b12ee-PU200/mtv/var/TC_fakerate_etazoom.pdf
- After fixes (should agree with the one just above): https://uaf-10.t2.ucsd.edu/~evourlio/SDL/hackathonFixes/hackathonFixes_PU200_NEVT175_aed0d8-PU200/mtv/var/TC_fakerate_etazoom.pdf
- Comparison of the last two: https://uaf-10.t2.ucsd.edu/~evourlio/SDL/hackathonFixes/compare_beforeHackathonPRsAndDNN_VS_hackathonFixes_1b12ee-PU200_aed0d8-PU200/mtv/var/TC_base_0_0_eff_etazoom.pdf
- Folder with all validation plots: https://uaf-10.t2.ucsd.edu/~evourlio/SDL/hackathonFixes/

Tl;dr: After fixes, things agree with the non-buggy version.

### DIFF
--- a/SDL/Hit.cuh
+++ b/SDL/Hit.cuh
@@ -85,7 +85,7 @@ namespace SDL
       return x - n * float(2.f * float(M_PI));
     }
     CUDA_HOSTDEV inline float phi(float x, float y) {
-      return phi_mpi_pi(float(M_PI) + atan2f(y,x));
+      return phi_mpi_pi(float(M_PI) + atan2f(-y,-x));
     }
     CUDA_HOSTDEV inline float deltaPhi(float x1, float y1, float x2, float y2) {
       float phi1 = phi(x1,y1);
--- a/SDL/Quintuplet.cu
+++ b/SDL/Quintuplet.cu
@@ -622,11 +622,11 @@ __device__ bool SDL::passT5RZConstraint(struct SDL::modules& modulesInGPU, struc
     const float& z5 = mdsInGPU.anchorZ[fifthMDIndex]/100;
 
     //following Philip's layer number prescription
-    const int layer1 = modulesInGPU.layers[lowerModuleIndex1];
-    const int layer2 = modulesInGPU.layers[lowerModuleIndex2];
-    const int layer3 = modulesInGPU.layers[lowerModuleIndex3];
-    const int layer4 = modulesInGPU.layers[lowerModuleIndex4];
-    const int layer5 = modulesInGPU.layers[lowerModuleIndex5];
+    const int layer1 = modulesInGPU.sdlLayers[lowerModuleIndex1];
+    const int layer2 = modulesInGPU.sdlLayers[lowerModuleIndex2];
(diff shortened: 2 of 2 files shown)


SDL/Hit.cuh
```diff
@@ -85,7 +85,7 @@ namespace SDL
       return x - n * float(2.f * float(M_PI));
     }
     CUDA_HOSTDEV inline float phi(float x, float y) {
-      return phi_mpi_pi(float(M_PI) + atan2f(y,x));
+      return phi_mpi_pi(float(M_PI) + atan2f(-y,-x));
```
slava77, line 88: is `phi_mpi_pi` even needed here? I missed the reason why 32c54ff278f9686674bb35bce92ca1512602e4f5 was applied here originally. Why are we computing [0, 2π] range first so that it can be immediately converted to [-π,π] ?

YonsiG: I think 32c544ff commit written a phi_mpi_pi is a roll back of commit #296 61a75dc (under the same PR). Before the PR 296 of changing dPhi, the original written way was "phi_mpi_pi(float(M_PI) + Atan2f(-y,-x));"  I meant to follow the previous way in Hit.cuh, but somehow didn't do it correctly.
I also think it's weird that the range goes to [0,2pi] first, since the atan2f() function returns [-pi,pi] directly..

YonsiG: I changed the function into "return ata2f(y,x)" and get the fake rate plot as this.
http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/fix_dphi/simpler_phi_PU200_NEVT-1_aed0d8D-PU200/mtv/var/TC_fakerate_etazoom.pdf
Compared to Manos's fake rate eta plots after fix, they are consistent with each other. Seems it's safe to use the simpler one

VourMa: Please do not add any update at this point. New updates should go in the `master` branch, after it is migrated to alpaka.

VourMa:
Superseded by #310

SDL/Hit.cuh
slava77: > Compared to Manos's fake rate eta plots after fix, they are consistent

a comparison plot would be better, identical results are expected (within limits of reproducibility)

PR #314, opened 2023-07-31
Move to Most Recent Alpaka Version + Newest Caching Allocator

GNiendorf:
Moves to gcc11 and CMSSW_13_0_0_pre4 (same as current CMSSW integration) in order to upgrade to the most recent version of Alpaka used by CMSSW. Also gets rid of our temporary hyperbolic functions and replaces them with Alpaka's which were introduced into Alpaka recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ cd CMSSW_13_0_0_pre4/src
 cmsenv
 git cms-init
 git remote add SegLink git@github.com:SegmentLinking/cmssw.git
-git fetch SegLink CMSSW_13_0_0_pre4_LST_X
+git fetch SegLink CMSSW_13_0_0_pre4_LST_X_alpaka
 git cms-addpkg RecoTracker Configuration
 git checkout CMSSW_13_0_0_pre4_LST_X
 cat <<EOF >lst.xml
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ git cms-init
 git remote add SegLink git@github.com:SegmentLinking/cmssw.git
 git fetch SegLink CMSSW_13_0_0_pre4_LST_X_alpaka
 git cms-addpkg RecoTracker Configuration
-git checkout CMSSW_13_0_0_pre4_LST_X
+git checkout CMSSW_13_0_0_pre4_LST_X_alpaka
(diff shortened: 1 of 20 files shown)


GNiendorf:
Comparison plots: [Here](https://www.classe.cornell.edu/~gsn27/www/www/PR314/mywork_7ec624-PU200_f93565-PU200/summary/)
This PR: [Here](https://www.classe.cornell.edu/~gsn27/www/www/PR314/mywork_f93565-PU200/summary/)
Current Master: [Here](https://www.classe.cornell.edu/~gsn27/www/www/PR314/mywork_7ec624-PU200/summary/)

Given that Philip is probably busy, I'll leave it to them to open a PR later on if this configuration doesn't work on their machine.

YonsiG:
Hi Gavin, sorry for coming to the PR late. I am wondering about is this enough for us to have setup.sh for both UCSD machine and Cornell machine? Previously we had setup_ucsd.sh and setup_lnx7188.sh 2 versions. In alpaka do we need to have separate setup files also?

YonsiG:
Hi Gavin, sorry for the late review and thank you for the efforts! I checked the external LST package and it runs smoothly. However, I met an error when compiling the LST set up as CMSSW external. 

When I did scram b -j 12, after the cms-checkdeps in README, it says:
/home/users/yagu/TrackLooper_Ntuple/test_PR314/workingFolder/CMSSW_13_0_0_pre4/src/RecoTracker/LST/plugins/alpaka/LSTProducer.cc:46:43: error: cannot bind non-const lvalue reference of type 'CUstream_st*&' to an rvalue of type 'CUstream_st*'
   46 |       lst_.run(ctx.queue().getNativeHandle(),


I'm not sure if the error comes with issues in this PR, but it would be nice if you have any ideas.

GNiendorf:
@YonsiG Just updated the README as you suggested. The CMSSW integration steps should work now.

YonsiG:
Thanks Gavin, I tested the codes on external packages, and test the CMSSW integration with cmsRun. Everything works smoothly on my side. Thanks for contribution and PR merged!

YonsiG (approved):

Issue #317, opened 2023-08-23
Write out code too fragile to changes in the pLS


VourMa:
As part of the CMSSW investigations, we thought that getting rid of the pLS duplicate cleaning could give us performance improvements for the patatrack configuration, and indeed it did. As a result, it is important to have the capability of switching between setups with and without the pLS duplicate cleaning enabled.

I started from the `cuda_branch` and, to get the maximum improvement gain, I commented out both calls to the pLS duplicate cleaning kernel (`checkHitspLS`):
- https://github.com/SegmentLinking/TrackLooper/blob/d559e31826748f2067609173d4f31de28fd6bb37/SDL/Event.cu#L1309
- https://github.com/SegmentLinking/TrackLooper/blob/d559e31826748f2067609173d4f31de28fd6bb37/SDL/Event.cu#L1111

Things in CMSSW ran fine. Then I wanted to test this in the standalone. Running with
```sdl -i PU200 -n 10 -w 0```
worked out fine. When I ran with
```sdl -i PU200 -n 10 -w 2```
I get the error:
```
Thread 1 (Thread 0x7f079bb76000 (LWP 351809) "sdl"):
#0  0x00007f079d96f402 in waitpid () from /lib64/libc.so.6
#1  0x00007f079d8d0bff in do_system () from /lib64/libc.so.6
#2  0x00007f07a085cd1b in TUnixSystem::StackTrace() () from /cvmfs/cms.cern.ch/el8_amd64_gcc10/cms/cmssw/CMSSW_12_5_0_pre5/external/el8_amd64_gcc10/lib/libCore.so
#3  0x00007f07a085a2b5 in TUnixSystem::DispatchSignals(ESignals) () from /cvmfs/cms.cern.ch/el8_amd64_gcc10/cms/cmssw/CMSSW_12_5_0_pre5/external/el8_amd64_gcc10/lib/libCore.so
#4  <signal handler called>
#5  matchedSimTrkIdxs (hitidxs=<error reading variable: Cannot access memory at address 0x0>, hittypes=..., verbose=false) at code/core/trkCore.cc:448
#6  0x000000000047f79a in parseTrackCandidate (event=0x7fff4625a360, idx=0) at code/core/write_sdl_ntuple.cc:806
#7  0x00000000004817bc in setOutputBranches (event=0x0) at code/core/write_sdl_ntuple.cc:243
[...]

PR #318, opened 2023-08-24
Compilation flag to run no pLS duplicate cleaning (CUDA version)

VourMa:
As per title, to allow exploring more configurations when using patatrack pixel tracks within CMSSW.

This works, checked by print outs that the flag is enabled when it is supposed to, and that there is the expected increase of number of pLS object. A full validation is not possible due to #317, and would in any case be a bit meaningless, given that the results are bound to be totally different(worse) than the ones we have now. Let's discuss if this should go in, or just stay as an extra branch to be used for the HLT tests, and when.

To be ported to master when ready.


VourMa:
This PR was useful for HLT configurations for the HLT annual review, which was based on the CUDA branch. Now that the HLT annual review is past us, I don't think it makes sense to continue development on non-master branches, so I am closing this PR, I will port the changes to the master branch, as they can be useful for LST configurations using Patatrack pixel tracks as input.

PR #319, opened 2023-08-24
Fixed indexing bug in crossCleanT5 kernel

ariostas:

PR #320, opened 2023-08-24
Fixed indexing bug in crossCleanT5 kernel

ariostas:


ariostas:
See #319 for comparisons and performance plots.

VourMa:
Just a comment on this merge: For the CMSSW integration, I have proposed that the `cuda_branch_callOnce` branch is used. This is because that branch includes the update to easily disable the pLS duplicate cleaning (see PR #318)  but it is pending understanding/decision on the (unwanted?) truncation of TCs (see issue #317). In case we want the update from this PR to go in the HLT results, some care needs to be taken which branch to use and with which updates.

PR #321, opened 2023-08-30
Remove Unused pt5 Code

GNiendorf:
None of this code is used, so it should probably be removed to avoid confusion.


YonsiG:
After checking these codes, running the pT5 building don't give error when removing these functions. PR merged.

YonsiG (approved):

PR #313, opened 2023-07-30
Add CPU Backend

GNiendorf:
Add a toggle to the current master branch that allows you to run on the code on the CPU. Largest difference is that kernel inputs must be passed by value rather than by reference. I also put the grid/thread sizes in a wrapper function to reduce code boilerplate and allow for a CPU serial setup. You can still parallelize across events though by changing the stream size.

It may make sense to transition to the wrapper functions found in the alpaka_interface, but I have to look into that more.

PR #323, opened 2023-09-14
checkHitspLS fixes

ariostas:
There were two issues that I found in the checkHitspLS kernel:

1. When running it with `secondpass = true`, it checks whether the segments are duplicates during the loop. When it is checked for index `ix` it's fine because this is the only index that could be marked as a duplicate, so it can be skipped if it's already known that it is a duplicate. However, for the one at index `jx` it is a bit ambiguous what should be done. If we check if `jx` is a duplicate and it is skipped if so (as it is currently done) then this process depends on parallelization since one thread can mark it as duplicate before some other thread checks for it. This is what what causing the difference in duplicate rates between CPU and GPU. However, it is also not clear if it should just not be checked since presumably it might end up removing extra things that should have been kept. For now, I removed the second check, which makes CPU and GPU runs agree, but an alternative approach is possible.
2. The logic of [this line](https://github.com/SegmentLinking/TrackLooper/blob/8dbf8b213b272ab7fef5f7df071c5dfa7f7b5623/SDL/Kernels.h#L543) was being interpreted by the compiler as
```
npMatched >= 1 or (dR2 < 0.00075f and (ix < jx))
```
whereas the intended logic was presumably
```
(npMatched >= 1 or dR2 < 0.00075f) and (ix < jx)
```
Looking back at when this was introduced by [this PR](https://github.com/SegmentLinking/TrackLooper/pull/151/) corroborates this. However, adding parenthesis as in the second option it ends up producing a huge increase in duplicate rates. This is because the old logic didn't use the `ix < jx` condition. So I also removed it to match the previous logic.

The result of these changes is a huge decrease in duplicate rates, but it also comes with a small decrease in efficiencies. It's not clear to me whether this drop in efficiency is worth it, or if we should try to find a better fix for the above issues.

Here is how the drop in duplicate rates looks:
image

And this is how the drop in efficiencies looks:
image

The CPU and GPU runs now agree almost perfectly. Here is a comparison of duplicate rates:
image

Here is the timing before and after these changes:
Screenshot 2023-09-14 at 2 40 43 PM
Screenshot 2023-09-14 at 2 32 58 PM

All of the comparisons were done with the changes in #313 (which is still not merged at the time of writing) on the master branch.

Edit: Removed link to comparison plots, since now the correspond to a different change.

### DIFF
--- a/SDL/Kernels.h
+++ b/SDL/Kernels.h
@@ -479,13 +479,10 @@ namespace SDL
                 float eta_pix1 = segmentsInGPU.eta[ix];
                 float phi_pix1 = segmentsInGPU.phi[ix];
 
-                for(int jx = globalThreadIdx[2]; jx < nPixelSegments; jx += gridThreadExtent[2])
+                for(int jx = ix + 1 + globalThreadIdx[2]; jx < nPixelSegments; jx += gridThreadExtent[2])
                 {
                     float eta_pix2 = segmentsInGPU.eta[jx];
                     float phi_pix2 = segmentsInGPU.phi[jx];
-
-                    if (ix == jx)
-                        continue;
 
                     if (alpaka::math::abs(acc, eta_pix2 - eta_pix1) > 0.1f)
                         continue;
@@ -494,11 +491,14 @@ namespace SDL
                         continue;
 
                     char quad_diff = segmentsInGPU.isQuad[ix] - segmentsInGPU.isQuad[jx];
-                    float ptErr_diff = segmentsInGPU.ptIn[ix] - segmentsInGPU.ptIn[jx];
                     float score_diff = segmentsInGPU.score[ix] - segmentsInGPU.score[jx];
-                    // Always keep quads over trips. If they are the same, we want the object with the lower pt Error
-                    if((quad_diff > 0) || (score_diff < 0 && quad_diff == 0))
-                        continue;
+                    // Always keep quads over trips. If they are the same, we want the object with better score
+                    int idxToRemove;
+                    if (quad_diff > 0) idxToRemove = jx;
+                    else if (quad_diff < 0) idxToRemove = ix;
+                    else if (score_diff < 0) idxToRemove = jx;
+                    else if (score_diff > 0) idxToRemove = ix;
+                    else idxToRemove = ix;
 
                     unsigned int phits2[4];
                     phits2[0] = segmentsInGPU.pLSHitsIdxs[jx].x;
@@ -525,14 +525,9 @@ namespace SDL
                                 break;
                         }
                     }
-                    // If exact match, remove only 1
-                    if((npMatched == 4) && (ix < jx))
+                    if(npMatched >= 3)
                     {
-                        rmPixelSegmentFromMemory(segmentsInGPU, ix);
-                    }
-                    if(npMatched == 3)
-                    {
-                        rmPixelSegmentFromMemory(segmentsInGPU, ix);
+                        rmPixelSegmentFromMemory(segmentsInGPU, idxToRemove);
                     }
                     if(secondpass)
                     {
@@ -540,9 +535,9 @@ namespace SDL
                         float dPhi = SDL::calculate_dPhi(phi_pix1, phi_pix2);
 
                         float dR2 = dEta*dEta + dPhi*dPhi;
-                        if(( (npMatched >= 1) || (dR2 < 0.00075f) ) && (ix < jx))
+                        if((npMatched >= 1) || (dR2 < 1e-5f))
                         {
-                            rmPixelSegmentFromMemory(segmentsInGPU, ix); 
+                            rmPixelSegmentFromMemory(segmentsInGPU, idxToRemove);
                         }
                     }
                 }
--- a/SDL/Kernels.h
+++ b/SDL/Kernels.h
@@ -540,7 +540,7 @@ namespace SDL
                         float dPhi = SDL::calculate_dPhi(phi_pix1, phi_pix2);
 
                         float dR2 = dEta*dEta + dPhi*dPhi;
-                        if(npMatched >= 1 or dR2 < 0.00075f and (ix < jx))
+                        if(( (npMatched >= 1) || (dR2 < 0.00075f) ) && (ix < jx))
                         {
                             rmPixelSegmentFromMemory(segmentsInGPU, ix); 
                         }


slava77:
> The result of these changes is a huge decrease in duplicate rates, but it also comes with a small decrease in efficiencies.

for efficiencies, even a 1% is already a large loss; here we have 10% drop.

ariostas:
> for efficiencies, even a 1% is already a large loss; here we have 10% drop.

Yeah, I'll think of a way to remedy this.

slava77 (commented):
doesn't this update remove all matching/overlapping pLSs?
The idea is to pick one if there is an overlap.

slava77:
> It is certainly overly aggressive, so it might remove all or most of them. The tricky part is to get it to pick one in a consistent way that doesn't depend on the order that threads run.

some other property can be used other than the index. e.g. pt (pick higher), ptError (pick smaller), or the innermost hit position (pick closer to the beam).
At least for 2-way overlaps it should be predictable. 
The 3+ way cases will still have ordering problems.

ariostas:
Here's the issue that I see. If you leave the `isDup` check in the nested for loop and fix the logic of the duplicate removal to
```
(npMatched >= 1 or dR2 < 0.00075f) and (ix < jx)
```
then again it is consistent. But then the efficiency still decreases
image
and there is a giant increase in duplicate rates.
image

I don't think that using pt or something else would fix this. It seems to me that the old code was semi-randomly removing things and it was luckily better than doing it consistently (unless a more significant change is made).

SDL/Kernels.h
```diff
@@ -490,7 +490,7 @@ namespace SDL
                     if (alpaka::math::abs(acc, eta_pix2 - eta_pix1) > 0.1f)
                         continue;
 
-                    if (secondpass && (!segmentsInGPU.isQuad[jx] || segmentsInGPU.isDup[jx]))
+                    if (secondpass && !segmentsInGPU.isQuad[jx])
```
slava77, line 493: what's the impact of this change on efficiency?

ariostas: Do you mean as in timing? If so, it seems to not be too much as shown by the timings I posted in the first comment.

If you mean reconstruction efficiency, then this is what can break it and make it remove all pLSs.

slava77: > If you mean reconstruction efficiency, then this is what can break it and make it remove all pLSs.

yes, my question was `what's the impact of this change on efficiency?`

If I understand correctly, the test against `isDup` here includes the first pass marking and removal of the check allows to consider all original pLSs. I'm not sure this is appropriate in the second pass.

if more rigorous dual info (from the first and second pass) is needed, then perhaps a change should be made from `bool* isDup` to `char* isDup` and set the value to 1 in the first pass and 2 in the second

ariostas: Oh yeah, that's a good point. I'll check tomorrow and let's talk about it in the meeting.

ariostas: If you make this change you end up with the same results as here: https://github.com/SegmentLinking/TrackLooper/pull/323#issuecomment-1720059884

I'm completely onboard with leaving the `isDup` check as it is or as a more robust version. What I think we should focus on is that the `npMatched >= 1 or dR2 < 0.00075f and (ix < jx)` statement was wrong and fixing it is very negatively affecting the results.

ariostas:
I went back to only fixing the logic of the if statement. My previous comment showed that there was a significant decrease in efficiency, but seems like it was just due to run-to-run variations that can happen with the wrong logic and processing it in parallel. Doing the comparison on CPU gives a better idea of how things change. Here are the plots comparing efficiency and duplicate rates on CPU.
image
image

It's clear that the efficiency stays pretty much the same.

Unfortunately, the duplicate rate still increases very dramatically. I tried to remedy this by changing the `dR2 < 0.00075f` cut and also by adjusting the `dR2` cuts in `crossCleanpLS`, but both of these changes result in significantly decreased efficiency. So we might need to come up with a new way to remove pLS duplicates.

Although in principle it is still possible to have slight differences with this fixed logic, the CPU and GPU plots now agree very well.
image
image

The full set of comparison plots can be found at [this link](https://www.classe.cornell.edu/~ar2285/www/LST/PR323/).

ariostas:
Actually, I just realized that the logic in the if statement is inconsistent with previous logic in the code. I'll fix this and see what happens.

ariostas:
Hopefully this is the last big update. I realized that the logic of the if statement was in conflict with this piece.
https://github.com/SegmentLinking/TrackLooper/blob/8dbf8b213b272ab7fef5f7df071c5dfa7f7b5623/SDL/Kernels.h#L500-L501
The issue is that those lines sometimes offload the duplicate removal to the loop iteration with `ix` and `jx` switched, but due to the `ix<jx` condition in the `if` statement these objects were not always removed. Now, it first marks which of `ix` or `jx` would be a better candidate to remove, and if the right conditions are satisfied then it is marked as duplicate.

I also changed the loop ranges so that only one combination is considered, which has the advantage of having some thread blocks finish much earlier. I expected this to give a significant speedup in multi-stream runs, but it seems like it doesn't make much of a difference. These are the timings before and after.
Screenshot 2023-09-19 at 12 15 49 PM
Screenshot 2023-09-19 at 12 22 13 PM

Here is how the efficiency and duplicate rates look now. (these are using CPU)
image
image

So there is an increase in efficiency, but also a bit of an increase in duplicate rates. These two can be tweaked by adjusting the `dR2` cut, which I decreased from `0.00075` to `1e-5`.  If left at the original value both the efficiency and duplicate rate were lower than before. The choice of `1e-5` worked reasonable well, and it's one that is used in other parts of the code like in the `crossCleanpT3` kernel was), but we can set it to around `1e-4` if we want to keep efficiency and duplicate rates similar to before.

Ironically, there is again a bit of a difference in duplicate rates between  as shown below. I could play around to see if I can make them match, but I think it's probably fine since at least we understand what's happening and now we know that it is using correct logic.
image

The full set of comparison plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR323/).

ariostas:
Quick update: if you remove the `dR2 < 1e-5` cut the efficiency doesn't get any better, but the duplicate rate does increase a little bit. So I think this is a reasonable value to keep. I'll mark the PR as ready to review now.

SDL/Kernels.h
```diff
@@ -479,26 +479,26 @@ namespace SDL
...
-                        continue;
-
                     if (alpaka::math::abs(acc, eta_pix2 - eta_pix1) > 0.1f)
                         continue;
 
                     if (secondpass && (!segmentsInGPU.isQuad[jx] || segmentsInGPU.isDup[jx]))
                         continue;
 
                     char quad_diff = segmentsInGPU.isQuad[ix] - segmentsInGPU.isQuad[jx];
-                    float ptErr_diff = segmentsInGPU.ptIn[ix] - segmentsInGPU.ptIn[jx];
```
YonsiG, line 497: Why this line is removed? Just curious

SDL/Kernels.h
```diff
@@ -479,26 +479,26 @@ namespace SDL
...
                     float score_diff = segmentsInGPU.score[ix] - segmentsInGPU.score[jx];
-                    // Always keep quads over trips. If they are the same, we want the object with the lower pt Error
-                    if((quad_diff > 0) || (score_diff < 0 && quad_diff == 0))
```
YonsiG, line 500: Hi Ariostas, I am not familiar with the thread competing issue, but I'm curious to learn about this. In the original code, you said that when different thread checking jx in parallel, some thread will find a duplicate, and it "continue" the second loop. But the other thread will still do it's work and check the jx,  right? Can you explain that why it's removing extra things?

ariostas: I think what happens is something like this. Let's say that we have segments with indices 0,1,2, such that this happens:
- when ix=1, jx=0 then 1 is marked as duplicate
- when ix=2, jx=1 then 2 is marked as duplicate
- with all other combinations nothing is marked as duplicate

If we do this process serially, then we start by checking ix=0, and nothing is marked as duplicate, Then we go to ix=1, jx=0 and 1 is marked as duplicate. Then we get to the iteration with ix=2, jx=1, but since 1 was already marked as duplicate then the loop continues. So in the end the segments with indices 0 and 2 remain.

If we do this in parallel, the following can happen. The loop iteration with ix=2, jx=1 can start before the one with ix=1, jx=2 (or at the same time), and since 1 hasn't been marked as duplicate yet then both iterations proceed. These two threads will mark 1 and 2 as duplicate, so in the end only the segment with index 0 will remain.

So the parallel computation can remove extra things, but it seems like the effect is fairly minor, so it's probably acceptable.

SDL/Kernels.h
```diff
@@ -479,26 +479,26 @@ namespace SDL
...
                     float phi_pix2 = segmentsInGPU.phi[jx];
 
-                    if (ix == jx)
-                        continue;
-
                     if (alpaka::math::abs(acc, eta_pix2 - eta_pix1) > 0.1f)
                         continue;
 
                     if (secondpass && (!segmentsInGPU.isQuad[jx] || segmentsInGPU.isDup[jx]))
                         continue;
```
YonsiG, line 491: I'm also curious about the original code here, why do we think that, if a pLS is not quad, (which means it's a triplet), or it is duplicated, we continue and don't remove it?

ariostas: That's a good question. I'm also not sure why that's done. We should ask someone more familiar with this part of the code.

VourMa: @YonsiG We skip adding the triplet and duplicate pLS in the TC candidates (in the `addpLSasTrackCandidateInGPU` kernel), so I believe the logic is to not waste time checking objects that are not going to be used in any case.

In the meeting, the idea to mark all of the objects to be skipped with a single variable, instead of having multiple flags in multiple places. I will make an issue out of that, so that we keep it in mind.

YonsiG: Thank you for explaining Manos!

YonsiG (approved):

Issue #327, opened 2023-09-27
Debug mode break when running sdl


YonsiG:
I am trying to use the debug mode for testing output branches, so a compilation flag -d is added. The compilation is successful, but when I run the command "sdl -i PU200 -v 0 -w 2 -n 100", it showed runtime error.

terminate called after throwing an instance of 'std::runtime_error'
  what():  /cvmfs/cms.cern.ch/el8_amd64_gcc11/external/alpaka/develop-20230621-9e2225ac6c979464a40749ef9d1e0331/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp(153) 'TApi::memcpyAsync( m_dstMemNative, m_srcMemNative, static_cast<std::size_t>(m_extentWidthBytes), m_uniformMemCpyKind, queue.getNativeHandle())' returned error  : 'cudaErrorInvalidValue': 'invalid argument'!
Aborted (core dumped)

The debug mode can be run in the original cuda version, not sure which PR introduces this issue.

PR #328, opened 2023-09-29
Fix debug mode

GNiendorf:
This fixes the crash found in Issue #327, where two memcopy's are trying to transfer data larger than the array size since those arrays were destacked when moved over to Alpaka.


GNiendorf:
@YonsiG Can you check that the output ntuple looks good before this is merged? Thank you.

YonsiG:
> @YonsiG Can you check that the output ntuple looks good before this is merged? Thank you.

Hi Gavin, I checked that before and after this PR the output ntuple looks good. And after this PR, the debug branches can be stored in the root files normally. Thanks!

YonsiG (approved):

PR #330, opened 2023-10-03
Minor cleanup

ariostas:
I went through the code and did some minor cleanup, including typo fixes and removal of unused code. I also added checks to some else statements to make sure things are working properly.

I'd say that the most important change that may need some discussion was changing a bunch of bitwise operators to logical operators. The logic is unchanged, but I'd say it's better practice to use logical operators unless there is a definite performance advantage, which seems like it's not really the case here. (The timings actually look a bit better, but it's likely just run-to-run variations.)

Here are the timings before and after:
Screenshot 2023-10-02 at 2 39 12 PM
Screenshot 2023-10-02 at 3 00 57 PM

And [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR330/) are the comparison plots to make sure that nothing was broken.


GNiendorf:
Looks good to me. Could you post screenshots from the CUDA profiler for one of the cases where you switched from bitwise to logical?

ariostas:
I'm not very familiar with `nvvp`, but here are screenshots comparing this part that has a bunch of these changes together.
https://github.com/SegmentLinking/TrackLooper/blob/1f11908ef24b5efedfcdccffd0ce8a581b0ef2b3/SDL/Segment.h#L186-L188

Before
Screenshot 2023-10-05 at 9 37 55 AM
After
Screenshot 2023-10-05 at 9 38 39 AM

It's clear that the current implementation with bitwise operators involves a lot less instructions, whereas the new changes result in a much larger block with a bunch of branching. However, it seems like the new changes have better latency since it doesn't have to wait for memory. So I'm not sure which of the two options ends up being better.

Also, one thing that I'm confused about is why the divergent branching (bottom panel) ends seems better for the new changes even though I thought it should be worse. So I'm not sure if I'm looking at the wrong thing.

I think we should just make a decision and try to be consistent. There were places that mixed the two types of operators. This was an example:
https://github.com/SegmentLinking/TrackLooper/blob/72fe680e2a449e00b9159e9003438e69ca3c9df5/SDL/Triplet.h#L426
And also, one minor improvement that could be made in cases like this one is to not bother anding with the previous value of `pass` since it would have returned early if it was `false`.

GNiendorf:
Yeah I agree, we're probably splitting hairs here. This PR looks good, so I'm merging it in. Thank you for the cleanup.

GNiendorf (approved):

PR #333, opened 2023-10-04
Make the DNN the default (the -N flag disables it)

VourMa:
Small change in the compilation flags, so that the T5 DNN selection runs by default, while the `-N` flag is not inverted to disable it. The changes have been validated by means of printouts, and they seem to be working fine.

I also added another extension to the `.gitignore` file to simplify the `git status` output.

When merging, please select to also squash the commits with the description of the last one, so as not to include the confusing intermediate commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ movetoweb.sh
 *.nsys-rep
 *.sqlite
 *.ncu-rep
+*.swp
 
 *.nfs*
 .directoryhash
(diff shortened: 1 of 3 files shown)


ariostas:
I'd say that the default behavior of the `sdl` binary and the `sdl_timing` script should match. So either both should have a toggle to activate the DNN, or have it on by default and have a toggle to disable it. Alternatively, the timing script could run both versions, but it's probably not worth it since the difference will probably stay constant.

jkguiang (approved):

PR #336, opened 2023-10-10
Update library paths

ariostas:
While trying to build a self-contained image for Github Actions to use, I noticed that the library paths don't match what the chosen version of SCRAM/CMSSW uses. Everything seems to be working fine, but it might be possible to cause unexpected crashes due to version differences.

I only updated the paths to match the current SCRAM/CMSSW version (el8_amd64_gcc11/13_0_0_pre4), but it would be nice to have a more permanent fix. For CUDA/Boost/ROOT we could do something like this.
```bash
export CUDA_ROOT="$(readlink -f /cvmfs/cms.cern.ch/$SCRAM_ARCH/cms/cmssw/$CMSSW_VERSION/external/$SCRAM_ARCH/bin/nvcc | rev | cut -d'/' -f 1,2 --complement | rev)"
```
However, this doesn't work with Alpaka. In general, the selected versions are in xml files in `/cvmfs/cms.cern.ch/$SCRAM_ARCH/cms/cmssw/$CMSSW_VERSION/config/toolbox/$SCRAM_ARCH/tools/selected/`, but I don't know of an easy way to extract them.

### DIFF
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -21,7 +21,7 @@ CXXFLAGS_CUDA        = -g --compiler-options -Wall --compiler-options -Wshadow -
 ALPAKAINCLUDE        = -I${ALPAKA_ROOT}/include -I/${BOOST_ROOT}/include -std=c++17
 ALPAKASERIAL         = -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
 ALPAKACUDA           = -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_ACC_GPU_CUDA_ONLY --expt-relaxed-constexpr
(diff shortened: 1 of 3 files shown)


GNiendorf:
I chose this Alpaka version because it was the most up to date version available. Changing it to an older version of Alpaka would also cause a mismatch between the caching allocator we use (since it is based on our current Alpaka version). I would be surprised if this PR runs to be honest, because I thought there were some things only introduced in the most recent version of Alpaka that we make use of in our code.

ariostas:
@GNiendorf Okay, then we definitely shouldn't change the version of Alpaka. I'm also surprised that it seems to run fine considering the mismatch. What do you think about the other changes?

GNiendorf:
I like your idea of getting rid of hard-coded paths. We can always move to a higher version of CMSSW in our setup scrips that uses the most recent version of Alpaka if we want to go down that road. Manos and Slava probably know what is best here though.

slava77:
If the goal is to simplify the setup substantially, one option is to simply use the CMSSW software stack and not decompose the parts.

ariostas:
This PR was superseded by #348.

PR #338, opened 2023-10-17
Add support for ROCm backend (enabled in CMSSW 13_2_X and later)

VourMa:
As per title. Fully technical, performance cannot be validated (no readily available machine with AMD GPU).

### DIFF
--- a/SDL/Constants.h
+++ b/SDL/Constants.h
@@ -27,7 +27,7 @@ using Dim = alpaka::DimInt<3u>;
 using Dim1d = alpaka::DimInt<1u>;
 using Vec = alpaka::Vec<Dim,Idx>;
 using Vec1d = alpaka::Vec<Dim1d,Idx>;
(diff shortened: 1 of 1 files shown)


GNiendorf:
I don't think this would compile unless this backend has a uint4 implementation. Do you know if it does, @VourMa?

VourMa:
> I don't think this would compile unless this backend has a uint4 implementation. Do you know if it does, @VourMa?

It compiles (and, in fact, it didn't compile in the opposite case), so I think that the implementation exists. In any case, our code compiles just fine, both standalone (expected, as it is not affected), as well as in CMSSW (where the ROCm backend is an option, stopped because of issue #339 for us, which is not a compilation problem).

GNiendorf:
Thank you for the clarification Manos, merging this.

GNiendorf (approved):

PR #340, opened 2023-10-17
adapting ModuleConnectionMap to event setup

slava77:
I'm opening a longer running branch to move away from using singletons.

The first commit is relatively trivial to make the access const. We could as well just merge it as is; there is no change in the interface or behavior.


VourMa (approved):
Tested and works fine. @slava77 should I merge, so that I can build on top of that for the upcoming developments?

PR #341, opened 2023-10-18
Add ability to select backend at run time

ariostas:
Added the ability to select a backend during run time in a way that works both for the standalone version and for the CMSSW setup. This was done by compiling two separate libraries and selecting one at run time. Here are some points describing the reasoning for the changes.

- For CMSSW both libraries need to have different names. We went with `sdl_gpu` and `sdl_cpu`, i.e. the files are `libsdl_gpu.so` and `libsdl_cpu.so`. @YonsiG and @VourMa will add additional changes to this PR and to the CMSSW repo to make it work.
- For the standalone version it is much easier to pick between the two libraries when they have the same name. This is why symlinks are created in `SDL/gpu/libsdl.so` and `SDL/cpu/libsdl.so`. The library is now dynamically linked at runtime, allowing us to override the which one it links to.
- By default, `sdl_make_tracklooper` will now compile both libraries, but to save some time it is possible to exclusively compile the GPU or CPU library with the `-G` and `-C` flags. When both libraries are compiled, the GPU one is used as the default one since its path appears first in `LD_LIBRARY_PATH`.
- `sdl_run` now has the option of only running the code and not recompiling it, simply by omitting the `-f` flag. A backend can be specified with the `-b` flag, e.g. `-b cpu`. This is the easiest way to pick a backend, but when running the `sdl` binary directly ~~it can be done with `LD_PRELOAD=${TRACKLOOPERDIR}/SDL/cpu/libsdl.so sdl <args>`.~~ EDIT: To run the `sdl` binary directly you need to use `LD_LIBRARY_PATH=${TRACKLOOPERDIR}/SDL/cpu/:$LD_LIBRARY_PATH sdl <args>` (see comment below).

I'll do some more testing to make sure that everything looks good, and other people should also test it and review the changes closely because there were significant modifications to scripts and makefiles. I can  update the readme when it's all settled.

### DIFF
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -4,20 +4,31 @@
 #
 
 CCSOURCES=$(filter-out LST.cc, $(wildcard *.cc))
-CCOBJECTS=$(CCSOURCES:.cc=_$(BACKEND).o)
+CCOBJECTS_CPU=$(CCSOURCES:.cc=_cpu.o)
+CCOBJECTS_GPU=$(CCSOURCES:.cc=_gpu.o)
 
 LSTSOURCES=LST.cc
-LSTOBJECTS=$(LSTSOURCES:.cc=_cpu.o)
+LSTOBJECTS_CPU=$(LSTSOURCES:.cc=_cpu.o)
+LSTOBJECTS_GPU=$(LSTSOURCES:.cc=_gpu.o)
 
-LIB=libsdl.so
+LIB_GPU=libsdl_gpu.so
+LIB_CPU=libsdl_cpu.so
+
+ifeq ($(BACKEND), cpu)
+  LIB_GPU=
+else ifeq ($(BACKEND), gpu)
+  LIB_CPU=
+endif
+
+LIBS=$(LIB_GPU) $(LIB_CPU)
 
 #
 # flags to keep track of
 #
 
 CXX                  = g++
-CXXFLAGS             = -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
-CXXFLAGS_CUDA        = -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_70 --use_fast_math --default-stream per-thread -I..
+CXXFLAGS_CPU             = -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
+CXXFLAGS_GPU        = -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_70 --use_fast_math --default-stream per-thread -I..
 ALPAKAINCLUDE        = -I${ALPAKA_ROOT}/include -I/${BOOST_ROOT}/include -std=c++17
 ALPAKASERIAL         = -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
 ALPAKACUDA           = -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_ACC_GPU_CUDA_ONLY --expt-relaxed-constexpr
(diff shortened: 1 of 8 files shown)


YonsiG:
Hi Andres, I am testing the cpu version of sdl_run but meet with some issues. I did 
"sdl_run -s PU200 -n 1 -t testcpu -b cpu", 
but It says 
"/home/users/yagu/TrackLooper_Ntuple/LST_in_cmssw/TrackLooper/bin/sdl_run: line 159: 93187 Aborted                 (core dumped) LD_PRELOAD=/home/users/yagu/TrackLooper_Ntuple/LST_in_cmssw/TrackLooper/SDL/cpu/libsdl.so sdl -i PU200 -o ./testcpu_PU200_NEVT1__LSTNtuple.root -n 1 >> ./testcpu_PU200_NEVT1__LSTRun.log 2>&1
ERROR: sdl command failed!"
in the log file, it says at the end of the program
"double free or corruption (!prev)"
Can you have a look at this? Thanks!

ariostas:
Thank you @YonsiG. I think at some point during testing I forgot to re-run the `setup.sh` script. What was happening was that with `LD_PRELOAD` it was preloading the library, but it was still loading the one it found from `LD_LIBRARY_PATH`. So it was trying to run two things at once. I changed it so that now instead of using `LD_PRELOAD` it temporarily updates `LD_LIBRARY_PATH` to point it to the correct library. So the correct way of running the `sdl` binary directly should be `LD_LIBRARY_PATH=${TRACKLOOPERDIR}/SDL/cpu/:$LD_LIBRARY_PATH sdl <args>`.

YonsiG:
Tried to run the sdl commands by using 
```LD_LIBRARY_PATH=${TRACKLOOPERDIR}/SDL/cpu/:$LD_LIBRARY_PATH ./bin/sdl -i PU200 -v 0 -w 2 -n 1```
There is one printout saying "ana.do_run_cpu: 0". Seems like it's not very correct/consistent

**Comment by ariostas** [comment]
I changed `sdl.cc` so that at runtime it checks which library is loaded. This way, it can correctly print whether it's running on cpu.

**Comment by YonsiG** [comment]
Thanks Andres! This PR gets validated together with PR#15 in SegmentLinking/cmssw , mtv plots are here for out of the box. http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/CMSSW_mtv/compileCPU_GPU/plots_final/plots_ootb/effandfakePtEtaPhi.pdf

**Review by YonsiG — APPROVED** [review]

**Comment by GNiendorf** [comment]
Also, how does this change work with the timing script and switching between backends?

**Comment by ariostas** [comment]
Hi @GNiendorf, the timing script should work fine with this PR. However, this was merged a bit prematurely and we need to add some documentation to the readme to explain the new changes. I'll work on it as soon as possible.

**Comment by YonsiG** [comment]
Additional comparison with the original master branch: 
CPU vs master CPU(same): http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/CMSSW_mtv/compileCPU_GPU/plots_master_compare/plots_cpu_compare/plots_ootb/effandfakePtEtaPhi.pdf
GPU vs master GPU(a bit fluctuations): http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/CMSSW_mtv/compileCPU_GPU/plots_master_compare/plots_gpu_compare/plots_ootb/effandfakePtEtaPhi.pdf

File: SDL/Makefile [hunk 1]
```diff
@@ -4,20 +4,31 @@
...
 
 #
 # flags to keep track of
 #
 
 CXX                  = g++
-CXXFLAGS             = -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
-CXXFLAGS_CUDA        = -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_70 --use_fast_math --default-stream per-thread -I..
+CXXFLAGS_CPU             = -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
+CXXFLAGS_GPU        = -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_70 --use_fast_math --default-stream per-thread -I..
```
VourMa, line 31: Minor comment: Could we fix the indentation in an upcoming PR?

bin/sdl.cc
```diff
@@ -221,14 +220,26 @@ int main(int argc, char** argv)
...
+        while (std::getline(maps, line)) {
+            size_t pos = line.find("/"); // Find the starting position of the path
+            if (pos != std::string::npos) {
+                std::string libraryPath = line.substr(pos, line.find(" ", pos) - pos);
+                if (libraryPath.size() >= 14 && libraryPath.compare(libraryPath.size() - 14, 14, "/libsdl_cpu.so") == 0) {
+                    ana.do_run_cpu = true;
+                    break;
+                }
+            }
+        }
```
VourMa, line 242: This looks to me like a weird way to determine the backend. Could you explain what is the logic behind it, and what kind of complications led to this?

bin/sdl_make_tracklooper
```diff
@@ -57,9 +59,13 @@ if [ -z ${MAKECUTVALUES} ]; then MAKECUTVALUES=false; fi
 if [ -z ${PRIMITIVE} ]; then PRIMITIVE=false; fi
 if [ -z ${T3T3EXTENSION} ]; then T3T3EXTENSION=false; fi
 if [ -z ${DONTUSENN} ]; then DONTUSENN=false; fi
-if [ -z ${CPUBACKEND} ]; then CPUBACKEND=false; fi
+if [ -z ${ONLYGPUBACKEND} ]; then ONLYGPUBACKEND=false; fi
+if [ -z ${ONLYCPUBACKEND} ]; then ONLYCPUBACKEND=false; fi
 if [ -z ${NOPLSDUPCLEAN} ]; then NOPLSDUPCLEAN=false; fi
 
+# If using both -G and -C, -G takes priority
+if [ "${ONLYGPUBACKEND}" == true ]; then ONLYCPUBACKEND=false; fi
```
VourMa, line 67: Could we print also a message if this happens?

Also, it seems like this may be unneeded, given the `elif` order below.

bin/sdl_make_tracklooper
```diff
@@ -158,10 +167,15 @@ else
     (cd SDL && make clean && make ${T3T3EXTENSIONOPT} ${T5CUTOPT} ${BACKENDOPT} ${NOPLSDUPCLEANOPT} -j 32 ${MAKETARGET} && cd -) >> ${LOG} 2>&1
 fi
 
-if [ ! -f SDL/libsdl.so ]; then
-    echo "ERROR: SDL/libsdl.so failed to compile!" | tee -a ${LOG}
-    echo "See ${LOG} file for more detail..." | tee -a ${LOG}
-    exit 1
+if [[ "$BACKENDOPT" != *"gpu"* ]] && [ ! -f SDL/libsdl_cpu.so ]; then
```
VourMa, line 170: Doesn't make more sense to make these equalities rather than inequalities? I imagine that adding a third option would not be possible as the logic is now.

VourMa (commented):
Took a look at the code and the logic seems ok, thanks a lot, @ariostas! Apart from my request for documenting the new interface in the README, I have some very minor comments. Also, I agree with Gavin's point about the backend naming. If it's not easy to fix now, I guess it can become an issue for when we go into the trouble of trying the AMD GPUs.

SDL/Makefile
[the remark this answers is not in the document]
ariostas: Yeah, that's a good point. So in general should we start calling the backend "cuda" instead of "gpu"? Or do you think it would be possible to give `sdl_run` something like `-b gpu` and depending on the hardware decide whether to use CUDA or HIP?

ariostas: I'll do this in the PR I'm preparing.

bin/sdl.cc
ariostas: It is definitely a bit of a weird way to do this. I think this was the only way to do this without adding stuff to the library. But maybe we can define a global constant with `#ifdef`s in the library that stores the backend that it was compiled with.

bin/sdl_make_tracklooper
ariostas: Yeah, I think it's not needed, but I agree that a message saying that that combination of flags is unintended would be good.

bin/sdl_make_tracklooper
ariostas: Yeah, I'll do that.

PR #342, opened 2023-10-18
Fixed overflow when adding pLSs

ariostas:
This PR fixes an issue that @VourMa found out. When loading pLSs into memory there was no check to make sure that the full array fits in the allocated space. What is done now is that if there are too many pLSs then it prints a warning and truncates the array, similar to what was done for track candidates in #324. We also changed the definition of `N_MAX_PIXEL_MD_PER_MODULES` since it should always be twice `N_MAX_PIXEL_SEGMENTS_PER_MODULE`.

No significant part of the algorithm was modified so there are no changes in timing or performance.

### DIFF
--- a/SDL/Constants.h
+++ b/SDL/Constants.h
@@ -115,8 +115,8 @@ ALPAKA_FN_HOST ALPAKA_FN_INLINE WorkDiv createWorkDiv(
 
 const unsigned int MAX_BLOCKS = 80;
 const unsigned int MAX_CONNECTED_MODULES = 40;
-const unsigned int N_MAX_PIXEL_MD_PER_MODULES = 100000;
 const unsigned int N_MAX_PIXEL_SEGMENTS_PER_MODULE = 50000;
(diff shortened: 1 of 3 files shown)


ariostas:
@VourMa found out that there was still an issue since `mdSize` was not properly updated to reflect the truncation. This was fixed in a new commit.

Here are timing comparisons between master and this PR:
Screenshot 2023-10-20 at 9 39 34 AM
PR342

Validation plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR342/PR342_PU200_NEVT-1_78a8ec-PU200/).
Comparison plots to master can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR342/comparison_PR342_8cb74c-PU200_78a8ec-PU200/).

As expected, timing and performance are unchanged.

VourMa (approved):
Thanks, @ariostas! Looks good to me, so I will go ahead and merge.

PR #345, opened 2023-10-20
Cleaned up PR341 and updated readme

ariostas:
I cleaned up some issues brought up after #341 was merged. The backend name of the gpu version was changed to cuda, so that later on it's easier to add HIP. This means that the CMSSW repo will need to be updated to use `libsdl_cuda.so`.

The makefiles were cleaned up a bit, and the new instructions were added to the readme. Someone else should test this to make sure everything is working well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ The above can be even simplified
 
     sdl_run -f -mc -s PU200 -n -1 -t myTag
 
+The `-f` flag can be ommitted when the code has already been compiled. If multiple backends were compiled then the `-b` flag can be used to specify a backend. For example
+
+    sdl_run -b cpu -s PU200 -n -1 -t myTag
+
 ## Command explanations
 
 Compile the code with option flags
@@ -44,6 +48,8 @@ Compile the code with option flags
     sdl_make_tracklooper -mc
     -m: make clean binaries
     -c: run with the cmssw caching allocator
+    -C: only compile CPU backend
+    -G: only compile GPU (CUDA) backend
     -h: show help screen with all options
 
 Run the code
@@ -57,7 +63,13 @@ Run the code
     -w: 0- no writeout; 1- minimum writeout; default: 1
     -o: provide an output root file name (e.g. LSTNtuple.root); default: debug.root
     -l: add lower level object (pT3, pT5, T5, etc.) branches to the output
+
+When running the `sdl` binary directly and multiple backends have been compiled, one can be chosen using the `LD_LIBRARY_PATH` environment variable. For example, one can explicitly use the CPU backend as follows.
+
+    LD_LIBRARY_PATH=$TRACKLOOPERDIR/SDL/cpu/:$LD_LIBRARY_PATH sdl <args>
     
+However, it is important to keep in mind that if that particular backend is was not compiled then it will find another backed without any notice.
+
 Plotting numerators and denominators of performance plots
 
     createPerfNumDenHists -i <input> -o <output> [-g <pdgids> -n <nevents>]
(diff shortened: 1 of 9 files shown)


bin/sdl_make_tracklooper
```diff
@@ -64,7 +64,11 @@ if [ -z ${ONLYCPUBACKEND} ]; then ONLYCPUBACKEND=false; fi
 if [ -z ${NOPLSDUPCLEAN} ]; then NOPLSDUPCLEAN=false; fi
 
 # If using both -G and -C, -G takes priority
-if [ "${ONLYGPUBACKEND}" == true ]; then ONLYCPUBACKEND=false; fi
+if [ "${ONLYGPUBACKEND}" == true ] && [ "${ONLYCPUBACKEND}" == true ]; then
```
VourMa, line 67: `ONLYGPUBACKEND` -> `ONLYCUDABACKEND` ?

VourMa (commented):
Everything seems ok, apart from a single variable that still includes `GPU` instead of `CUDA`. Also, I applied a few README fixes, and made the CMSSW counterpart of this PR (SegmentLinking/cmssw#17).

I invite @GNiendorf to review this PR as well, since he had comments on the previous one.

bin/sdl_make_tracklooper
ariostas: Fixed. Thanks!

ariostas:
I fixed a path that I forgot to update in the Makefile, I addressed the comment by @VourMa, and I adjusted the timing script so that it only compiles the backend that it uses.

I think the last thing that needs to be done, although probably not in this PR, is to modify the `sdl_make_tracklooper` flags to select more general backends. Currently it uses `-C` abd `-G` to select between CPU and CUDA, but if HIP is added then `-G` might become ambiguous. In my opinion, there's no obvious solution since ideally it would be something like `-c`, `-C`, `-H` for CPU, CUDA, and HIP respectively, but `-c` is already used for the caching allocator. The alternative of something like `-b cuda` also doesn't work since `sdl_run` (I think) requires compilation flags to all be single letters. Probably we should just worry about it when the time comes.

VourMa:
> I think the last thing that needs to be done, although probably not in this PR, is to modify the sdl_make_tracklooper flags to select more general backends.

Until we have a machine to compile and run the HIP backend, I don't think it makes sense to go into the trouble of including its compilation in the `sdl_make_tracklooper` but maybe make it an issue, @ariostas?
On my side, I think that this PR is good to go.

GNiendorf:
Can you explain the changes to the README regarding the CMSSW integration before I merge this in? I wasn't at the hackathon so I'm not sure where these changes are coming from.

VourMa:
@GNiendorf Let's do this in the meeting today, so that it's easier/faster, and then merge or act on any specific comments you may have?

README.md
```diff
@@ -174,40 +186,37 @@ For convenience, the workflow has been run for 100 events and the output is stor
...
 Running the configuration file with `cmsRun`, the output file will have a name starting with `DQM`. The name is the same every time this step runs,
 so it is good practice to rename the file, e.g. to `tracking_Iters01LST.root`.
 The MTV plots can be produced with the command:
 ```bash
 makeTrackValidationPlots.py --extended tracking_Iters01LST.root
 ```
+Comparison plots can be made by including multiple ROOT files as arguments.
 
-**Note:** In case one wants to run step2 as well, similar modifications as in step4 (PU files, `--no_exec` flag and input file/number of events) need to be applied.
+**Note:** In case one wants to run step2 as well, similar modifications as in step4 (`--no_exec` flag and input file/number of events) need to be applied.
```
ariostas, line 219: Do we need to include the PU files for step2? Or are they only for step1?

VourMa: We need them. This is my bad, because I though this sentence was referring to step4 (wherePU files are not needed). Let me fix it with a quick commit. Sorry, and thanks for catching this!

PR #346, opened 2023-10-20
WIP: Static function

YonsiG:
move global variables to be allocated inside static function to support compiling LST in CMSSW_13_2 or later. From this CMSSW version, it introduces ROCm backend, and complains about no device found.


YonsiG:
Now the global variables in Event.cc and ModuleConnectionMap.cc are put into classes and functions.

YonsiG:
This PR is very useful if we want to port the global variables into static functions. Especially the trick it uses to deal with the module map variables can be used as templates for more variables. Unfortunately, this effort didn't go through the end. But all the global variables are documented in the LST Object Building Selections document. For further reference and work to implement this, refer to https://docs.google.com/document/d/1MZqSnR-q8YmWN0X4CZIjjV2mYRVW8UxCZnn1NR34ppw/edit?usp=sharing

PR #349, opened 2023-10-27
Added compilation flags for more optimization

### LINKED ISSUE #334: Optimize compilation flags for CPU version
According to the discussion in https://github.com/SegmentLinking/TrackLooper/pull/313#discussion_r1322064922, the optimization of the compilation flags for the CPU version of LST can provide performance gains, and it's worth investigating in a separate PR.

ariostas:
During the hackathon last week I tested a bunch of different compilation flags. This PR adds the best ones that I found, which turned out to be the same ones that were discussed in https://github.com/SegmentLinking/TrackLooper/pull/313#discussion_r1322064922. Closes #334.

## GPU

For the GPU version I simply added `-O3` so that the host code is also optimized. It makes a small, but noticeable difference. Here are the timings before and after.

Screenshot 2023-10-27 at 8 36 10 AM
Screenshot 2023-10-27 at 8 42 54 AM

The performance plots are unchanged. Here are a few comparisons.

image
image
image

The full set of plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR349/comparison_gpu_before_vs_after_d481e6-PU200_9ff0f0-PU200/summary/).

## CPU

For the CPU version I ended up adding `-march=native -mtune=native -Ofast -fno-reciprocal-math -fopenmp-simd`, which is what Slava had suggested (plus `-march=native -mtune=native`, which are fairly standard). Here are the timings before and after.

Screenshot 2023-10-27 at 8 08 30 AM
Screenshot 2023-10-27 at 8 30 06 AM

You can see that there are improvements in all steps, some much larger than others. One thing that is very interesting is that the non-cached allocator is somehow faster than the cached one.

Here are some performance comparisons.

image
image
image

The full set of plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR349/comparison_cpu_before_vs_after_d481e6-PU200_9ff0f0-PU200/summary/).

### DIFF
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -27,8 +27,8 @@ LIBS=$(LIB_CUDA) $(LIB_CPU)
 #
 
 CXX                  = g++
-CXXFLAGS_CPU         = -O2 -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
-CXXFLAGS_CUDA        = -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared -arch=compute_70 --use_fast_math --default-stream per-thread -I..
+CXXFLAGS_CPU         = -march=native -mtune=native -Ofast -fno-reciprocal-math -fopenmp-simd -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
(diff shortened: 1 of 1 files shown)


slava77:
The change in the CPU pT5 time is surprising to me.
It would be nice to check it a bit in more depth.

Which CPU is this?
Perhaps running `igprof -pp` is the easiest.
on Intel the IntelAdvisor may be more informative.

slava77 (approved):

PR #350, opened 2023-10-27
Separate duplicate flags for first and second passes of checkHitspLS

ariostas:
I made a small change to the `checkHitspLS` kernel. Now the duplicate flag is a `char` instead of a `bool`. The first pass uses the lowest bit to mark it as a duplicate, and the second pass uses the second lowest bit. In the second pass, only the ones marked as duplicate in the first pass are skipped. This way, we finally get perfect agreement between CPU and GPU, and additionally we slightly lower the duplicate rate, while maintaining the same efficiency.

Here is how duplicate rates looked before between CPU and GPU.
image
And here is how they look now.
image

Here is a comparison of duplicate rate on GPU before and after the change.
image

And here are efficiency and fake rate comparisons on GPU.
image
image
image

The full set of plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR350/).

Here are the timings on GPU before and after. They seem to be the same within run-to-run variations.
Screenshot 2023-10-26 at 3 22 10 PM
Screenshot 2023-10-26 at 4 11 54 PM

### DIFF
--- a/SDL/Kernels.h
+++ b/SDL/Kernels.h
@@ -27,9 +27,9 @@ namespace SDL
         pixelQuintupletsInGPU.isDup[pixelQuintupletIndex] = 1;
     };
 
-    ALPAKA_FN_ACC ALPAKA_FN_INLINE void rmPixelSegmentFromMemory(struct SDL::segments& segmentsInGPU, unsigned int pixelSegmentArrayIndex)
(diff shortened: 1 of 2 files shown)


VourMa:
> They seem to be the same within run-to-run variations.

I wouldn't say so. I think that the timing is consistently lower, which makes sense in view of the duplicate rate reduction.
It could be useful to check whether the CPU timing stays the same - if I would have to take a guess, I would say so, since the duplicate rate didn't change for that.

VourMa (approved):
The PR and its validation/timing look good. I am approving and merging this, as my previous comment is an optional test.

PR #355, opened 2023-11-14
updates to get module maps used as ES products jointly with CMSSW 

slava77:
coupled with updates in https://github.com/SegmentLinking/cmssw/pull/16

- split somewhat heavy methods related to filling module data from Module.h to   ModuleMethods
- move `LST::loadMaps` to a free function to avoid making an instance of LST where it's not needed
- add `LST::loadAndFillES` method to be called by an CMSSW ES producer
- const pointers in modules SoA data

### DIFF
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -170,7 +170,7 @@ void SDL::initModules(const char* moduleMetaDataFilePath) {
 
   // nModules gets filled here
   loadModulesFromFile(
-      modulesInGPU, modulesBuffers, nModules, nLowerModules, *pixelMapping, queue, moduleMetaDataFilePath);
+      modulesBuffers, nModules, nLowerModules, *pixelMapping, queue, moduleMetaDataFilePath);
 }
 
 // Temporary solution to the global variables. Should be freed with shared_ptr.
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -1,5 +1,6 @@
 #include "Event.h"
 
+using Acc = SDL::Acc;
 SDL::modules* SDL::modulesInGPU = new SDL::modules();
 SDL::modulesBuffer<Acc>* SDL::modulesBuffers = new SDL::modulesBuffer<Acc>(devAcc);
 std::shared_ptr<SDL::pixelMap> SDL::pixelMapping = std::make_shared<pixelMap>();
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -172,7 +172,6 @@ void SDL::initModules(const char* moduleMetaDataFilePath) {
   // Set the relevant data pointers.
   modulesBuffersES = modulesBuffers;
   modulesInGPU->setData(*modulesBuffersES);
-
 }
 
 // Temporary solution to the global variables. Should be freed with shared_ptr.
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -1,8 +1,8 @@
 #include "Event.h"
 
-using Acc = SDL::Acc;
 SDL::modules* SDL::modulesInGPU = new SDL::modules();
-SDL::modulesBuffer<Acc>* SDL::modulesBuffers = new SDL::modulesBuffer<Acc>(devAcc);
+SDL::modulesBuffer<SDL::Dev>* SDL::modulesBuffers = new SDL::modulesBuffer<SDL::Dev>(devAcc);
+SDL::modulesBuffer<SDL::Dev> const* SDL::modulesBuffersES = nullptr;
 std::shared_ptr<SDL::pixelMap> SDL::pixelMapping = std::make_shared<pixelMap>();
 uint16_t SDL::nModules;
 uint16_t SDL::nLowerModules;
@@ -166,11 +166,13 @@ void SDL::Event::resetEvent() {
 void SDL::initModules(const char* moduleMetaDataFilePath) {
   QueueAcc queue(devAcc);
 
-  // Set the relevant data pointers.
-  modulesInGPU->setData(*modulesBuffers);
-
   // nModules gets filled here
   loadModulesFromFile(modulesBuffers, nModules, nLowerModules, *pixelMapping, queue, moduleMetaDataFilePath);
+
+  // Set the relevant data pointers.
+  modulesBuffersES = modulesBuffers;
+  modulesInGPU->setData(*modulesBuffersES);
+
 }
 
 // Temporary solution to the global variables. Should be freed with shared_ptr.
@@ -591,7 +593,7 @@ void SDL::Event::createTriplets() {
 
   // Allocate and copy module_nConnectedModules from device to host
   auto module_nConnectedModules_buf = allocBufWrapper<uint16_t>(devHost, nLowerModules, queue);
-  alpaka::memcpy(queue, module_nConnectedModules_buf, modulesBuffers->nConnectedModules_buf, nLowerModules);
+  alpaka::memcpy(queue, module_nConnectedModules_buf, modulesBuffersES->nConnectedModules_buf, nLowerModules);
   alpaka::wait(queue);
 
   uint16_t* module_nConnectedModules = alpaka::getPtrNative(module_nConnectedModules_buf);
@@ -1181,10 +1183,10 @@ void SDL::Event::addMiniDoubletsToEventExplicit() {
   alpaka::memcpy(queue, nMDsCPU_buf, miniDoubletsBuffers->nMDs_buf, nLowerModules);
 
   auto module_subdets_buf = allocBufWrapper<short>(devHost, nLowerModules, queue);
-  alpaka::memcpy(queue, module_subdets_buf, modulesBuffers->subdets_buf, nLowerModules);
+  alpaka::memcpy(queue, module_subdets_buf, modulesBuffersES->subdets_buf, nLowerModules);
 
   auto module_layers_buf = allocBufWrapper<short>(devHost, nLowerModules, queue);
-  alpaka::memcpy(queue, module_layers_buf, modulesBuffers->layers_buf, nLowerModules);
+  alpaka::memcpy(queue, module_layers_buf, modulesBuffersES->layers_buf, nLowerModules);
 
   auto module_hitRanges_buf = allocBufWrapper<int>(devHost, nLowerModules * 2, queue);
   alpaka::memcpy(queue, module_hitRanges_buf, hitsBuffers->hitRanges_buf, nLowerModules * 2);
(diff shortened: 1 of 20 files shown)


slava77:
now rebased

SDL/LST.cc
```diff
@@ -1,34 +1,39 @@
...
     static std::once_flag mapsLoaded;
-    std::call_once(mapsLoaded, &SDL::LST::loadMaps, this);
+    std::call_once(mapsLoaded, &SDL::LST::loadMaps);
+
     TString path = get_absolute_path_after_check_file_exists(
-        TString::Format("%s/data/centroid_CMSSW_12_2_0_pre2.txt",TrackLooperDir_.Data()).Data());
+        TString::Format("%s/data/centroid_CMSSW_12_2_0_pre2.txt",trackLooperDir().Data()).Data());
     static std::once_flag modulesInited;
     std::call_once(modulesInited, SDL::initModules, path);
 }
```
VourMa, line 17: Should this be deleted, now that the maps are loaded in the ESProducer? Or has it stayed until we know that the ESRpoducer method works and will be deleted in the future?

VourMa (commented):
Not sure how ready this was but I took a look and have just a small question here. More questions on the CMSSW side...

slava77:
@ariostas 
please check the linter error report.

`error: invalid argument '-std=c++17' not allowed with 'C'`  looks like the reason for `error: cannot use 'try' with exceptions disabled`

I was expecting a proposed patch file with changes (what's usually done with cmssw setup).
Instead the bad formatting comes out with comments like
```
File SDL/Constants.h does not conform to Custom style guidelines. (lines 14, 66)
```
which are not very useful/practical.
I'm not sure though if the `'-std=c++17' not allowed` have prevented something to complete.

slava77:
> `error: invalid argument '-std=c++17' not allowed with 'C'` looks like the reason for `error: cannot use 'try' with exceptions disabled`

I got rid of it by adding `--language=c++`

ariostas:
Hi @slava77, the `error: invalid argument '-std=c++17'` error is not preventing anything from running. It's an annoying thing that I explain more in a comment on your PR https://github.com/SegmentLinking/TrackLooper/pull/360.

Regarding the patch file, I thought that it would be the same amount of work to apply the patch file vs to just run `clang-format` manually, so I didn't think it was worth it. But I can implement generating a patch file, if you prefer.

slava77:
> Regarding the patch file, I thought that it would be the same amount of work to apply the patch file vs to just run `clang-format` manually, so I didn't think it was worth it. But I can implement generating a patch file, if you prefer.

we should at least have the commands needed to run clang-format and clang-tidy in the readme.
While clang-format seems trivial (it only needs the file name), the clang-tidy is not trivial

slava77:
I don't see any activity in the CI frame after my /run commands.
@ariostas are they blocked by the failed linter or does it take longer to react?

ariostas:
@slava77 oh sorry, that's a bug I have to fix. What happened was that I set it so that only one job runs at a time, to prevent us from accidentally running multiple iterations of the same thing. Try writing both commands on a single comment (in separate lines), and then let's check if they run.

slava77:
The /run action actually came back to me in https://github.com/SegmentLinking/TrackLooper/actions/runs/7399973283/job/20132579034; although I still do not see it in this PR thread.

It came back with what's probably from 
```
/run cmssw from-CMSSW_13_3_0_pre3_LST_X/modules-dev
Branch name is invalid. Ignoring...
```

not sure what I was supposed to pass instead to pick up https://github.com/SegmentLinking/cmssw/pull/16

the errors from the CI are pretty hard to decode

slava77:
> The PR was built and ran successfully with CMSSW. Here are some plots.

@ariostas 
did you fix something or did I misunderstand the previous CI reports that said the jobs were canceled or had problems?

Are the plots meant to have no reference? It would be much better to have a comparison

slava77:
cmssw on 100 events look OK now
eff/fr
image

hits
image

I think that the tiny differences are from reproducibility on GPU.

slava77:
@ariostas 
is there a way to rerun the linter? I tried to rerun from the last attempt interface, but it used the old commit/variant before your latest updates.

ariostas:
@slava77 the linter workflow is triggered differently and github uses the workflow file present in the current branch. So you would have to rebase, but it's probably not worth it.

ariostas:
I think that the cmssw test will fail because I haven't updated the recipe to use lst_headers.xml. Now that I think about it I could have updated it without breaking the action for the current default cmssw branch. It's an easy fix so I'll do it tonight.

ariostas:
/run standalone
/run cmssw from-CMSSW_13_3_0_pre3_LST_X/modules-dev

Should all work now. However, one thing to keep in mind is that I set the CI so that if you specify a custom branch of CMSSW then it doesn't produce comparison plots since presumably the changes in both repos depend on each other or the CMSSW setup is different. This could be solved by having a fixed reference file somewhere, as we discussed in the meeting.

ariostas:
> In a context of the CMSSW branch being an unmerged PR there is still a well defined "master" reference that would come from running the cmssw test out of the box

Even in that context, if let's say there's a change to the xml files one might need to scram uninstall, reinstall and rebuild. I'm expecting that most likely something will break, but I'll think if there's a way to do it without resorting to wiping the cmssw setup and starting from scratch. Do you think this kind of changes might not be an issue?

There's also the (very infrequent) case of changing to a later cmssw release, which would definitely require rebuilding everything. But we can probably not worry about those for now.

slava77:
> Even in that context, if let's say there's a change to the xml files one might need to scram uninstall, reinstall and rebuild.

OK, this PR pair lst:355,cmssw:16 is special due to the creation of an extra header package

> There's also the (very infrequent) case of changing to a later cmssw release, which would definitely require rebuilding everything. But we can probably not worry about those for now.

assuming the CI is established (we are almost there), I'd expect that the "latest" reference is generally available. So, we do not need to rerun the baseline every time.

ariostas:
> I'd expect that the "latest" reference is generally available

Yeah, that's true. I'll think of a good way to do this.

For lst:355,cmssw:16, did anything else about the CMSSW setup change? The CI failed with this error and I'm not sure why, so I'll look into it.
```
Exception Message:
No data of type "SDL::modulesBuffer<alpaka_common::DevHost>" with label "" in record "TrackerRecoGeometryRecord"
 Please add an ESSource or ESProducer to your job which can deliver this data.
----- End Fatal Exception -------------------------------------------------
```

slava77:
> No data of type "SDL::modulesBuffer<alpaka_common::DevHost>"

that's my mistake, I did not update the lstProducer/tracking related file.
My tests were on a manually modified config.
I will fix it soon.

slava77:
> There was a problem while building and running with CMSSW. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/7590589660).

this is 
`ERROR: Could not find the file = "/data/endcap_orientation_data_CMSSW_12_2_0_pre2.txt"`

it looks like the `LST_BASE` is not set.

The [log ](https://github.com/SegmentLinking/TrackLooper/actions/runs/7590589660/job/20677355757#step:8:916) has a message
```
**** Following environment variables are going to be unset.
      LST_BASE
```
It sounds like something is off in the CI setup.

@ariostas do you think it's on the CI side or in my code updates?

ariostas:
It seems like there's a bug where if there's a new line character in the line that contains the branch name it ends up messing up the branch name check. I'll try to run it again.

/run cmssw from-CMSSW_13_3_0_pre3_LST_X/modules-dev

ariostas:
@slava77 seems like the issue now is that it doesn't have the required files in `RecoTracker/LST/data` on the CMSSW side.

ariostas:
Still have to fix the bug where a new line or some weird character messes it up.

/run cmssw from-CMSSW_13_3_0_pre3_LST_X/modules-dev

slava77:
> Still have to fix the bug where a new line or some weird character messes it up.

oh, I thought it was fixed already
are you restarting it manually or is the current solution to add some random text with a newline before the `/run` command?

ariostas:
For now the solution is to put the branch name at the very end of the comment without any spaces or new lines at the end. I wonder if for some reason it's inserting a \r or something like that that it doesn't like. I'll do some testing and submit a PR tomorrow.

SDL/Event.cc
```diff
@@ -1,7 +1,8 @@
 #include "Event.h"
 
 SDL::modules* SDL::modulesInGPU = new SDL::modules();
-SDL::modulesBuffer<Acc>* SDL::modulesBuffers = new SDL::modulesBuffer<Acc>(devAcc);
+SDL::modulesBuffer<SDL::Dev>* SDL::modulesBuffers = new SDL::modulesBuffer<SDL::Dev>(devAcc);
```
GNiendorf, line 4: I see you changed Acc to SDL::Dev here, but you should do this for all of our objects to be consistent I think. I don't know if this matters since the performance plots look unchanged and this uses SDL::Dev while all the other Buffers still have Acc, but good to be consistent.

Same in all of the header files that define a buffer, to remove confusion we should move the "Acc" named templated parameter to Dev so it's clear what we pass in. See the random examples below.

https://github.com/SegmentLinking/TrackLooper/blob/25e08e79a0def04b7b51e8ec84275e8e84fd8829/SDL/Event.cc#L319

https://github.com/SegmentLinking/TrackLooper/blob/25e08e79a0def04b7b51e8ec84275e8e84fd8829/SDL/MiniDoublet.h#L96-L99

slava77: > I don't know if this matters since the performance plots look unchanged and this uses SDL::Dev while all the other Buffers still have Acc, but good to be consistent.

right, Acc or Dev does not seem to matter for the buffers at the final implementation level (after all the template parsing is done); however code exposed to CMSSW needs to be templated from a Dev. 

Dev seems to be more appropriate based on Alpaka docs for a buffer https://alpaka.readthedocs.io/en/latest/_images/structure_assoc.png

SDL/Kernels.h
```diff
@@ -18,12 +18,12 @@ namespace SDL {
 
   ALPAKA_FN_ACC ALPAKA_FN_INLINE void rmPixelTripletFromMemory(struct SDL::pixelTriplets& pixelTripletsInGPU,
                                                                unsigned int pixelTripletIndex) {
-    pixelTripletsInGPU.isDup[pixelTripletIndex] = 1;
+    pixelTripletsInGPU.isDup[pixelTripletIndex] = true;
```
VourMa, line 21: Out of curiosity, why this change? Did it cause an issue? Is this defined as a bool variable?

SDL/LST.cc
```diff
@@ -1,58 +1,62 @@
...
-}
 
-TString SDL::LST::get_absolute_path_after_check_file_exists(const std::string name) {
-  std::filesystem::path fullpath = std::filesystem::absolute(name.c_str());
-  if (not std::filesystem::exists(fullpath)) {
-    std::cout << "ERROR: Could not find the file = " << fullpath << std::endl;
-    exit(2);
+  void loadMaps() {
+    // Module orientation information (DrDz or phi angles)
+    auto const& tldir = trackLooperDir();
```
VourMa, line 17: What's the point of defining this, if `trackLooperDir()` is used almost everywhere in the following? Was it meant to be a short version but then it was forgotten?

VourMa (commented):
I left a couple of comments, mostly for my own understanding, rather than real actions. I understand that there are two open issues with this PR:

1) It has a conflict.
2) We should replace `Acc` -> `SDL::Dev`

I will approve and merge when 1) is resolved, while 2) will be followed up in a new PR (do we need an issue or is it going to be followed up soon?).

SDL/Kernels.h
slava77: this was from a clang check; a part missed in #356

SDL/LST.cc
slava77: removed now

slava77:
> We should replace `Acc` -> `SDL::Dev`

done now in [eca2c59](https://github.com/SegmentLinking/TrackLooper/pull/355/commits/eca2c5998917f6952343e00f6b59a0ed80ccc3c7)

VourMa (approved):
This is out of @slava77's way

Issue #290, opened 2023-05-17
Fix code style


VourMa:
The indentation, and the style in general, is all over the place, all over the code. This creates issues understanding the code and even developing it every now and then. 

It would be useful to run the whole code through clang-format at some point and implement a CI to keep newly changed files tidy. A code snippet to do that was provided by @pwittich [here](https://github.com/apollo-lhc/cm_mcu/blob/cd555aaac104af2ab433197d4a3231bb58e39ade/.github/workflows/c-cpp.yml#L100-L126).
The CI can be set up independently and already start working on the files that we change from now on.

GNiendorf:
@VourMa Is this still an issue? Figure we should close this.

VourMa:
Indeed, this can be closed given all the work on the CI by Andres. Closing now.

PR #356, opened 2023-11-29
Format code in preparation for CI and CMSSW integration

ariostas:
This PR simply adds the `.clang-format` and `.clang-tidy` files that are used in CMSSW, and makes the changes resulting from running `clang-format` on the SDL directory.

I limited it to only formatting the SDL directory, since that's what will go into CMSSW, and the rest of the things will probably be moved to a `TrackLooper-external` repo at some point.

After this change, we can now make the CI enforce consistent formatting, and keep working towards being compliant with the CMSSW requirements. This commit, along with the upcoming CI, will address #290.

I skimmed through the changes and they seem reasonable. A lot of it is just changing the indentation from using 4 to 2 spaces.

The changes made in this PR are purely formatting, but here are some timings and comparison plots to make sure that the changes don't affect the way the compiler interprets things.

Before and after timings:
Screenshot 2023-11-29 at 1 37 31 PM
Screenshot 2023-11-29 at 1 22 20 PM

Here are some comparison plots:
image
image
image

The differences look like the typical run-to-run variations we've been seeing. The full set of comparison plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR356/comparison_0caabb-PU200_428fdf-PU200/summary/).


slava77:
I scrolled through a few file diffs and the patterns look relatively familiar to what I've seen in CMSSW and didn't notice any significant bad cases that may suggest adding `nolint` kind of protection.

ariostas:
> is clang-tidy part going to be in another step?

The things that clang-tidy can automatically fix are pretty limited, so we'll have to go through the warnings manually. I can do this at some point.

There's only one error on the clang-tidy side, and the rest at just warnings, so it's probably fine. Maybe I can just fix the error in this PR and leave looking through the warnings for later?

VourMa (approved):

PR #357, opened 2023-12-18
Add CI workflows

ariostas:
This PR implements CI workflows to perform static checks with `clang-format` and `clang-tidy`, as well as to perform test runs of the code in standalone and cmssw setups.

The static checks are performed automatically for all commits to PRs. The CI doesn't fix anything, but it gives warnings for formatting and other potential issues. I couldn't get `clang-tidy` to work perfectly, but it works well enough for now. My plan is that when in a few months the CMSSW integration has progressed enough, I'll update the workflow to use `scram b code-format` and `scram b code-checks` to more accurately mirror what is done by the [cmsbuild](https://github.com/cmsbuild) bot.

The testing workflows are triggered by leaving comments on PRs containing `/run standalone` and/or `/run cmssw`. The latter command accepts an optional argument to specify the branch of the CMSSW fork that should be used.

I have documented everything more extensively in the `.github/workflows/README.md` file and in the [SegmentLinking/TrackLooper-actions](https://github.com/SegmentLinking/TrackLooper-actions) repository.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,22 @@
+# CI workflows
+
+This directory contains workflows that are run for Continuous Integration (CI) testing.
+
+## Static code checks
+
+The workflow that runs static code checks is based on the clang-format and clang-tidy requirements from CMSSW. They are run automatically for commits in pull-requests, and should be roughly equivalent to running `scram build code-checks` and `scram build code-format`. They are fairly lenient, checking only the lines that were changed without making any modifications. The aim is to make sure that the code is compliant for the CMSSW integration. More information can be found [here](https://cms-sw.github.io/PRWorkflow.html). As work towards the full integration with CMSSW progresses, we will tune these checks to become more stringent and closer to what is done during the CMSSW CI checks.
+
+## Testing worflows
+
+The workflows that run the code and produce validation and comparison plots for the standalone and CMSSW setups depend on custom GitHub actions that are located in [SegmentLinking/TrackLooper-actions](https://github.com/SegmentLinking/TrackLooper-actions). Most of the workflow is offloaded to these actions, so it is easy to adjust the process without modifying this repository. These workflows are much more time-consuming, so they must be manually triggered by leaving a comment on the PR containing `/run standalone` and/or `/run cmssw`.
+
+When testing the CMSSW integration, a PR in this repository might depend on a corresponding PR in the [CMSSW fork](https://github.com/SegmentLinking/cmssw). Since the two PRs need to be tested together, using `/run cmssw` would not work. For this reason, there is an optional parameter to specify a branch of the CMSSW repository, `/run cmssw some-branch`. This command needs to be in its own line in the comment or it will not work correctly.
+
+## First-time configuration
+
+There is a one-time configuration step that needs to be done in order to allow the CI to upload the plots to the archival repository. We leave the instructions here in case it ever needs to be done again.
+
+1. Generate a new SSH key with `ssh-keygen -t ed25519`. Save it as `tmp_deploy_key`, and leave it without a password.
+1. Create a new repository secret named `DEPLOY_PRIVATE_KEY` and set it to the contents of `tmp_deploy_key`.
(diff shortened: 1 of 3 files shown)


VourMa:
> /run standalone

I was hoping that this would work for this PR, as the CI worked but I guess I was mistaken. As a result, there is not easy way for me to check how this works.

VourMa:
In [CI results of this PR](https://github.com/SegmentLinking/TrackLooper/actions/runs/7254163937/job/19762311106?pr=357) I see multiple mentions of `/cvmfs/cms.cern.ch/common/scram: 14: COUNT++: not found`. Is this something to be fixed?

VourMa:
In (SegmentLinking/TrackLooper-actions/cmssw/action.yml](https://github.com/SegmentLinking/TrackLooper-actions/blob/main/cmssw/action.yml), I see a mention of your username, @ariostas. Is this going to work for others? Does it make a real difference?

VourMa:
All in all, does it make sense to just merge this PR to be able to test it in action? I couldn't find any smoking gun issues, apart from the minor comments above, and I don't think that anything would break in any case.

ariostas:
@VourMa thanks for your comments.

Unfortunately, as you said, we can't self-test this PR since this action runs only when it's in the main branch.

The errors you're seeing are expected. I'm running the setup script to set some environment variables, but since the linux version doesn't match what cmssw expects there are some issues, but it doesn't affect anything that we need.

As for my username, it appears because I'm storing the docker images in my personal docker hub account. They have restricted a lot when it comes to free organization accounts, but if you prefer we can create some personal account for LST and store it there. I added instructions in the docker folder of the actions repo for how to do this if we need to move the images somewhere else later on.

If you want to try things you can look at the repo in the SegmentLinkinTests organization. But I think it makes sense to just merge it. If something is broken most likely we'll be able to fix it from the other repo. But I tested things, so everything should be good to go.

VourMa:
Thanks for the reply!

> As for my username, it appears because I'm storing the docker images in my personal docker hub account. They have restricted a lot when it comes to free organization accounts, but if you prefer we can create some personal account for LST and store it there. I added instructions in the docker folder of the actions repo for how to do this if we need to move the images somewhere else later on.

I think this is ok as long as it works for others. The instructions are useful for the long-term, thank you!

> If you want to try things you can look at the repo in the SegmentLinkinTests organization. But I think it makes sense to just merge it. If something is broken most likely we'll be able to fix it from the other repo. But I tested things, so everything should be good to go.

If I merge, would I be able to run the `/run standalone` command in my PR, so that we test like that? Or would I have to rebase my source branch of that PR or something?

ariostas:
@VourMa you can run it with your PR after merging. You don't need to rebase. However, the formatting won't be checked unless you push something new. Maybe let's make sure that the forgetting is fine by manually running clang-format.

VourMa (approved):
From a first look, the code looks good. I am approving and merging this, so that we can test it in action. In case we find a cornercase where things don't work as expected, we will fix it in a follow-up PR.

VourMa:
> However, the formatting won't be checked unless you push something new. Maybe let's make sure that the forgetting is fine by manually running clang-format.

Should I just push a "dummy" commit to PR #358 to trigger the tests? Or should I run something manually on my terminal?

.github/workflows/README.md
```diff
@@ -0,0 +1,22 @@
+# CI workflows
+
+This directory contains workflows that are run for Continuous Integration (CI) testing.
+
+## Static code checks
+
+The workflow that runs static code checks is based on the clang-format and clang-tidy requirements from CMSSW. They are run automatically for commits in pull-requests, and should be roughly equivalent to running `scram build code-checks` and `scram build code-format`. They are fairly lenient, checking only the lines that were changed without making any modifications. The aim is to make sure that the code is compliant for the CMSSW integration. More information can be found [here](https://cms-sw.github.io/PRWorkflow.html). As work towards the full integration with CMSSW progresses, we will tune these checks to become more stringent and closer to what is done during the CMSSW CI checks.
```
slava77, line 7: is there a way to run the code checks locally without the CI?

PR #360, opened 2023-12-30
specify C++ language in clang-tidy static-checks.yml

slava77:
it looks like clang-tidy/clang does not recognize quite a few of our files as C++; I propose to add `--language=c++` next to  `-std=c++17` to resolve this

I tested it locally and the `error: invalid argument '-std=c++17' not allowed with 'C'` appearing while running is gone

### DIFF
--- a/.github/workflows/static-checks.yml
+++ b/.github/workflows/static-checks.yml
@@ -43,7 +43,7 @@ jobs:
           tidy-checks: ''
           lines-changed-only: true
           ignore: .github|bin|code|data|efficiency
-          extra-args: '-ferror-limit=0 -std=c++17 -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED ${{ env.include-flags }}'
+          extra-args: '-ferror-limit=0 --language=c++ -std=c++17 -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED ${{ env.include-flags }}'
       - name: Fail job if there were checks that failed
         if: steps.linter.outputs.checks-failed > 0
         run: exit 1


slava77:
I'm curious why are we using clang-format-12 and clang-tidy-12 in the linter instead of what's coming with llvm setup in cmssw on cvmfs (16.0.3 in 13_3 and more recently 17.0.3)?

ariostas:
Hi @slava77, this was part of the reason why I said that the `clang-tidy` part of the CI is good enough, but definitely not perfect. `clang-tidy` assumes that `.h` files (so most of our files) are in C. The thing is that, as far as I can tell, there are only two popular and actively maintained GitHub Actions that run `clang-tidy`: [`cpp-linter-action`](https://github.com/cpp-linter/cpp-linter-action) (the one we use) and [`clang-tidy-review`](https://github.com/ZedThree/clang-tidy-review) (which I had more issues with). Both of them don't seem to expose a low enough level of control of `clang-tidy` to fix this sort of things. You can see in this test PR https://github.com/SegmentLinkingTests/TrackLooper/pull/3 that even with that additional flag you still get the same behavior. The purpose of the `-std=c++17` flag is to fix some other errors you would get in  `.cpp` files. So I think that the only options would be to either implement a `clang-tidy` action on our own with the right parameters, try forking one of the existing ones and fix it, or just deal with some fictitious warnings for now until we can test with `scram b code-checks`. I opted for the last option, but if you prefer we can try the second option.

Regarding version 12 vs 16, I think I accidentally checked the version in CMSSW 13_0_0 instead of 13_3_0, but we can fix that easily.

slava77:
> You can see in this test PR [SegmentLinkingTests#3](https://github.com/SegmentLinkingTests/TrackLooper/pull/3) that even with that additional flag you still get the same behavior.

Uhm, the differences in my local setup were
-  I listed the arguments combined after ` -- ` instead of `--extra-arg`
- clang-tidy was from the cmssw distribution

I'm curious if one of these is the reason for the failure in SegmentLinkingTests#3

slava77:
> Uhm, the differences in my local setup were

and also I passed the full file, not just the changed lines

ariostas:
> I listed the arguments combined after -- instead of --extra-arg

Yeah, that's the issue. I don't know how to pass arguments like that to the action.

ariostas:
> and also I passed the full file, not just the changed lines

The action uses the full file, but only reports warnings/errors in lines that were changed.

ariostas:
I finally figured out how to fix this issue. I'll close this PR and tomorrow I'll submit a new PR with all the CI fixes.

PR #361, opened 2024-01-05
Fix a few issues with the CI

ariostas:
I fixed a few issues that came up after the CI was implemented. Here they are:

- `clang-tidy` defaults to interpreting `.h` files as C code. The action used to run this check does not allow enough control over the command that is run to specify the right parameters. This issue is also not solved by a standard `compile_commands.json` compilation database since it typically only contains commands for `.cc` files. This was solved by using the `compdb` python package to also generate compilation commands for headers.
- The version of `clang-format` and `clang-tidy` was 12, but now it was set to 16 to match the version in CMSSW 13_3_0.
- The CI was implemented in a way that prevents equivalent jobs to run at the same time. However, there was a bug where new comments were cancelling previous jobs because the concurrency grouping was not precise enough.
- When running the CMSSW checks, there is the option to specify a branch. There is a check of the branch name to prevent code injection, but the regex that was used was not generic enough. The check is now done with a `git` command, so it should be completely accurate.
- It used to not be very visible when things were running in the CI. Now the CI adds associated checks to the commit that is being used.
- There are now phony targets in the makefile to fix formatting and check for issues. Instructions were added to the readme.

There was also the issue of the linter not being able to post comments. I'm not sure why that is happening, but I found online that turning actions off and on again might solve this issue. I did this, so hopefully it will work now.

I will also update the CMSSW workflow so that it produces comparison plots instead of just validation plots. But that is all done in [TrackLooper-actions](https://github.com/SegmentLinking/TrackLooper-actions) so no changes need to be made here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,6 @@ bin/sdl: bin/sdl.o $(OBJECTS)
 %.o: %.cc
 	$(CXX) $(PTCUTFLAG) $(T3T3EXTENSION) $(CFLAGS) $(EXTRACFLAGS) $(CUTVALUEFLAG) $(PRIMITIVEFLAG) $(DOQUINTUPLET) $(ALPAKAINCLUDE) $(ALPAKASERIAL) $< -c -o $@
 
-
-.PHONY: $(ROOUTIL) efficiency clean format check check-fix
-
 $(ROOUTIL):
 	$(MAKE) -C code/rooutil/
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -67,4 +70,11 @@ clean:
 	rm -f SDL/*.o
 	cd efficiency/ && make clean
 
-.PHONY: $(ROOUTIL) efficiency
+format:
+	clang-format -i SDL/*.cc SDL/*.h
+
+check:
+	clang-tidy SDL/*.cc SDL/*.h -- --language=c++ -std=c++17 -I. -I$(ROOT_ROOT)/include -I$(CMSSW_BASE)/src -I${ALPAKA_ROOT}/include -I/${BOOST_ROOT}/include -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -I/cvmfs/cms.cern.ch/el8_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/include/c++/11.4.1 -I/cvmfs/cms.cern.ch/el8_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/include/c++/11.4.1/x86_64-redhat-linux-gnu/
```
slava77, line 77: To avoid copy-paste I'd use some common var to be used both in `check` and `check-fix`.
Is it possible to have this rule in `SDL/Makefile` instead?

for the gcc part I used a probably more maintainable  
```
`g++ -E -x c++ - -v < /dev/null 2>&1 | sed -e 's/^ //g' | awk '/starts here:/{p="";s=1}/End of search/{ms=p;s=1}{if(s==0)p=p" -I"$0;s=0}END{print ms}'`
```

Also, our default build includes
`-DT4FromT3 -DCACHE_ALLOC -DDUP_pLS -DDUP_T5 -DDUP_pT5 -DDUP_pT3 -DCrossclean_T5 -DCrossclean_pT3 -DPT_CUT=0.8`
at least some of these change the code visible to the compiler.

ariostas: Yeah, those are very good points. I'll take care of those.

ariostas:
Following @slava77's suggestions I moved the format and check targets to the makefile in the SDL directory, which seems more appropriate. I also included more flags to clang-tidy to match the ones used during compilation. I had to make a modification to the suggested command to collect the include paths since including one of those paths caused a bunch of errors. There is an equivalent path, but with a "fixed" suffix, which works well.

I considered producing a `compile_commands.json` database and use the same python package that is used in the CI, but overall it makes things more complex and we get the same warnings/errors, so I didn't think it was worth it.

slava77 (approved):

slava77:
> I'm currently testing producing comparison plots in the CMSSW workflow

just to be sure, before pushing the "merge"; is that test passing now?

ariostas:
I just pushed the last changes to the TrackLooper-actions repo, so it's all ready now. Everything is working in the test repo, so it should work here too.

PR #365, opened 2024-01-25
test first pass pLS cleaning with 2 hits

slava77:
3 hit requirement seems too restrictive; at least as a reference I wanted to check if performance changes for the better or not by reducing the check down to 2 hits

### DIFF
--- a/SDL/Kernels.h
+++ b/SDL/Kernels.h
@@ -486,7 +486,7 @@ namespace SDL {
                 break;
             }
           }
-          if (npMatched >= 3) {
+          if (npMatched >= 2) {
             rmPixelSegmentFromMemory(segmentsInGPU, idxToRemove, secondpass);
           }
           if (secondpass) {


slava77:
> The PR was built and ran successfully in standalone mode. Here are some of the comparison plots. ... Duplicate rate vs eta comparison
> 

The comparison went off scale (someone may want to contribute to improving the comparison script).
Looking at another test, e.g. https://raw.githubusercontent.com/SegmentLinking/TrackLooper-plots-archive/main/PR355_0f6afe7_standalone/dup_eta_comp.png
the baseline duplicates in the peak are just around 0.14; around x2 reduction.

slava77:
it looks like the improvement is visible, although smaller with cmssw as well.

@ariostas 
we usually keep "blue" as a reference in MTV plots so that the ratio is new/ref.
It might be good to update (it's just a matter of swapping the order of inputs to the validation script).

ariostas:
> we usually keep "blue" as a reference in MTV plots so that the ratio is new/ref.

Ah okay, I'll fix that.

slava77:
(to note from the meeting) perhaps this can be kept in the |eta|>2.5 
but then I guess a test with triplets in post-LST pLC-TCs is needed

PR #366, opened 2024-01-26
CI tweaks

ariostas:
I made the following tweaks to the CI.

Here are the changes in this PR:
- Removed the cmssw branch validation code. There was a bug where if there was a new line after the name it would think that it was invalid. This was because new lines in GitHub are \r\n, so the branch name ended up being set to "some_branch\r". The branch validation code is now delegated to the action.
- Updated a few actions that were outdated, and moved to using the version of the `TrackLooper-actions` actions in the main branch instead of a specific tag since we're the only ones using them so there's no point in keeping tags.
- Added `dependabot.yml` so that GitHub keeps the actions in the CI up to date.

Here are the changes on the `TrackLooper-actions` side:
- Fixed the branch validation code to strip command characters.
- When an integer number is given as the branch name it is now interpreted as a PR number.
- GitHub is bumped runners on public repos to 4 cores, so I updated the script to use the extra cores. It should mainly be noticeable in the standalone testing. (For the cmssw testing there are other bottlenecks that I haven't figured out how to overcome.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


slava77:
> Here are the changes on the `TrackLooper-actions` side:

you may want to consider making PRs in `TrackLooper-actions` so that updates are more clearly visible (the self-merge is OK)

slava77 (approved):

PR #371, opened 2024-02-19
Replace hardcoded values of MAX_CONNECTED_MODULES

### LINKED ISSUE #369: Increasing MAX_CONNECTED_MODULES Results in Crash
https://github.com/SegmentLinking/TrackLooper/blob/94be13002c832e9b2f8ab6ac287b07895d5cf0f4/SDL/Constants.h#L120

Increasing the above, even to just 41, results in a runtime error for an illegal memory access

Screenshot 2024-01-30 at 8 08 02 PM

ariostas:
There were a few places where the value of MAX_CONNECTED_MODULES was hardcoded. Closes #369 

Here are comparison plots showing what happens when MAX_CONNECTED_MODULES is increased from to 50, which in theory shouldn't have any effect.

TC_base_0_0_eff_ptzoom
TC_base_0_0_eff_etacoarsezoom

The differences are just the usual small run-to-run variations.


GNiendorf (approved):
Looks good, thank you for the fix!

Issue #372, opened 2024-02-29
move CMSSW base to 14_1_X


slava77:
At the end of 14_0_X cycle the pixel tracking alpaka implementation was merged (in 14_0_0_pre3). Also, alpaka 1.1.0 was integrated.
A combination of these is a good motivation to update our reference from 13_3 to 14_1.

CMSSW_14_1_0_pre0 could be a good start, although if we get to it by next week, a pre1 may be available already.

slava77:
@ariostas 
are you available to have a look? (since you made the update the last time).

ariostas:
Yeah, I'll take a look

Issue #354, opened 2023-11-07
Error with Warnings and Ntuple Writeout


GNiendorf:
If the warning printouts are enabled, and you run for example `./bin/sdl -n 175 -s 8 -w 1`  which has the ntuple writout enabled, this will throw the following error. Found by @ariostas from PR #351:

```
ERROR
 hittype: 0
 trk.pix_simHitIdx().size(): 162652
 trk.ph2_simHitIdx().size(): 213716
162652 0
206907 0
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 206907) >= this->size() (which is 162652)
```

ariostas:
The issue happens because of this line, and it only happens with CUDA. I'm still not sure why this happens.

https://github.com/SegmentLinking/TrackLooper/blob/17e0885cd71e988d57bfe619bedb07ded14735f9/SDL/Segment.h#L738

GNiendorf:
@ariostas If you serialize that kernel in CUDA then it runs fine. Very strange.

ariostas:
So it turns out that what's breaking is the atomic addition in this line.
https://github.com/SegmentLinking/TrackLooper/blob/17e0885cd71e988d57bfe619bedb07ded14735f9/SDL/Segment.h#L742

I still don't fully understand why this happens, but my guess is that since `printf` probably uses some kind of atomic operation to add stuff to the buffer that is printed out after the kernel finishes, then it ends up using some kind of atomic operation for the branching of that `if` statement, and ends up conflicting with the atomic addition. So to make sure that this doesn't happen I'll try to move the warnings to the end of kernels and far away from atomic operations.

ariostas:
I wrote minimal examples that have this bug using Alpaka and with pure CUDA. They can be found [here](https://gist.github.com/ariostas/255a01d1c3ab1dd9b85142e1bcf180e4).

slava77:
> I wrote minimal examples that have this bug using Alpaka and with pure CUDA. They can be found [here](https://gist.github.com/ariostas/255a01d1c3ab1dd9b85142e1bcf180e4).

What's the expected symptom of the bug in this example?

ariostas:
> What's the expected symptom of the bug in this example?

The two kernel calls produce different outputs. The sum in the version that is compiled with the `printf` is wrong.

slava77:
> > What's the expected symptom of the bug in this example?
> 
[...]

PR #373, opened 2024-03-01
Upgrade CMSSW to 14_1_0_pre0

ariostas:
Following @slava77's suggestion in #372, this PR upgrades CMSSW to 14_1_0_pre0. There were some minor changes that needed to be made to account for the newer version of Alpaka.

Since it now uses CUDA 12.2, I checked using the Warnings flag to see if the printf bug #354 was fixed. It wasn't fixed, but I did notice that there were some errors due to the warnings being treated as errors that I forgot to fix in #367, so I fixed them.

I marked this PR as draft since it shouldn't be merged yet for a few reasons:
- cgpu1 currently has a driver version that is not compatible with CUDA 12
- There needs to be a corresponding branch in the cmssw repo (I'll work on that)
- We could use pre1 if it happens to release soon

### DIFF
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -28,7 +28,7 @@ LIBS=$(LIB_CUDA) $(LIB_CPU)
 
 # Different architectures to optimize for
 GENCODE_SM70 := -gencode arch=compute_70,code=[sm_70,compute_70]
-GENCODE_SM89 := -gencode arch=compute_89,code=sm_89
+GENCODE_SM89 := -gencode arch=compute_89,code=[sm_89,compute_89]
 
 CXX                  = g++
 CXXFLAGS_CPU         = -march=native -mtune=native -Ofast -fno-reciprocal-math -fopenmp-simd -g -Wall -Wshadow -Woverloaded-virtual -fPIC -fopenmp -I..
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -27,7 +27,7 @@ LIBS=$(LIB_CUDA) $(LIB_CPU)
 #
 
 # Different architectures to optimize for
-GENCODE_SM70 := -gencode arch=compute_70,code=sm_70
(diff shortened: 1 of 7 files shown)


slava77 (approved):
the changes look good to me.
I don't have a strong feeling of having to wait for a pre1 release instead.
There is an effort to add cuda 12.4, perhaps that will be a better opportunity than catching pre1.

ariostas:
I added a bug fix that @slava77 found out since it's somewhat related to the CUDA upgrade and it's small enough that it doesn't really need its own PR.

`nvcc` was only generating SASS code for `sm_70` and `sm_89`. This meant that the A30 on cgpu-1 couldn't be used since it's `sm_80`. Now it also generates PTX code for `compute_70`, which can be JIT-compiled to `sm_80`.

SDL/Makefile
```diff
@@ -27,7 +27,7 @@ LIBS=$(LIB_CUDA) $(LIB_CPU)
 #
 
 # Different architectures to optimize for
-GENCODE_SM70 := -gencode arch=compute_70,code=sm_70
+GENCODE_SM70 := -gencode arch=compute_70,code=[sm_70,compute_70]
 GENCODE_SM89 := -gencode arch=compute_89,code=sm_89
```
slava77, line 31: for long term (and at least semantic consistency), perhaps also
```suggestion
GENCODE_SM89 := -gencode arch=compute_89,code=[sm_89,compute_89]
```
?

ariostas: yeah, sounds good

slava77:
> There is an effort to add cuda 12.4, perhaps that will be a better opportunity than catching pre1.

CUDA 12.4 is available since 14_1_0_pre2 (released Mar 28)

ariostas:
> CUDA 12.4 is available since 14_1_0_pre2 (released Mar 28)

I think we should wait for pre_3 since the ROCm fix didn't make it to pre_2

Issue #306, opened 2023-07-20
Remove Remaining Legacy CPU Code After Alpaka Move


GNiendorf:
[Hit.cc](https://github.com/SegmentLinking/TrackLooper/blob/alpaka_move/code/core/Hit.cc),  [Hit.h](https://github.com/SegmentLinking/TrackLooper/blob/alpaka_move/code/core/Hit.h),  [MathUtil.cc](https://github.com/SegmentLinking/TrackLooper/blob/alpaka_move/code/core/MathUtil.cc), and [MathUtil.h](https://github.com/SegmentLinking/TrackLooper/blob/alpaka_move/code/core/MathUtil.h) should be removed from /code/core and replaced with the relevant Alpaka functions where needed.

VourMa:
What do we miss to be done with this issue? From a quick look, I couldn't find where we are using the files mentioned in the issue description.

GNiendorf:
Yes, the code is still used in the GNN ntuple code: The function setGnnNtupleBranches for example.

PR #376, opened 2024-03-13
Remove cppitertools + Small Cleanup

GNiendorf:
Small cleanup PR that removes cppitertools and unused/duplicate files. I noticed that cppitertools was only used in trkCore.cc and imported but unused in a bunch of other places and figured I would remove it and get rid of the directory.


GNiendorf:
My last commit gets rid of the unused helper.cc/h and renames helper_v2.cc/h to just helper.cc/h

GNiendorf:
Last commit makes pt cut configurable with --pt_cut flag for lst_plot_performance.py and -p flag for createPerfNumDenHists.

GNiendorf:
@ariostas Not sure how to coordinate this, but after this PR is merged I think you'll have to update your CI a bit: Getting this error on the standalone "rm: cannot remove 'src/helper_v2.o': No such file or directory" since I renamed helper_v2 to just helper and deleted the old helper.cc that was just an outdated version of the _v2

ariostas:
@GNiendorf Yeah, I'll change it so that it runs `make clean && echo "skip"` so that the CI doesn't stop if it doesn't delete something.

The CMSSW check still doesn't work since I haven't updated the step2 file, but hopefully in a couple of hours it should work.

GNiendorf:
Last commit is an attempt at fixing issue #306, removing as much of the remaining legacy CPU code as I can and moving only what's used in the GNN output filling to SDLMath.h. @jkguiang Can you check the GNN ntuple to confirm there are no changes when you get a chance? Thank you.

GNiendorf:
It looks like make_classfiles is a duplicate of this code here: https://github.com/cmstas/Software/blob/739fe8d21e76d52623f8ff70e68c2f17ae923176/makeCMS3ClassFiles/make_classfiles.py#L254

GNiendorf:
I think this PR is ready for review when you get a chance @slava77.

GNiendorf:
I think the variation in the CMSSW plots is because my other PR #374 was merged in recently, so this is really showing the change due to that PR.

sgnoohc:
i don't see any problematic things. so it LGTM.

code/core/SDLMath.h
```diff
@@ -100,5 +101,141 @@ namespace SDLMath
 
     };
 
+    inline float Phi_mpi_pi(float x)
```
slava77, line 104: could replace `Helix::Phi_mpi_pi` method

efficiency/src/helper.cc
```diff
@@ -267,130 +296,34 @@ std::vector<float> getPtBounds()
     return pt_boundaries;
 }
 
-//__________________________________________________________________________________________
-/*std::vector<int> matchedSimTrkIdxs(std::vector<int> hitidxs, std::vector<int> hittypes)
+std::vector<float> getPtBounds(int mode)
```
slava77, line 299: this should replace `std::vector<float> getPtBounds()` with smth like `return getPtBounds(ana.ptbound_mode)` to keep consistent ranges/definitions

slava77 (commented):
there are a couple of simple updates that could be done now.
although with some good arguments could be done in a followup PR.

GNiendorf:
Last commit fixes known issue where the y-axis would be cut off sometimes in comparison plots. Left is before, right is after my last commit.

TC_fakerate_etacoarsezoom-1

TC_fakerate_etacoarsezoom

efficiency/src/helper.cc
GNiendorf: This code wasn't written by me but I agree. It seems like the function without a mode input is never used so I'm just going to remove it.

GNiendorf:
@slava77 This PR should be good to go if you don't have any comments about my last 2 commits.

PR #377, opened 2024-03-15
symbol decoupling for multi-arch alpaka builds

slava77:
- SDL part
    - add templates parameters to classes that generate symbols in `libsdl_*` libraries to avoid ODR violations. Wherever possible the general template is an incomplete class while the proper implementation is made for an template instance using `SDL::Acc` or `SDL::Dev`
    - factor out the non-const globals to `SDL::Globals`
    - remove most of header dependencies in LST.h to avoid exporting the details to CMSSW
    - replace TObject-related dependencies in `LST.{h,cc}`: `TString -> std::string`
    - reimplement `Event::getFullModules` and `Event::getModules` using `modulesBuffer::copyFromSrc` to reduce copy-paste
    - SDL/ModuleMethods.h : replace globally defined maps by `struct ModuleMetaData` and use an instance to replace the globals; move `struct pixelMap ` to a separate file to simplify dependencies in code/core
    - Makefile:
        - allow ROOT header dependencies in building of SDL (to keep LST.cc build common with other files)
        - use `GENCODE_CUDA` for multi-arch strings instead of separate
    - ~[todo]~ cleanup `Event::free*` declarations (all are moved to `SDL::Globals`)
- standalone:
    - build two variants sdl_cpu and sdl_cuda
    - bin/sdl.cc: add `-b` for backend, but it's currently almost useless, affects only a printout
    - ~[todo]~ add a softlink to `bin/sdl` to reuse existing tests instead of swapping two binaries

~[will update the note] still need to compile properly in a 14.1.X setup~

Other TODOs (~probably~ outside of this PR):
- improve Makeflies to directly reuse the same set of compiler arguments where they need to be; currently we have a lot of copy-pasted repeats
- factor out code independent of alpaka into a separate folder or library; e.g. `libsdl_common` and have the rest in `libsdl_alpaka_{cpu,cuda}`. Candidates for the common are `*Geometry` and related map loading methods
    - use more of these common methods in `trkCore` (esp the map loading parts)
- investigate if in the standalone it's possible to access and run and access multiple alpaka backends from the same executable


SDL/LST.cc
```diff
@@ -1,64 +1,244 @@
 #include "LST.h"
+#include "Event.h"
+#include "Globals.h"
+
+#include "Math/Vector3D.h"
+
+typedef ROOT::Math::DisplacementVector3D<ROOT::Math::Cartesian3D<float>> XYZVector;
```
slava77, line 7: this is redundant; to be removed

SDL/Event.h
```diff
@@ -171,20 +174,13 @@ namespace SDL {
...
     pixelTripletsBuffer<alpaka::DevCpu>* getPixelTriplets();
     pixelQuintupletsBuffer<alpaka::DevCpu>* getPixelQuintuplets();
-    modulesBuffer<alpaka::DevCpu>* getModules();
-    modulesBuffer<alpaka::DevCpu>* getFullModules();
+    modulesBuffer<alpaka::DevCpu>* getModules(bool isFull = false);
+
+    static void initModules(const MapPLStoLayer& pLStoLayer,
+                   const char* moduleMetaDataFilePath = "data/centroid.txt");  //read from file and init
+    static void freeModules();
+    static void freeEndcap();
```
slava77, line 182: to remove

slava77 (commented):
notes to self

VourMa:
When this is ready for review, let's check if it helps with the "temporary fix" mentioned in this [comment](https://github.com/SegmentLinking/TrackLooper/pull/374#discussion_r1523554434). Otherwise, let's make it an issue.

FYI, @GNiendorf.

slava77:
/run standalone
/run cmssw slava77:from-CMSSW_13_3_0_pre3_LST_X/arch-decoupled

slava77:
> There was a problem while building and running in standalone mode. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8398773101).

this is from ```if ! [ -f bin/sdl ]; then echo "Build failed. Printing log..."; cat .make.log*; false; fi
echo "Running LST..."```

in the multi-arch build there is no `bin/sdl` anymore. `sdl_run` would make a softlink when selected.

should I create `bin/sdl` anymore or expect/make an update in TrackLooper-actions?

slava77:
oh, there is also some asymmetry in how the build is done 

the new PR is built multi-arch
https://github.com/SegmentLinking/TrackLooper-actions/blob/b70f91d675ed6b0a09e70d96cd6bd9b245a89998/standalone/run.sh#L15-L18

while the master is built for CPU only
https://github.com/SegmentLinking/TrackLooper-actions/blob/b70f91d675ed6b0a09e70d96cd6bd9b245a89998/standalone/run.sh#L15-L18

note that I think I fixed the build of rooutil, so `make code/rooutil/` will not be necessary

slava77:
/run standalone
/run cmssw slava77:from-CMSSW_13_3_0_pre3_LST_X/arch-decoupled

slava77:
> There was a problem while building and running in standalone mode. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8403180022).


```
Line Segment Tracking binaries compilation successful!
...

sdl -i PU200 -o LSTNtuple_after.root -s 4
sdl: error while loading shared libraries: libsdl_cpu.so: cannot open shared object file: No such file or directory
```

I'm not sure what's happening

GNiendorf:
Any chance the changes in this PR allow you to remove the pixel maps as a global and instead initialize them in sdl.cc, and then pass the pointer for the pixel maps to each event object that gets created?

slava77:
> Any chance the changes in this PR allow you to remove the pixel maps as a global and instead initialize them in sdl.cc, and then pass the pointer for the pixel maps to each event object that gets created?

this proposal would affect LST.cc as well.
I'll think about it, but would rather have it in a followup PR.

slava77:
/run standalone
/run cmssw slava77:from-CMSSW_13_3_0_pre3_LST_X/arch-decoupled

ariostas:
> is there a problem to pick a branch from my repo?

Oh yeah, sorry. Currently it only works with branches of `SegmentLinking/cmssw`.

ariostas:
@slava77 actually, it should work with just the PR number

/run cmssw 18

ariostas:
Ah, so now it fails because the CMSSW PR is not merged with the default branch before testing and in this case the PR branch is based on CMSSW_13_3_X.

slava77:
> There was a problem while building and running with CMSSW. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8424153605).

my guess is that this is related to `git checkout SegLink_cmssw` instead of a merge.
My branch in https://github.com/SegmentLinking/cmssw/pull/18 is still from 13_3_0_pre3

slava77:
I will make a 14_4 branch; but we better start doing a merge in the actions so that the test is made on a code as it would be after the PR is merged instead of how it looked in the developers area

ariostas:
> I will make a 14_1 branch; but we better start doing a merge in the actions so that the test is made on a code as it would be after the PR is merged instead of how it looked in the developers area

yeah, I'll fix that

SDL/Makefile
```diff
@@ -27,20 +27,20 @@ LIBS=$(LIB_CUDA) $(LIB_CPU)
...
+CXXFLAGS_CUDA        = -O3 -g --compiler-options -Wall --compiler-options -Wshadow --compiler-options -Woverloaded-virtual --compiler-options -fPIC --compiler-options -fopenmp -dc -lineinfo --ptxas-options=-v --cudart shared $(GENCODE_CUDA) --use_fast_math --default-stream per-thread -I..
 CMSSWINCLUDE        := -I${CMSSW_BASE}/src
 ifdef CMSSW_RELEASE_BASE
 CMSSWINCLUDE        := ${CMSSWINCLUDE} -I${CMSSW_RELEASE_BASE}/src
 endif
 ALPAKAINCLUDE        = -I${ALPAKA_ROOT}/include -I/${BOOST_ROOT}/include -std=c++17 ${CMSSWINCLUDE}
 ALPAKASERIAL         = -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
 ALPAKACUDA           = -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_ACC_GPU_CUDA_ONLY --expt-relaxed-constexpr
-ROOTCFLAGS           = -pthread -m64 -I$(ROOT_ROOT)/include
+ROOTINCLUDE	     = -I$(ROOT_ROOT)/include
```
ariostas, line 42: It would be nice to replace the tab here with four spaces to keep it consistent.

bin/sdl_run
```diff
@@ -51,37 +51,41 @@ if [ -z ${SAMPLE} ]; then usage; fi
...
-  if [ ! -f ${TRACKLOOPERDIR}/bin/sdl ] || ( [ ! -f ${TRACKLOOPERDIR}/SDL/libsdl_cuda.so ] && [ ! -f ${TRACKLOOPERDIR}/SDL/libsdl_cpu.so ] ); then
-    echo "SDL library or binary has not been compiled. Please use -f flag."
-    usage
-  fi
-fi
-
 # If a backend is specified then make sure that corresponding library exists
 # and set the environment variable to preload it.
 if [ -z ${BACKEND} ]; then
   BACKEND="default"
```
ariostas, line 57: If you use `sdl_run` without `-b` then `BACKEND_INT` is not set, so the `sdl` command fails.

ariostas:
@slava77 This looks amazing. The are only a couple of minor issues. As I mentioned above, `sdl_run` doesn't work if you don't give it a `-b` parameter. It might be better to pick some default. Also, `sdl_timing` needs to be slightly adjusted so that it uses `sdl_cpu` and `sdl_cuda`. Apart from that, all the changes look good to me, and everything else works well.

> remove most of header dependencies in LST.h to avoid exporting the details to CMSSW

This is great. That's what I've been wanting to do.

> investigate if in the standalone it's possible to access and run and access multiple alpaka backends from the same executable

Now that symbols are separate, it should be relatively straightforward. (See https://github.com/fwyzard/test-rocm-cuda). However, we'll probably need to factor out `libsdl_common` before.

> When this is ready for review, let's check if it helps with the "temporary fix" mentioned in this https://github.com/SegmentLinking/TrackLooper/pull/374#discussion_r1523554434. Otherwise, let's make it an issue.

@GNiendorf Seems like this was not fixed, so let's make it an issue

> oh, there is also some asymmetry in how the build is done

At the time I did it like that to save some cpu time on the CI. The PR definitely needed to be tested for both backends, but the master branch it was fine to only compile it for cpu since the gpu compilation was supposed to already be tested.

> note that I think I fixed the build of rooutil, so make code/rooutil/ will not be necessary

Thank you, yeah that should be fine now. I'll update the CI script.

> should I create bin/sdl anyways or expect/make an update in TrackLooper-actions?
> what should be the default (cuda/cpu)?

I like the symlink approach you implemented. I'd be fine with just having separate `sdl_<backend>` binaries with no symlinks, and I can update the CI accordingly. I think since the CMSSW integration is underway we should get the ROCMm backend up and running to make sure that it works properly, but that's for a different PR.

> I will make a 14_1 branch; but we better start doing a merge in the actions so that the test is made on a code as it would be after the PR is merged instead of how it looked in the developers area

Still haven't gotten around to doing this, but it's on my to-do list.

bin/sdl_run
slava77: I defaulted to `cuda`  to preserve the historical choice.

There is a formal conflict though, the `bin/sdl` in multi-arch build defaults to `cpu` to get the CI going.

SDL/Makefile
slava77: I also updated the lines related to `CMSSW_WERRORS_CPU = ` below

bin/sdl_run
ariostas: Okay, it makes sense to default to cuda. I'll adjust the CI to take this into account.

slava77:
> Also, `sdl_timing` needs to be slightly adjusted so that it uses `sdl_cpu` and `sdl_cuda`.

I'm not sure about this one: it runs out of the box for me. 

The way `sdl_timing` was setup was that each variant was built exclusively for one backend, e.g. for cuda/GPU with `sdl_make_tracklooper -mG` in this case `sdl` points to `sdl_cuda`, via the logic triggered by  `ONLYCUDABACKEND = true`.

Did you see a problem?

slava77:
timing results
image

SDL/Globals.h
```diff
@@ -0,0 +1,32 @@
...
+  struct Globals<Dev> {
+    static SDL::modulesBuffer<Dev>* modulesBuffers;
+    static SDL::modulesBuffer<Dev> const* modulesBuffersES;  // not owned const buffers
+    static uint16_t nModules;
+    static uint16_t nLowerModules;
+    static std::shared_ptr<SDL::pixelMap> pixelMapping;
+
+    static EndcapGeometry<SDL::Dev>* endcapGeometry;
+    static TiltedGeometry<SDL::Dev> tiltedGeometry;
+    static ModuleConnectionMap<SDL::Dev> moduleConnectionMap;
```
VourMa, line 24: I am trying to understand the differences in this part of the code. What is the logic behind the fact that some objects are under the `SDL` namespace and others are not, and then the inverse is true for their template parameter?

bin/sdl.cc
```diff
@@ -220,12 +221,13 @@ int main(int argc, char** argv)
...
     //_______________________________________________________________________________
-    // check if cpu library was loaded
+    // backends
     // 0 = cpu serial
     // 1 = cpu threads
     // 2 = cuda
     // 3 = hip
-    ana.do_run_cpu = SDL::getBackend() < 2;
+    ana.backend = result["backend"].as<int>();
+    ana.do_run_cpu = ana.backend < 2;
```
VourMa, line 230: From the last commits, I understand that the default (0) for the backend is CUDA. Isn't this line contradicting with that (at least conceptually)?

ariostas:
Thank you, @slava77. You're right, `sdl_timing` works fine with these changes. I'm not sure why it didn't work for me yesterday. After the last couple of tweaks, everything works and all the changes look good to me.

ariostas (approved):

VourMa (commented):
Thanks for this huge PR, @slava77. Everything looks great to me. I left some comments mostly for my own understanding.

Makefile
```diff
@@ -30,40 +31,45 @@ CUTVALUEFLAG_FLAGS = -DCUT_VALUE_DEBUG
...
+rooutil:
 	$(MAKE) -C code/rooutil/
 
-efficiency:
+efficiency: rooutil
 	$(MAKE) -C efficiency/
 
 clean:
-	rm -f $(OBJECTS) bin/*.o $(EXES)
+	rm -f $(OBJECTS) bin/*.o $(EXES) bin/sdl
```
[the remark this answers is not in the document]
ariostas: `bin/sdl` is a symlink that gets switched to point to the binary with the selected backend. I think Slava just added it so that the CI works without any changes, but I'd be fine with removing it and adjusting the CI.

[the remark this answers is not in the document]
VourMa: I see now, I missed the lines in `sdl_make_tracklooper`. No changes required, I just wanted to understand how it worked. I am resolving the two related comments.

bin/sdl.cc
ariostas: That's a good point. It's not really affecting anything since `ana.do_run_cpu` only affects a printout, but it would be good to stick with some convention and have correct printouts.

SDL/PixelMap.h
GNiendorf, line 1: Is there a pixelmap loading function? If there is, maybe we should put in this file? To keep it consistent with the other files like TiltedGeometry/EndcapGeometry/etc.

SDL/EndcapGeometry.cc
```diff
@@ -1,27 +1,16 @@
...
-
-void SDL::freeEndcap() {
-  if (SDL::endcapGeometry != nullptr) {
-    delete SDL::endcapGeometry;
-    SDL::endcapGeometry = nullptr;
-  }
-}
-
-SDL::EndcapGeometry::EndcapGeometry(unsigned int sizef)
+SDL::EndcapGeometry<SDL::Dev>::EndcapGeometry(unsigned int sizef)
```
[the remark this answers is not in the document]
slava77: if I understand correctly, `Acc` is more complex and combines both the memory representation (relevant for `Dev`) and the execution details.
I wanted to be minimal.

I had to move from `Acc` to `Dev` for buffers in the previous PR for compatibility with CMSSW. Here the dependence is only on buffers. If we decide to expose this part to ES, it will need to stay dependent on `DeV`

SDL/Globals.h
slava77: it's due to a mix of reusing search/replace and copy-paste patterns from different places.

SDL/Event.cc
```diff
@@ -157,36 +150,25 @@ void SDL::Event::resetEvent() {
...
     modulesInCPU = nullptr;
   }
-  if (modulesInCPUFull != nullptr) {
-    delete modulesInCPUFull;
-    modulesInCPUFull = nullptr;
-  }
 }
 
-void SDL::initModules(const char* moduleMetaDataFilePath) {
+void SDL::Event<SDL::Acc>::initModules(const MapPLStoLayer& pLStoLayer, const char* moduleMetaDataFilePath) {
```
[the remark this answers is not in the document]
slava77: after looking more carefully, not for this PR, something to combine in a followup with what Gavin proposed for pixelMap initialization and not using globals

SDL/PixelMap.h
slava77: it's a bad coding pattern to have the data format to know all the ways it can possibly be filled.
We should consider cleaning up the other files to decouple the formats from the algorithms to fill them.

bin/sdl.cc
slava77: I got rid of this part in [aa92da0](https://github.com/SegmentLinking/TrackLooper/pull/377/commits/aa92da09ddf2dd7659e55e5bcfd87df9dd4cfdcc)

SDL/PixelMap.h
GNiendorf: What do you mean? I'm suggesting that if there is a function which loads the pixel maps from their txt file, it would make sense to have that function be stored in PixelMap.h where the class is defined and not scattered randomly somewhere else in the codebase.

GNiendorf: This function, shouldn't it also be moved to PixelMap.h? @slava77 

https://github.com/SegmentLinking/TrackLooper/blob/f4a2fdd7b897fa695ca6613ac8a403fe692a68f5/SDL/ModuleMethods.h#L57-L129

slava77: > What do you mean? I'm suggesting that if there is a function which loads the pixel maps from their txt file, it would make sense to have that function be stored in PixelMap.h

I've called this a bad coding pattern, suggesting to keep methods that fill the `pixelMap` and, similarly, other objects in separate algorithm functions/classes.

slava77: the argument is driven mainly by preference for code modularity and also simplifying dependencies especially for the code consuming this type `pixelMap` of data.

Why would someone/something using `pixelMap` need to know how to fill it and depend on related methods?

GNiendorf: Yes, I see what you are saying now and agree with you now as I edited my responses above after thinking about it more. We should make an issue for this since it isn't being addressed in this PR.

Issue #378, opened 2024-03-15
The previous step2 file is incompatible with CMSSW_14_1_X


ariostas:
The CI was running into segfaults with CMSSW_14_1_0_pre0 after I switched to using the full file with 100 events, but it worked briefly while it was still using a newer file that only had 10 events.

I was able to reproduce the issue on the UCSD and Cornell machines, and after some testing I verified that the issue was the old step2 file.

I'm generating a new step2 file with 100 events using CMSSW_14_1_0_pre0 and we'll have to deploy it to the CI, and the UCSD and Cornell machines.

slava77:
@ariostas   what's the `cmsDriver` command used for testing?
where is the reference step2 file coming from? 
is it one from `evourlio/LST/samples/RelValTTbarForCMSSW1300pre4/CMSSW_13_0_0_pre4/RelValTTbar_14TeV/GEN-SIM-DIGI-RAW/130X_mcRun4_realistic_v2_2026D88PU200-v1/` ?

ariostas:
@slava77 it's this one: `evourlio/LST/step2_21034.1_100Events.root`

slava77:
> @slava77 it's this one: `evourlio/LST/step2_21034.1_100Events.root`

the provenance shows that this was in CMSSW_13_0_0_pre4. I'm aware of some L1-related data format changes; but then we are not reading L1 objects, not obvious if it's still possibly related.

anyways, it sounds like it's not a part of a large production.
It should be more straightforward to reproduce this file.
I suggest to use `HLT:@fake` in step2; perhaps even try without `L1TrackTrigger,L1`: the output file will be a bit smaller and will also not have products that we are not using.

ariostas:
@slava77 @VourMa Could someone who has write permissions to `/data2/segmentlinking` on cgpu-1 create a `CMSSW_14_1_0_pre0` directory and make a symlink to `/ceph/cms/store/user/anriosta/LST/samples/CMSSW_14_1_0_pre0/step2_21034.1_100Events.root`?

VourMa:
I can do it.

VourMa:
Done!

PR #381, opened 2024-03-29
Add ROCm backend

### LINKED ISSUE #347: Add option to compile and run with HIP backend
PR #338 added support for the HIP backend, enabling the option to run on AMD or NVIDIA cards. In PRs #341 and #345 we added the option of selecting a backend at runtime, but only for CPU and CUDA. We still need to add an option to compile and run with HIP, but since we currently don't have machines with AMD cards we can postpone this for later.

ariostas:
Since the CMSSW integration is now in progress, I think it makes sense to include the ROCm backend so that the CI can check if it compiles correctly (even if we currently don't have a machine to test it).

I modified things a bit, but it still defaults to only compiling for CUDA and CPU. The backend compilation flags are no longer exclusive, so for example `-GR` would compile it for both CUDA and ROCm`.

I did have to fix a missing ifdef in `Constants.h`, but apart from that it was just adjusting scripts and updating the readme. I also removed the mention of `bin/sdl` instead mentioning `bin/sdl_<backend>`, so we can start to deprecate it.

I tested all the scripts and everything seems to be working, but I'll mark this as a draft PR for now while I make sure it's ready. The ROCm compilation flags are probably not optimal, but they work.

Closes #347

### DIFF
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -99,18 +99,12 @@ CUTVALUEFLAG_FLAGS = -DCUT_VALUE_DEBUG
 
 $(LIB_CPU): $(CCOBJECTS_CPU) $(LSTOBJECTS_CPU)
 	$(LD_CPU) $(SOFLAGS_CPU) $^ -o $@
-	mkdir -p cpu
-	ln -sf ../$@ cpu/$(@:_cpu.so=.so)
 
 $(LIB_CUDA): $(CCOBJECTS_CUDA) $(LSTOBJECTS_CUDA)
 	$(LD_CUDA) $(SOFLAGS_CUDA) $^ -o $@
-	mkdir -p cuda
-	ln -sf ../$@ cuda/$(@:_cuda.so=.so)
 
 $(LIB_ROCM): $(CCOBJECTS_ROCM) $(LSTOBJECTS_ROCM)
 	$(LD_ROCM) $(SOFLAGS_ROCM) $^ -o $@
-	mkdir -p rocm
-	ln -sf ../$@ rocm/$(@:_rocm.so=.so)
 
 explicit: $(LIBS)
 
@@ -126,9 +120,6 @@ clean:
 	rm -f *.o
 	rm -f *.d
 	rm -f *.so
-	rm -rf cpu/
-	rm -rf cuda/
-	rm -rf rocm/
 
 .PHONY: clean explicit explicit_cache explicit_cache_cutvalue format check check-fix
 
--- a/SDL/Makefile
+++ b/SDL/Makefile
@@ -6,21 +6,35 @@
 CCSOURCES=$(filter-out LST.cc, $(wildcard *.cc))
 CCOBJECTS_CPU=$(CCSOURCES:.cc=_cpu.o)
 CCOBJECTS_CUDA=$(CCSOURCES:.cc=_cuda.o)
+CCOBJECTS_ROCM=$(CCSOURCES:.cc=_rocm.o)
 
 LSTSOURCES=LST.cc
 LSTOBJECTS_CPU=$(LSTSOURCES:.cc=_cpu.o)
 LSTOBJECTS_CUDA=$(LSTSOURCES:.cc=_cuda.o)
+LSTOBJECTS_ROCM=$(LSTSOURCES:.cc=_rocm.o)
 
-LIB_CUDA=libsdl_cuda.so
-LIB_CPU=libsdl_cpu.so
-
-ifeq ($(BACKEND), cpu)
-  LIB_CUDA=
-else ifeq ($(BACKEND), cuda)
-  LIB_CPU=
+# Default to CPU and CUDA backends
+ifeq ($(BACKEND),)
+  LIB_CPU=libsdl_cpu.so
+  LIB_CUDA=libsdl_cuda.so
 endif
 
-LIBS=$(LIB_CUDA) $(LIB_CPU)
+ifneq ($(findstring cpu,$(BACKEND)),)
+  LIB_CPU=libsdl_cpu.so
+endif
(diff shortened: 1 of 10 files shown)


ariostas:
Updated the readme with the path to the new step2 file so that we can close #378.

/run standalone
/run cmssw

bin/sdl_make_tracklooper
```diff
@@ -63,17 +67,22 @@ if [ -z ${MAKECUTVALUES} ]; then MAKECUTVALUES=false; fi
...
-# If using both -G and -C, -G takes priority
-if [ "${ONLYCUDABACKEND}" == true ] && [ "${ONLYCPUBACKEND}" == true ]; then
-  echo "WARNING: -C and -G flags should not be used together."
-  echo "         Only the CUDA backend will be compiled"
-  ONLYCPUBACKEND=false
+# Default to only CPU and CUDA backends
+if [ "${CPUBACKEND}" == false ] && [ "${CUDABACKEND}" == false ] && [ "${ROCMBACKEND}" == false ]; then
+  CPUBACKEND=true
+  CUDABACKEND=true
+elif [ "${ALLBACKENDS}" == true ]; then
```
slava77, line 82: isn't `ALLBACKENDS = true` set by `A` flag (in absence of `GCR`)  going to select the first `if` and will not make it to the `elif`?

SDL/Makefile
```diff
@@ -72,16 +94,24 @@ CUTVALUEFLAG_FLAGS = -DCUT_VALUE_DEBUG
...
 
 $(LIB_CPU): $(CCOBJECTS_CPU) $(LSTOBJECTS_CPU)
 	$(LD_CPU) $(SOFLAGS_CPU) $^ -o $@
 	mkdir -p cpu
 	ln -sf ../$@ cpu/$(@:_cpu.so=.so)
 
+$(LIB_CUDA): $(CCOBJECTS_CUDA) $(LSTOBJECTS_CUDA)
+	$(LD_CUDA) $(SOFLAGS_CUDA) $^ -o $@
+	mkdir -p cuda
+	ln -sf ../$@ cuda/$(@:_cuda.so=.so)
```
slava77, line 108: do we really need these softlinks now? IIUC, the dependencies are built depending on a specific `libsdl_xyz` and the LD_LIBRARY_PATH substitution is not used anymore

bin/sdl_run
```diff
@@ -161,6 +130,37 @@ if [[ ${PRECOMPILED} != true ]]; then
...
+  fi
+  ln -s -f ${TRACKLOOPERDIR}/bin/sdl_cpu ${TRACKLOOPERDIR}/bin/sdl
+elif [ "${BACKEND}" == "rocm" ]; then
+  if [ ! -f ${TRACKLOOPERDIR}/SDL/libsdl_rocm.so ]; then
+    echo "Error: ROCM backend was not compiled."
+    exit 1
+  fi
+  ln -s -f ${TRACKLOOPERDIR}/bin/sdl_rocm ${TRACKLOOPERDIR}/bin/sdl
+elif [ "${BACKEND}" == "default" ]; then
+  if [[ ! -L ${TRACKLOOPERDIR}/bin/sdl || ! -e ${TRACKLOOPERDIR}/bin/sdl ]]; then
```
slava77, line 155: isn't `! -L` redundant here (meaning that `-e` covering all kinds of files)?

SDL/Makefile
ariostas: That's true. Those are not needed after your PR. I'll remove them.

bin/sdl_run
ariostas: Oh yeah, I'm not sure why I had it like that. I think we should remove the `-e` one since `-L` also checks if the link is valid.

ariostas: Nevermind, `-L` doesn't check if it's valid. I guess I was using `-L` to check if it's a link and `-e` to check if it's valid. But `-e` should be enough.

bin/sdl_run
```diff
@@ -50,38 +50,7 @@ if [ -z ${FLAGS} ]; then PRECOMPILED=true; else PRECOMPILED=false; fi
...
-if ${PRECOMPILED}; then
-  if [ ! -h ${TRACKLOOPERDIR}/bin/sdl ] || ( [ ! -f ${TRACKLOOPERDIR}/SDL/libsdl_cuda.so ] && [ ! -f ${TRACKLOOPERDIR}/SDL/libsdl_cpu.so ] ); then
-    echo "SDL library or binary has not been compiled. Please use -f flag."
-    usage
-  fi
-fi
+if [ -z ${BACKEND} ]; then  BACKEND="default"; fi
 
 # Check that the FLAGS start with "-" character
 if [[ ${PRECOMPILED} == true ]] || [[ ${FLAGS:0:1} == "-" ]]; then
```
ariostas, line 56: Note to self: this line seems wrong

slava77 (approved):

slava77:
I couldn't get the code/ to compile with ROCm.
`sdl_make_tracklooper -mcA` fails 
then in `g++ ... -DALPAKA_ACC_GPU_HIP_ENABLED -DALPAKA_HOST_ONLY -DALPAKA_DISABLE_VENDOR_RNG code/core/write_sdl_ntuple.cc -c -o code/core/write_sdl_ntuple_rocm.o`  there is a 
`include/alpaka/meta/CudaVectorArrayWrapper.hpp:21:18: fatal error: hip/hip_runtime.h: No such file or directory`

the include path is missing in the `CFLAGS`
I tried to add it but then there were more problems.

@ariostas did this part compile on your side?

ariostas:
> did this part compile on your side?

It definitely used to compile, but now it doesn't. I'm guessing I had modified something in my environment and forgot about it. I'll fix it.

Issue #265, opened 2023-04-07
Performance Plots Are Not Deterministic


GNiendorf:
If you run the master branch twice over the same events and create performance plots you get slightly different efficiencies each time. See below for example with the current master branch.

![TC_base_0_0_eff_ptzoom](https://user-images.githubusercontent.com/25272611/230644428-bbd97560-e6ff-4e8d-9e7b-c8d4ad02fd42.png)

VourMa:
Do we still think this is an issue? I believe that the coming of the CPU version showed us that the performance plots are deterministic.

GNiendorf:
Andres has done a lot of excellent work making the cpu/gpu difference minimal and making the code more reproducible because of that, but there are still small differences on the GPU backend if you rerun the code so I think this issue should remain open for now. @ariostas You've looked at this more than anyone now, do you know for sure what is causing these small differences? Not really important but probably good to keep in mind.

Difference on our current master:
![TC_base_0_0_eff_ptzoom](https://github.com/SegmentLinking/TrackLooper/assets/25272611/e924a7c6-2a78-4d29-9512-751b1eee1d8c)

VourMa:
> there are still small differences on the GPU backend if you rerun the code

But that can't be avoided on GPU as long as we have truncations in our code, as different things may get truncated in different runs. This is not unexpected behavior. I would see a problem if there were differences without truncations in the GPU backend, or if there were differences in the CPU backend.

ariostas:
I changed some stuff to avoid truncation, and it seems like that's the reason. You can see the plots comparing two runs [here](https://www.classe.cornell.edu/~ar2285/www/LST/comparison_no_truncation/summary/). They look pretty much identical, although in some plots there are some super tiny differences, and I'm not sure why.

However, one thing we might want to look into is that for some objects (especially segments and triplets) I had to increase maximum occupancy by a huge amount in order to avoid truncating. It might be that we're truncating a significant portion of objects. It could be that the code has changed significantly since the occupancy bounds were chosen.

GNiendorf:
@ariostas Closing this PR, thank you for fixing so many issues associated with this.

PR #383, opened 2024-04-03
Warnings cleanup

slava77:
This attempts to address compilation warnings in SDL/

- shadowing
- `alpaka::memset` 3rd argument is a byte (uint8)
- maybe-initialized
- a few consts are added in function arguments (mostly incidental; not systematic)

I think there is a bug in `runPixelQuintupletDefaultAlgo`  where the cut on `passPT5RZChiSquaredCuts` looks like it needs to be applied only for pLS pt < 5 GeV, but it's apparently applied to all because pLS pt at that point of the code is not defined.

### DIFF
--- a/SDL/Hit.h
+++ b/SDL/Hit.h
@@ -93,11 +93,11 @@ namespace SDL {
           hitRangesUpper_buf(allocBufWrapper<int>(devAccIn, nModules, queue)),
           hitRangesnLower_buf(allocBufWrapper<int8_t>(devAccIn, nModules, queue)),
           hitRangesnUpper_buf(allocBufWrapper<int8_t>(devAccIn, nModules, queue)) {
-      alpaka::memset(queue, hitRanges_buf, -1);
-      alpaka::memset(queue, hitRangesLower_buf, -1);
-      alpaka::memset(queue, hitRangesUpper_buf, -1);
-      alpaka::memset(queue, hitRangesnLower_buf, -1);
-      alpaka::memset(queue, hitRangesnUpper_buf, -1);
+      alpaka::memset(queue, hitRanges_buf, 0xff);
+      alpaka::memset(queue, hitRangesLower_buf, 0xff);
+      alpaka::memset(queue, hitRangesUpper_buf, 0xff);
+      alpaka::memset(queue, hitRangesnLower_buf, 0xff);
+      alpaka::memset(queue, hitRangesnUpper_buf, 0xff);
       alpaka::wait(queue);
     }
   };
--- a/SDL/Hit.h
+++ b/SDL/Hit.h
@@ -192,12 +192,8 @@ namespace SDL {
                                   struct SDL::modules modulesInGPU,
                                   struct SDL::hits hitsInGPU,
                                   int const& nLowerModules) const {
-      using Dim = alpaka::Dim<TAcc>;
-      using Idx = alpaka::Idx<TAcc>;
-      using Vec = alpaka::Vec<Dim, Idx>;
-
-      Vec const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
-      Vec const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+      auto const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+      auto const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
 
       for (int lowerIndex = globalThreadIdx[2]; lowerIndex < nLowerModules; lowerIndex += gridThreadExtent[2]) {
         uint16_t upperIndex = modulesInGPU.partnerModuleIndices[lowerIndex];
@@ -226,12 +222,8 @@ namespace SDL {
                                   struct SDL::hits hitsInGPU,
                                   unsigned int const& nHits) const  // Total number of hits in event
     {
-      using Dim = alpaka::Dim<TAcc>;
-      using Idx = alpaka::Idx<TAcc>;
-      using Vec = alpaka::Vec<Dim, Idx>;
-
-      Vec const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
-      Vec const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+      auto const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+      auto const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
       for (unsigned int ihit = globalThreadIdx[2]; ihit < nHits; ihit += gridThreadExtent[2]) {
         float ihit_x = hitsInGPU.xs[ihit];
         float ihit_y = hitsInGPU.ys[ihit];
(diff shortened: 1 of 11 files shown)


SDL/PixelTriplet.h
```diff
@@ -2577,7 +2575,7 @@ namespace SDL {
 
     rzChiSquared = computePT5RZChiSquared(acc, modulesInGPU, lowerModuleIndices, rtPix, zPix, rts, zs);
 
-    if (pixelRadius < 5.0f * kR1GeVf) {
+    if (/*pixelRadius*/ 0 < 5.0f * kR1GeVf) {  // FIXME: pixelRadius is not defined yet
```
slava77, line 2578: @YonsiG this is the part I was talking about during the meeting

slava77:
I think that this is ready for review; I guess we can pick a reviewer in the meeting tomorrow, unless someone volunteers sooner

SDL/Module.h
```diff
@@ -137,18 +137,18 @@ namespace SDL {
...
+      alpaka::memset(queue, hitRangesUpper_buf, 0xff);
+      alpaka::memset(queue, hitRangesnLower_buf, 0xff);
+      alpaka::memset(queue, hitRangesnUpper_buf, 0xff);
+      alpaka::memset(queue, mdRanges_buf, 0xff);
+      alpaka::memset(queue, segmentRanges_buf, 0xff);
+      alpaka::memset(queue, trackletRanges_buf, 0xff);
+      alpaka::memset(queue, tripletRanges_buf, 0xff);
+      alpaka::memset(queue, trackCandidateRanges_buf, 0xff);
+      alpaka::memset(queue, quintupletRanges_buf, 0xff);
+      alpaka::memset(queue, quintupletModuleIndices_buf, 0xff);
```
ariostas, line 151: For some of these the warning makes sense, but again I think it might be better to remove them or set them to either `~0` or `0xffffffff` since they are 32-bit integers

SDL/PixelTriplet.h
ariostas: I wonder whether this was contributing to #265. I'll check this, as well as fix the other non-deterministic kernels that @YonsiG found.

ariostas:
When compiling for cpu I get one last warning. (It's pretty easy to fix)

```bash
In file included from Kernels.h:9,
                 from Event.h:9,
                 from Event.cc:1:
In function 'void SDL::addQuintupletToMemory(triplets&, quintuplets&, unsigned int, unsigned int, uint16_t&, uint16_t&, uint16_t&, uint16_t&, uint16_t&, float&, float&, float&, float&, float&, float&, float&, float&, float&, float, float, float, float, uint8_t, unsigned int, bool)',
    inlined from 'void SDL::createQuintupletsInGPUv2::operator()(const TAcc&, SDL::modules, SDL::miniDoublets, SDL::segments, SDL::triplets, SDL::quintuplets, SDL::objectRanges, uint16_t) const [with TAcc = alpaka::AccCpuSerial<std::integral_constant<long unsigned int, 3>, long unsigned int>]' at Quintuplet.h:3104:40:
Quintuplet.h:206:52: warning: 'rzChiSquared' may be used uninitialized [-Wmaybe-uninitialized]
  206 |     quintupletsInGPU.rzChiSquared[quintupletIndex] = rzChiSquared;
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
Quintuplet.h: In function 'void SDL::createQuintupletsInGPUv2::operator()(const TAcc&, SDL::modules, SDL::miniDoublets, SDL::segments, SDL::triplets, SDL::quintuplets, SDL::objectRanges, uint16_t) const [with TAcc = alpaka::AccCpuSerial<std::integral_constant<long unsigned int, 3>, long unsigned int>]':
Quintuplet.h:3051:103: note: 'rzChiSquared' was declared here
 3051 |             float innerRadius, outerRadius, bridgeRadius, regressionG, regressionF, regressionRadius, rzChiSquared,
      |   
```

But this is really nice. It's great so see such a clean compilation log.

SDL/Hit.h
```diff
@@ -93,11 +93,11 @@ namespace SDL {
...
-      alpaka::memset(queue, hitRanges_buf, -1);
-      alpaka::memset(queue, hitRangesLower_buf, -1);
-      alpaka::memset(queue, hitRangesUpper_buf, -1);
-      alpaka::memset(queue, hitRangesnLower_buf, -1);
-      alpaka::memset(queue, hitRangesnUpper_buf, -1);
+      alpaka::memset(queue, hitRanges_buf, 0xff);
+      alpaka::memset(queue, hitRangesLower_buf, 0xff);
+      alpaka::memset(queue, hitRangesUpper_buf, 0xff);
+      alpaka::memset(queue, hitRangesnLower_buf, 0xff);
+      alpaka::memset(queue, hitRangesnUpper_buf, 0xff);
```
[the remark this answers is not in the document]
slava77: the argument type in the function is `std::uint8_t const &  	byte`; so `-1` is not good.

Or do you mean that there is another warning here after switching to `0xff`?

SDL/Module.h
slava77: that's what memset does here, it sets the values to `0xffffffff` for the destination types.
The source is a byte
https://alpaka-group.github.io/alpaka/namespacealpaka.html#aa7b4d9f42d421578d38e8923da784b37
`Sets the bytes of the memory of view to the given value.`

slava77:
> When compiling for cpu I get one last warning. (It's pretty easy to fix)

I'm curious why I don't get this warning

slava77:
> > When compiling for cpu I get one last warning. (It's pretty easy to fix)
> 
> I'm curious why I don't get this warning

are you compiling in a special setup without `DUSE_RZCHI2` ?

slava77:
> When compiling for cpu I get one last warning. (It's pretty easy to fix)

please check after 8f67f82

SDL/Hit.h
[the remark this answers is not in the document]
ariostas: Ah sorry, I didn't realized that it was settings the memory byte-wise. I thought it was a templated function that took an argument of the same type as the buffer.

SDL/Module.h
ariostas: Same comment as above.

ariostas:
> are you compiling in a special setup without DUSE_RZCHI2 ?

Oh yeah, I was testing it directly running make and I forgot to include some of the extra flags. There are no warnings now regardless of the flags.

Now that I think about it, I also need to fix LSTCore so that it uses all these extra flags.

Thank you, @slava77!

ariostas (approved):

ariostas:
I just noticed that the ROCm compilation in CMSSW seems to use a couple of extra warning flags, but I guess we can worry about that later.

SDL/PixelTriplet.h
YonsiG: Thank you for finding it Slava! I'm just curious, why do we want to only use the rzChiSquared only for < 5GeV? For larger pt, in principal, the track on rz should be very straight line

slava77: good point, in rz computations hopefully do not explode numerically and we could safely apply the cut everywhere (well, we do already, by virtue of the bug here).

Note though that `computePT5RZChiSquared` is not using the helix correction.
Why? Doesn't this lead to inefficiency at low pt?

YonsiG: Hi Slava, thank you for your explanation! I don't remember there is a reason to keep the linear constraint for pT5 r-z. I think it's just we haven't applied the helix here yet. Can add it on my to-do list if necessary.

PR #385, opened 2024-04-04
Remove device initialization from library

ariostas:
Currently, a device and queue are initialized within the library as global variables. This can cause various issues.
- Since the device is initialized as a global variable, it is done every time the library is loaded. This is what's currently causing `edmWriteConfigs` to crash when compiling the `LST` package for ROCm since the machines we use don't have any AMD cards.
- The queue part was mostly fine, but we were still creating a new queue that was not assigned to us by CMSSW. It's probably better to stick with the queue that they give us.

As far as I understand, these two things could result in using a higher (or different) computational resources than what CMSSW assigned for the job, so it can end up being killed.

This PR is still a bit rough, and I haven't worked on the CMSSW side. I'll give a better description of the changes once it's ready.


SDL/LST.cc
```diff
@@ -74,7 +76,8 @@ void SDL::LST<SDL::Acc>::loadAndFillES(alpaka::QueueCpuBlocking& queue, struct m
                            pLStoLayer);
 }
 
-void SDL::LST<SDL::Acc>::run(SDL::QueueAcc& queue,
+void SDL::LST<SDL::Acc>::run(SDL::Dev& devAccIn,
+                             SDL::QueueAcc& queue,
```
slava77, line 80: does it make sense to have a queue and a device separately? Isn't a queue uniquely attached to a device?

ariostas: That's a good point. I'll tidy things up.

slava77:
@ariostas 
what else remains to be done to make this non-draft?

ariostas:
This is ready for review. I'll elaborate a bit more regarding the changes.

As the PR title suggests, the devices (both host and accelerator) are no longer initialized in the library and are no longer global variables. Instead, accelerator queues are passed as arguments both in standalone and CMSSW. When the host device is needed, it now uses a CMSSW function to get it.

Apart from that, I also made some changes so that we use types defined in CMSSW instead of defining our own types. They mostly match, but `SDL::Idx` changed from `size_t` to `uint32_t`, which causes a bunch of annoyances. Lines like `alpaka::memcpy(queue, dst, src, 1);` failed to compile since the size now explicitly needs to be unsigned, so it would have to be replaced by `1u` or simply omitted. So I took this as a chance to do a bit of cleanup and remove the size argument when both buffers have the same size and it is obvious what it should be from the context. This make the code a bit shorter and less prone to errors if at some point some size is changed.

slava77:
/run cmssw 21

I think the last commit in 21 was after the last test was triggered

slava77 (approved):

PR #387, opened 2024-04-10
Fix race conditions in duplicate removal

ariostas:
There were some minor race conditions during duplicate removal that were found by @YonsiG. The one in `removeDupPixelQuintupletsInGPUFromMap` didn't affect anything since it only happens for a second pass, but a second pass was never done. The one in `removeDupQuintupletsInGPUBeforeTC` was causing small run-to-run variations that were contributing to #265.

The change doesn't seem to have any effect on efficiency, but the fake rate and duplicate rate decrease a tiny bit. (these plots are from cpu runs for consistency)

<table>
<tr>
<td>

</td>
<td>

</td>
<td>

</td>
</tr>
</table>

The full set of comparison plots can be found [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR387/PR387_cpu_comparison_fc97a6-PU200_09ce68-PU200/compare/).

I'll add some more plots below showing the improvement to #265.

### DIFF
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -930,11 +930,8 @@ void SDL::Event<SDL::Acc>::createPixelTriplets() {
       createWorkDiv(blocksPerGridDupPixTrip, threadsPerBlockDupPixTrip, elementsPerThread);
 
   SDL::removeDupPixelTripletsInGPUFromMap removeDupPixelTripletsInGPUFromMap_kernel;
-  auto const removeDupPixelTripletsInGPUFromMapTask(
-      alpaka::createTaskKernel<Acc>(removeDupPixelTripletsInGPUFromMap_workDiv,
-                                    removeDupPixelTripletsInGPUFromMap_kernel,
-                                    *pixelTripletsInGPU,
-                                    false));
+  auto const removeDupPixelTripletsInGPUFromMapTask(alpaka::createTaskKernel<Acc>(
+      removeDupPixelTripletsInGPUFromMap_workDiv, removeDupPixelTripletsInGPUFromMap_kernel, *pixelTripletsInGPU));
 
   alpaka::enqueue(queue, removeDupPixelTripletsInGPUFromMapTask);
   alpaka::wait(queue);
--- a/SDL/Event.cc
+++ b/SDL/Event.cc
@@ -1158,8 +1158,7 @@ void SDL::Event<SDL::Acc>::createPixelQuintuplets() {
   auto const removeDupPixelQuintupletsInGPUFromMapTask(
       alpaka::createTaskKernel<Acc>(removeDupPixelQuintupletsInGPUFromMap_workDiv,
                                     removeDupPixelQuintupletsInGPUFromMap_kernel,
-                                    *pixelQuintupletsInGPU,
-                                    false));
+                                    *pixelQuintupletsInGPU));
 
   alpaka::enqueue(queue, removeDupPixelQuintupletsInGPUFromMapTask);
 
--- a/SDL/Kernels.h
+++ b/SDL/Kernels.h
@@ -325,9 +325,7 @@ namespace SDL {
 
   struct removeDupPixelTripletsInGPUFromMap {
     template <typename TAcc>
-    ALPAKA_FN_ACC void operator()(TAcc const& acc,
-                                  struct SDL::pixelTriplets pixelTripletsInGPU,
-                                  bool secondPass) const {
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, struct SDL::pixelTriplets pixelTripletsInGPU) const {
       auto const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
       auto const gridThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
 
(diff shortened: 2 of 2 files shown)


slava77:
I'm interested to see the plots by type (before and after variants).

ariostas:
Here are the results of running LST on GPU five times for the master branch, this PR, and this PR + no truncations. The links to the full set of comparison plots are in the titles of the table.

<table>
<tr><td><a href=https://www.classe.cornell.edu/~ar2285/www/LST/PR387/multiple_runs_master_fc97a6-PU200_fc97a6-PU200_fc97a6-PU200_fc97a6-PU200_fc97a6-PU200/compare/>master</a></td><td><a href=https://www.classe.cornell.edu/~ar2285/www/LST/PR387/multiple_runs_dup_fixes_0e9ed1-PU200_0e9ed1-PU200_0e9ed1-PU200_0e9ed1-PU200_0e9ed1-PU200/compare/>this PR</a></td><td><a href=https://www.classe.cornell.edu/~ar2285/www/LST/PR387/multiple_runs_dup_fixes_notrunc_0e9ed1D-PU200_0e9ed1D-PU200_0e9ed1D-PU200_0e9ed1D-PU200_0e9ed1D-PU200/compare/>this PR + no truncations</a></td></tr>
<tr>
<td>

</td>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
<td>

</td>
</tr>
</table>

As shown in the above plots, this PR seems to help with #265 and once occupancies are increased so that nothing gets truncated the results match almost perfectly. Since the match is almost perfect, but not exactly, (as shown in the image below), there  might be another minor race condition. But it is nice to see that the match is pretty much perfect now.

image

ariostas:
> I'm interested to see the plots by type (before and after variants).

I'll regenerate the plots so that they include all the different types

VourMa:
> I'll regenerate the plots so that they include all the different types

Since the request for object breakdown has been popping up frequently the last few meetings, does it make sense to save those plots when running the CI? I tried to browse in the links above a bit but I couldn't find those plots, even though they are supposed to run in our default validation workflow (right?).

ariostas:
> even though they are supposed to run in our default validation workflow (right?).

I also thought they were generated by default, but only TC plots are on by default.

https://github.com/SegmentLinking/TrackLooper/blob/fc97a6bd2068fcb947b9f39871c7019a5afcfed6/efficiency/python/lst_plot_performance.py#L26

I think it does make sense to also generate the breakdown in the CI as long as the size doesn't increase too much.

GNiendorf:
Once #265 gets cleaned up fully, do you think it would be beneficial to add a check for race conditions in the CI? If you made a very simple parallelized CPU version and checked for differences in the output ntuple between two small runs maybe it could be done? Really nice PR btw.

ariostas:
> do you think it would be beneficial to add a check for race conditions in the CI?

That would be nice, if we actually manage to get things to perfectly match. I guess we would also need to add a toggle to increase occupancies so that things don't get truncated, but that wouldn't be too much work.

> I'm interested to see the plots by type (before and after variants).

I updated the plots so that now you can see them by type. (There's a bug in the python script, but we can just fix that later.)

The T5 plots actually now match exactly, and the pLS plots also match exactly, as far as I can tell. For pT5s and pT3s there are tiny variations. (the full set of plots are [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR387/multiple_runs_dup_fixes_notrunc_0e9ed1D-PU200_0e9ed1D-PU200_0e9ed1D-PU200_0e9ed1D-PU200_0e9ed1D-PU200/compare/)) I'll check to see if I can figure out what's happening.

image
image

slava77:
> if (score1 > score2 or ((score1 == score2) and (ix > jx))) {

how often does the second case happen?
or, related, how is the score defined? is this a relatively continuous variable?

slava77:
> > if (score1 > score2 or ((score1 == score2) and (ix > jx))) {
> 
> how often does the second case happen? or, related, how is the score defined? is this a relatively continuous variable?

actually, if ordering by index is not unique, then we could add something else well defined, like the T5 radius or something as the second tier ordering

slava77:
> Would pt be a good second tier order across the different objects or would it be better to use something more specific to each type?

considering this is going to be used very rarely, more or less any variable would be OK.
pT should be good.

ariostas:
I made all the aforementioned fixes. Now, after increasing occupancies so as to avoid truncations, all plots match exactly. You can find the comparison plots [here](https://www.classe.cornell.edu/~ar2285/www/LST/PR387/multiple_runs_all_fixes_notrunc_547774D-PU200_547774D-PU200_547774D-PU200_547774D-PU200_547774D-PU200/compare/).

/run standalone
/run cmssw

ariostas:
Oops, I messed something up

ariostas:
> It happens fairly rarely.

I take that back. Seems like it happens frequently, and I didn't find other variable that were able to discriminate them consistently further. I think this is worth exploring, but is probably out of the scope of this PR.

/run standalone
/run cmssw

slava77:
> I take that back. Seems like it happens frequently

was this explosion of duplicates from [a change](https://github.com/SegmentLinking/TrackLooper/commit/f9c9aacdd7a072a6735ae44e892b7e3cf506660c#diff-513afa5353eaddc16b1f209efb7c54130e0a0c283032f1ca762fe429573b6893R431) in `checkHitspLS` (a new addition) or in the other dup cleaners? The change in `checkHitspLS` is the clearly less restrictive case. 

For the other dup cleaners, I guess the scores can be the same if the duplicates differ only in the non-anchor hits. In that case the scores and pt etc will be the same since we are not using non-anchor hit data after MD/LS selections and the `or` that replaces `i < j` will now always fail. We should add something dependent on non-anchor hits.

slava77 (approved):

PR #389, opened 2024-04-15
Fix pT3 and pT5 construction

ariostas:
I fixed the `pixelIndexOffsetNeg` variable, which was being set incorrectly causing duplicate pT3s and pT5s to be constructed. I also replaced hardcoded values of `size_superbins`.


VourMa:
@ariostas Is this ready for review?

ariostas:
> Is this ready for review?
@VourMa yes, it's ready

VourMa (approved):
Thanks for the fix, @ariostas! It looks good to me so I am merging.

PR #391, opened 2024-04-22
Move to Binary Data Files + Include 0.6 pT Maps

GNiendorf:
This PR moves the data files to a binary format and adds the 0.6 pT maps into the data dir.

Followup PR of https://github.com/SegmentLinking/LSTGeometry/pull/11


SDL/Constants.h
```diff
@@ -145,7 +145,7 @@ namespace SDL {
 
   // Temporary fix for module buffer allocation.
   const unsigned int modules_size = 26401;
-  const unsigned int pix_tot = 1794686;
+  const unsigned int pix_tot = 1795336;
```
GNiendorf, line 148: https://github.com/SegmentLinking/LSTGeometry/blob/15f482747748e89632fbef65989355ddbb8d94d7/Constants.py#L17

I'm pretty sure this small change comes from my last PR on LSTGeometry where I consolidated the constants and changed B to be 3.8112 from 3.8 to keep it consistent with the Tracklooper definition.

ariostas (approved):
Thank you @GNiendorf, this is great! All the changes look good to me and everything is working well. It's really nice to not have a bunch of giant text files floating around.

I'll approve the PR and merge it in a few hours in case anyone else has any comments.

This is beyond the scope of this PR, but I wanted to ask you what you would think about consolidating the data into a single binary file. Using our own format like in this PR would be too messy, but it could be done with a ROOT file. The benefit I can think of is having a single smaller file (since it's compressed), but it would add extra complexity so it could be just an extra nuisance.

GNiendorf:
Thank you @ariostas. I like your suggestion of moving everything to a single file, although that's probably a lot of work for what it's worth right now. Could be a good project though for a new student or later down the line if this becomes a larger issue.

PR #392, opened 2024-04-29
remove the redundant beta cuts in T3 kernels

YonsiG:
In the T3 kernels, we have some functions called runTripletDefaultAlgoBBBB, and also the BBEE and EEEE regions. Those cuts are duplicating with the passPointingConstraintBBB ... functions. On the other hand, introducing beta iteration cuts to triplet objects are not physically meaningful. In this PR, we still keep the first beta cut as geometric requirement and move to passPointingConstraint function, but remove all the beta iteration functions.

Another improvement is that the triplet radius is calculated and saved for each of the T3s. In this case, we don't need to calculate the radius of triplet multiple times when building T5, pT3 and pT5s. This will also bring timing improvement.

Slides describing the influence of this PR is summarized here. https://indico.cern.ch/event/1407246/contributions/5914814/attachments/2846057/4976175/check%20the%20T3%20cuts%20EEEE%202.pdf

### DIFF
--- a/SDL/Triplet.h
+++ b/SDL/Triplet.h
@@ -830,104 +830,18 @@ namespace SDL {
                                                                 float& deltaBetaCut,
                                                                 float& kZ) {
     bool pass = true;
-    bool isPS_InLo = (modulesInGPU.moduleType[innerInnerLowerModuleIndex] == SDL::PS);
-    bool isPS_OutLo = (modulesInGPU.moduleType[outerInnerLowerModuleIndex] == SDL::PS);
 
     float rt_InLo = mdsInGPU.anchorRt[firstMDIndex];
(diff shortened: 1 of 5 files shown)


SDL/Triplet.h
```diff
@@ -89,6 +95,9 @@ namespace SDL {
     Buf<TDev, FPX> betaIn_buf;
     Buf<TDev, FPX> betaOut_buf;
     Buf<TDev, FPX> pt_beta_buf;
+    Buf<TDev, FPX> Circle_Radius_buf;
+    Buf<TDev, FPX> Circle_CenterX_buf;
+    Buf<TDev, FPX> Circle_CenterY_buf;
```
slava77, line 100: please keep lower case and camelCase as in the other members

YonsiG: Addressed, thank you!

PR #394, opened 2024-04-29
Fix timing issues

ariostas:
This PR fixed the timing issues found by @GNiendorf. There were two issues that caused the observed slowdowns.

1. In the standalone version, the `sdl_<backend>` binary was creating a single queue and using it for all streams. This was effectively serializing the computation into a single stream.
2. By fixing the race conditions in #387, the loops ended up checking many more objects, slowing things down. I rewrote the logic of the loops a bit to bring the timing down to the original levels.

EDIT: Removed the timing comparison since there are new commits.

### DIFF
--- a/SDL/Kernels.h
+++ b/SDL/Kernels.h
@@ -12,8 +12,9 @@
 
 namespace SDL {
   ALPAKA_FN_ACC ALPAKA_FN_INLINE void rmQuintupletFromMemory(struct SDL::quintuplets& quintupletsInGPU,
-                                                             unsigned int quintupletIndex) {
-    quintupletsInGPU.isDup[quintupletIndex] = true;
+                                                             unsigned int quintupletIndex,
+                                                             bool secondpass = false) {
+    quintupletsInGPU.isDup[quintupletIndex] |= 1 + secondpass;
   };
 
   ALPAKA_FN_ACC ALPAKA_FN_INLINE void rmPixelTripletFromMemory(struct SDL::pixelTriplets& pixelTripletsInGPU,
@@ -212,10 +213,8 @@ namespace SDL {
           float phi1 = __H2F(quintupletsInGPU.phi[ix]);
           float score_rphisum1 = __H2F(quintupletsInGPU.score_rphisum[ix]);
 
-          for (unsigned int jx1 = globalThreadIdx[2]; jx1 < nQuintuplets_lowmod; jx1 += gridThreadExtent[2]) {
+          for (unsigned int jx1 = globalThreadIdx[2] + ix1 + 1; jx1 < nQuintuplets_lowmod; jx1 += gridThreadExtent[2]) {
             unsigned int jx = quintupletModuleIndices_lowmod + jx1;
-            if (ix == jx)
-              continue;
 
             float eta2 = __H2F(quintupletsInGPU.eta[jx]);
             float phi2 = __H2F(quintupletsInGPU.phi[jx]);
@@ -231,10 +230,9 @@ namespace SDL {
 
             int nMatched = checkHitsT5(ix, jx, quintupletsInGPU);
             if (nMatched >= 7) {
-              if (score_rphisum1 > score_rphisum2) {
+              if (score_rphisum1 >= score_rphisum2) {
                 rmQuintupletFromMemory(quintupletsInGPU, ix);
-              } else if ((score_rphisum1 == score_rphisum2) && (ix < jx)) {
-                rmQuintupletFromMemory(quintupletsInGPU, ix);
+                break;
               }
             }
           }
@@ -260,7 +258,7 @@ namespace SDL {
 
         unsigned int quintupletModuleIndices_lowmod1 = rangesInGPU.quintupletModuleIndices[lowmod1];
 
-        for (unsigned int lowmodIdx2 = globalThreadIdx[2]; lowmodIdx2 < *(rangesInGPU.nEligibleT5Modules);
+        for (unsigned int lowmodIdx2 = globalThreadIdx[2] + lowmodIdx1; lowmodIdx2 < *(rangesInGPU.nEligibleT5Modules);
              lowmodIdx2 += gridThreadExtent[2]) {
           uint16_t lowmod2 = rangesInGPU.indicesOfEligibleT5Modules[lowmodIdx2];
           unsigned int nQuintuplets_lowmod2 = quintupletsInGPU.nQuintuplets[lowmod2];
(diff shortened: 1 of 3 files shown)


ariostas:
As you can see, the standalone check now includes timing information. I ran some tests and it turned out to be fairly consistent. It should be reliable enough to let us know of significant slowdowns so that this sort of thing doesn't happen again.

The efficiency has some small variations and the fake rate and duplicate rate have small increases compared to master. I'm looking to see if it's possible to keep them as close to how they were before without increasing computing time or introducing race conditions.

ariostas:
I'm done making changes now. The efficiency is unchanged, and the fake rate and duplicate rate have slight increases, which are just reverting the slight improvements in #387, but now without race conditions. Here are the new timing comparisons on cgpu-1.

```
*** This PR ***
Total Timing Summary
Average time for map loading = 514.994 ms
Average time for input loading = 8088.29 ms
Average time for SDL::Event creation = 0.0074591 ms
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       Reset    Event     Short             Rate
   avg     26.2      1.4      0.4      1.8      1.7      0.4      0.7      0.5      1.2      0.1      34.5       7.9+/-  1.5      36.9   explicit_cache[s=1]
   avg     18.4      1.6      0.5      2.4      2.1      0.4      1.1      0.7      1.7      0.2      29.1      10.3+/-  2.3      16.0   explicit_cache[s=2]
   avg     11.2      1.9      0.8      3.5      3.0      0.6      2.3      1.4      2.6      0.4      27.8      16.1+/-  3.6       7.7   explicit_cache[s=4]
   avg     21.3      2.5      1.1      4.8      4.4      0.7      3.3      2.0      4.0      0.6      44.7      22.8+/-  5.4       8.0   explicit_cache[s=6]
   avg     23.8      3.1      1.5      6.4      6.1      0.9      4.6      2.8      4.9      0.9      54.9      30.3+/-  7.2       7.4   explicit_cache[s=8]

*** master ***
Total Timing Summary
Average time for map loading = 452.888 ms
Average time for input loading = 7285.15 ms
Average time for SDL::Event creation = 0.00649407 ms
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       Reset    Event     Short             Rate
   avg     21.7      1.4      0.4      1.8      1.7      0.4      0.8      0.6      6.1      0.1      34.9      12.8+/-  6.1      37.4   explicit_cache[s=1]
   avg     10.8      1.8      0.6      5.4      2.8      0.5      2.6      2.1      6.4      0.4      33.2      22.0+/-  8.3      17.8   explicit_cache[s=2]
   avg     22.0      4.0      1.2      8.1      6.4      1.1      6.8      5.8      7.9      0.7      63.9      40.8+/- 12.1      16.7   explicit_cache[s=4]
   avg     33.0      5.5      1.4     12.7      9.3      1.6     11.1      9.9     10.1      1.6      96.2      61.7+/- 16.0      16.7   explicit_cache[s=6]
   avg     46.5      6.8      1.9     14.9     13.4      1.8     15.6     14.1     11.8      2.8     129.6      81.4+/- 19.1      16.8   explicit_cache[s=8]

*** old commit before any of these issues (fc97a6b) ***
Total Timing Summary
Average time for map loading = 823.935 ms
Average time for input loading = 7428.95 ms
Average time for SDL::Event creation = 0.0654289 ms
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       Reset    Event     Short             Rate
   avg     23.8      1.4      0.4      1.8      1.5      0.6      0.8      0.6      1.4      0.1      32.5       8.0+/-  1.6      34.0   explicit_cache[s=1]
   avg      8.2      1.5      0.6      2.4      1.9      0.6      1.4      0.9      1.9      0.2      19.6      10.8+/-  2.3      10.7   explicit_cache[s=2]
   avg     11.5      2.0      0.8      3.6      3.2      0.7      2.3      1.5      2.9      0.4      29.1      16.8+/-  3.5       7.9   explicit_cache[s=4]
   avg     16.9      2.6      1.3      5.3      4.9      0.9      3.9      2.3      4.1      0.6      42.8      25.0+/-  5.2       7.5   explicit_cache[s=6]
   avg     25.8      2.8      1.4      5.6      5.4      0.9      4.4      2.7      4.8      1.1      54.9      28.2+/-  9.6       8.1   explicit_cache[s=8]
```

I'll run the cmssw check for completeness.
/run cmssw

GNiendorf:
Screenshot 2024-04-29 at 3 00 19 PM

Timing looks normal again on 4555 as well, thank you for the fix!

SDL/Kernels.h
```diff
@@ -12,8 +12,9 @@
 
 namespace SDL {
   ALPAKA_FN_ACC ALPAKA_FN_INLINE void rmQuintupletFromMemory(struct SDL::quintuplets& quintupletsInGPU,
-                                                             unsigned int quintupletIndex) {
-    quintupletsInGPU.isDup[quintupletIndex] = true;
+                                                             unsigned int quintupletIndex,
+                                                             bool secondpass = false) {
+    quintupletsInGPU.isDup[quintupletIndex] |= 1 + secondpass;
```
GNiendorf, line 17: What is the reason for this change?

SDL/Quintuplet.h
```diff
@@ -85,7 +85,7 @@ namespace SDL {
     Buf<TDev, FPX> phi_buf;
     Buf<TDev, FPX> score_rphisum_buf;
     Buf<TDev, uint8_t> layer_buf;
-    Buf<TDev, bool> isDup_buf;
+    Buf<TDev, char> isDup_buf;
```
GNiendorf, line 88: Why was isDup changed to char? At first glance I don't see why this is required, so I'm just curious.

SDL/Kernels.h
ariostas: This is related to the change to `char`. The reason for this is that there are two passes for duplicate cleaning, first in `removeDupQuintupletsInGPUAfterBuild` and then in `removeDupQuintupletsInGPUBeforeTC`. Previously, `removeDupQuintupletsInGPUBeforeTC` skipped over quintuplets that were already marked as duplicated to save time, but there was no way to differentiate whether it was marked as duplicate in the first or second pass, which is what gives rise to the race condition. In #387 I removed skipping over quintuplets marked as duplicate so that it's always consistent, but this way it ends up doing many more comparisons, which is why it slowed down a lot. So now, what I did was change `isDup` to a char so that we can use the lowest two bits to mark duplicates in the first and second pass. So we can still skip all quintuplets that were marked as duplicates in the first pass, while keeping things consistent. This is the same thing I did in #350.

SDL/Quintuplet.h
ariostas: See comment above.

SDL/Kernels.h
ariostas: I should also mention that we're not taking up more memory after this change since both `bool` and `char` use 1 byte

GNiendorf: Got it, thank you! Merging.

GNiendorf (approved):

PR #395, opened 2024-04-30
Make code compatible with LSTCore structure

ariostas:
This PR makes the code compatible with the LSTCore structure, while still allowing it to be used as an external package like before. The changes consist of modifying the way the TrackLooper directory is found, and the include path to some headers.


SDL/EndcapGeometry.h
```diff
@@ -9,7 +9,11 @@
 #include <vector>
 #include <stdexcept>
 
+#ifdef LST_STANDALONE
 #include "Constants.h"
+#else
+#include <RecoTracker/LSTCore/interface/alpaka/Constants.h>
```
slava77, line 15:
```suggestion
#include "RecoTracker/LSTCore/interface/alpaka/Constants.h"
```
will be needed after the files are actually in CMSSW (the rule/style is to use `<>` for system and external includes

ariostas: Oh yeah sorry, I'll fix that

SDL/LST.cc
```diff
@@ -1,12 +1,27 @@
...
 
 namespace {
-  std::string trackLooperDir() { return getenv("LST_BASE"); }
+  std::string trackLooperDir() {
+    const char* path = getenv("LST_BASE");
+    std::string path_str;
+    if (path != nullptr) {
+      path_str = path;
+    } else {
+      path_str = getenv("CMSSW_BASE");
```
slava77, line 20:
```suggestion
      // FIXME: temporary solution, will need to pass a value from `FileInPath`
      // in the `LSTProducer` or a related ES producer
      path_str = getenv("CMSSW_BASE");
```

ariostas: We're not going to be able to put data files in the CMSSW repo, so I think we'll just have to rework this whole thing.

slava77:
> There was a problem while building and running with CMSSW. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8900496771).

the errors are from attempt to `#include "RecoTracker/LSTCore/interface/alpaka/Constants.h"` and such.

should we keep `LST_STANDALONE` in the current CMSSW (need a PR to SegmentLinking/cmssw) ? I suppose we are not ready yet to just merge https://github.com/SegmentLinking/cmssw/pull/19 together with this one.

ariostas:
> should we keep LST_STANDALONE in the current CMSSW (need a PR to SegmentLinking/cmssw) ?

Yeah, I forgot that in https://github.com/SegmentLinking/cmssw/pull/19 I had made some other slight modifications to the LST package. For now let's just define `LST_STANDALONE` for the CMSSW compilation while we get ready to switch to LSTCore.

SDL/LST.cc
VourMa: Even as a temporary solution, does it make sense to follow this suggestion?
https://mattermost.web.cern.ch/cms-o-and-c/pl/abax445idjyd8ec4p4ji84zahh

slava77 (approved):

PR #396, opened 2024-04-30
Upgrade to CMSSW 14_1_0_pre3

ariostas:
CMSSW 14_1_0_pre3 was released recently and it has some nice additions, including CUDA 12.4 (which fully fixes the printf issue from a while back), and fixes the ROCm compilation so that is can now compile LSTCore.

This PR goes along with the new [CMSSW_14_1_0_pre3_LST_X branch](https://github.com/SegmentLinking/cmssw/tree/CMSSW_14_1_0_pre3_LST_X) of the CMSSW repo.

This works fine on cgpu-1, but I have to make sure that it also works on the Cornell machines since they have a slightly older nvidia driver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ For lnx7188 and lnx4555 this needs to be done once
     cd /cdat/tem/${USER}/
     git clone git@github.com:SegmentLinking/LSTPerformanceWeb.git
 
-### Setting up container (only for lnx7188 and lnx4555)
+### Setting up container (only for lnx7188)
 
-For lnx7188 and lnx4555 this needs to be done before compiling or running the code:
+For lnx7188 this needs to be done before compiling or running the code:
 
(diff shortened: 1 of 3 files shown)


ariostas:
With the current setup, the ROCm version fails to link on the Cornell machines. There's an issue with incompatible versions of `libstdc++` and I haven't been able to track down why it tries to use different versions.

For lnx4555 we can simply not use Apptainer and change the `setup.sh` script so that it uses the `el9_amd64_gcc12` platform.

For lnx7188, I have tried switching the platform to `slc7_amd64_gcc12` (without Apptainer) and also tried using various different container images, but nothing works. The CUDA and CPU versions work fine, though.

So do you guys think it would be fine if I modify the `setup.sh` so that it uses `el9` for lnx4555 and `el8` for lnx7188 and cgpu-1, and we stop using Apptainer for lnx4555, and just accept that the ROCm version doesn't compile on lnx7188 (which we can't run anyway)? Maybe @GNiendorf should be the one to comment on this.

GNiendorf:
@ariostas That sounds fine to me.

GNiendorf (approved):
Looks good, will merge after CI finishes.

PR #397, opened 2024-05-03
Make 0.6 GeV Default pT Cut

GNiendorf:
PR for discussion of moving the current pT cut from 0.8 to 0.6 GeV.


slava77:
I did not remember during the meeting that there was no toggle to go to 0.6 from 0.8.

Perhaps we can first test what @VourMa proposed, to check if 0.6 GeV bin files can be used safely (no physics change and no significant slowdown) with the old default 0.8 GeV.

GNiendorf:
> Perhaps we can first test what @VourMa proposed, to check if 0.6 GeV bin files can be used safely (no physics change and no significant slowdown) with the old default 0.8 GeV.

Sure, sounds like a good check. I'll make another commit after the CI finishes running and rerun the CI with only the files changed.

slava77:
> Fake rate vs pT comparison 	

this confirms that fakes are not localized to pt<0.8 GeV. The effect is apparently from the `if (pt > 5) passThrough;` logic.
I don't think this should block/prevent the low-pt variant to become the default.
But this re-iterates the necessity to review the pass-through selection logic.

GNiendorf:
Yes, that's what I saw before. pT3's contribute most to the fakerate increase at high pT.

Screenshot 2024-05-03 at 1 58 20 PM

slava77:
> Yes, that's what I saw before. pT3's contribute most to the fakerate increase at high pT.

indeed.
the main update since then is that pT is more consistently defined and the step at 5 GeV is now explicit.

GNiendorf:
Breakdown plot from above, I see what you're saying now @slava77.

Screenshot 2024-05-03 at 2 41 29 PM

ariostas:
> There was a problem while building and running with CMSSW. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8942427586).

I just updated the CI to use the new version of CMSSW. @GNiendorf I'll restart the cmssw check.

> Any chance you could change the CI to also include the breakdown plots for the master branch in the tar file?

I did it this way to keep the size of the archive as small as possible. You could look at the commit history, and check the breakdown plots of the previous PR. But if there's strong reasons to keep both PR and master for each PR I'm happy to implement that.

GNiendorf:
> I did it this way to keep the size of the archive as small as possible. You could look at the commit history, and check the breakdown plots of the previous PR. But if there's strong reasons to keep both PR and master for each PR I'm happy to implement that.

Oh yeah fair point.

GNiendorf:
@slava77 Timing with 0.6 Maps
Screenshot 2024-05-03 at 4 31 19 PM

Current Master
Screenshot 2024-05-03 at 4 33 27 PM

slava77:
> There was a problem while building and running with CMSSW. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8945125648).

From 
```
Module: LSTOutputConverter:highPtTripletStepTrackCandidates (crashed)
```
this may need some manual debugging.

It looks like the 0.6 GeV maps are mostly OK; the CPU variant LS kernel is apparently slower (it's less visible/significant in the GPU case)

slava77:
> > There was a problem while building and running with CMSSW. The logs can be found [here](https://github.com/SegmentLinking/TrackLooper/actions/runs/8945125648).
> 
> From
> 
> ```
> Module: LSTOutputConverter:highPtTripletStepTrackCandidates (crashed)
> ```
> 
> this may need some manual debugging.
> 
> It looks like the 0.6 GeV maps are mostly OK; the CPU variant LS kernel is apparently slower (it's less visible/significant in the GPU case)

@GNiendorf 
perhaps you can reproduce this locally.
@ariostas since you know the CI setup, you may get there faster.
after it crashes locally, just run under the GDB.
`USER_CXXFLAGS="-g" scram b ... ` can be used to add the debug symbols

GNiendorf:
Given that this is an issue within the CMSSW setup, maybe @ariostas or @VourMa could take a look?

ariostas:
@GNiendorf yeah, I'll take a look

ariostas:
I opened a PR to fix the issue in https://github.com/SegmentLinking/cmssw/pull/24.

Also, looking at the logs you see that for one of the events you get this warning:
```
*********************************************************
* Warning: Pixel line segments will be truncated.       *
* You need to increase N_MAX_PIXEL_SEGMENTS_PER_MODULE. *
*********************************************************
```
So it seems like it's pretty close to the edge and it might be worth increasing that a bit.

I'll run the CI with the above PR to make sure that it works.

/run cmssw 24

slava77:
> The PR was built and ran successfully with CMSSW. Here are some plots.

@ariostas please remind me why the reference is not shown in TrackLooper+cmssw PR test? Can we show the master as is as a reference?

ariostas:
Currently it doesn't do the comparison if using a PR or a different branch, but yeah I should change it so that it does the comparison as long as the cmssw version is the same.

GNiendorf:
Timing with 0.6 GeV Threshold:
Screenshot 2024-05-09 at 12 39 05 PM
Current Master:
Screenshot 2024-05-09 at 12 41 24 PM

/run standalone

slava77:
> The PR was built and ran successfully in standalone mode. Here are some of the comparison plots.

The efficiency loss at high pt worries me. 

Does the efficiency in the muon sample look OK?

My more likely guess is that the higher pt tracks in ttbar are in jets and the duplicate cleaning there tends to select fakes instead of good tracks.

GNiendorf:
@slava77 I changed the code over to the 0.6 maps and the pls threshold to 0.6, and then changed the ptCut variable in the algorithm over for each object, one at a time. Seems like the efficiency starts getting worse when the segments ptCut is moved to 0.6 and slowly gets worse from there. Any suggestions? Ignore the label in the first plot, black is master.

Screenshot 2024-05-13 at 4 16 18 PM
Screenshot 2024-05-13 at 4 16 25 PM
Screenshot 2024-05-13 at 4 16 32 PM
Screenshot 2024-05-13 at 4 16 37 PM
Screenshot 2024-05-13 at 4 16 42 PM
Screenshot 2024-05-13 at 4 16 45 PM

GNiendorf:
@slava77 Seems like most of the efficiency loss at high pT is because the occupancies are too low. I increase the occupancies and get the following:

TC_base_0_0_eff_ptzoom-3

slava77:
> Seems like most of the efficiency loss at high pT is because the occupancies are too low.

oh, nice; this is a simpler explanation.
Does a warning for overflows come out by default or does it need a debug recompilation?

GNiendorf:
> > Seems like most of the efficiency loss at high pT is because the occupancies are too low.
> 
> oh, nice; this is a simpler explanation. Does a warning for overflows come out by default or does it need a debug recompilation?

You have to use the -w flag when compiling to get those warnings.

GNiendorf:
It seems like the T5 occupancies were set incredibly high. @YonsiG Do you know what percent they were set to? Even at 99.99% I find that the low pT occupancies are lower than the current occupancies.

slava77:
@GNiendorf 
is your stat analysis based on non-zero occupancies?
or does it increase empty modules?

GNiendorf:
> @GNiendorf is your stat analysis based on non-zero occupancies? or does it increase empty modules?

I changed it to only consider non-zero occupancies.

/run standalone
/run CMSSW

GNiendorf:
I'll run the performance plots for 1000 events and post them here.

SDL/Segment.h
```diff
@@ -965,27 +965,27 @@ namespace SDL {
...
+          occupancy = 253;
         else if (category_number == 0 && eta_number == 3)
-          occupancy = 62;
+          occupancy = 61;
         else if (category_number == 1 && eta_number == 0)
-          occupancy = 191;
+          occupancy = 1011;
         else if (category_number == 1 && eta_number == 1)
-          occupancy = 128;
+          occupancy = 653;
```
slava77, line 978: some updates are going up quite a bit, like here. On the other hand some others like in T5 https://github.com/SegmentLinking/TrackLooper/pull/397/files#diff-d9d3d7c519ab64f27b6e501281a0eb282be964d2802ddca40387348a68bb6cb0R3182 are going down

Cat1 is layers 4,5,6. It looks like something like layer 6 or even 5 already opens up a lot.
On one hand, an additional category may be useful.
On the other, it would be good to see some plots of the segment angles (is it phi change): there was a discussion to truncate this so that it doesn't hit pi/2 or even larger.

slava77:
@GNiendorf 
do you have plots showing how the new occupancy numbers were derived?

SDL/Segment.h
slava77: > Cat1 is layers 4,5,6. It looks like something like layer 6 or even 5 already opens up a lot.

given this is a segment; the possible layers are 4 and 5.

GNiendorf:
> @GNiendorf do you have plots showing how the new occupancy numbers were derived?

https://github.com/SegmentLinking/TrackLooper/blob/changept/scripts/compute_occupancies.ipynb

slava77:
> https://github.com/SegmentLinking/TrackLooper/blob/changept/scripts/compute_occupancies.ipynb

`file_path = "occ_1000_p06.root"`

is this 1K events?

GNiendorf:
> > https://github.com/SegmentLinking/TrackLooper/blob/changept/scripts/compute_occupancies.ipynb
> 
> `file_path = "occ_1000_p06.root"`
> 
> is this 1K events?

Yes, from PU200RelVal

GNiendorf:
If I increase the occupancies by 10x from where they are in this PR, I get the following:




So it seems like the efficiency at high pT is recovered fully if you put the occupancies high enough.

slava77:
> If I increase the occupancies by 10x from where they are in this PR, I get the following:
> So it seems like the efficiency at high pT is recovered fully if you put the occupancies high enough.

if the increase goes far above what we have in the master, it could be that what we observe is in part just a recovery of the inefficiency brought in by truncation initially.

How does master compare with just the increase of the occupancy cutoffs (without the change in the pt cut)?

GNiendorf:
If I just increase the occupancies on master by 10x I get the following.

TC_base_0_0_eff_ptzoom-1

GNiendorf:
This PR Timing
Screenshot 2024-05-30 at 10 08 08 PM

Master Timing
Screenshot 2024-05-30 at 10 10 55 PM

Max Memory Usage, Master (L40, 175 events, caching allocator on)
5621MiB /  46068MiB

Max Memory Usage, This PR (same setup)
9419MiB /  46068MiB

slava77:
Does ` 46068MiB` correspond to the total memory of the card less some restricted value? (specs appear to say the L40 has 48 GB)?

How well is the caching allocator shared between modules of the same process?

What is happening in case of multiple processes? I think that we can expect between 8 and 16 thread jobs populating a 256-core node with at best 2 cards. So, from 8 to as many as 32 (or more?) processes will talk to the same card.

@dan131riley 
perhaps you know this already, if memory is more restricted that what we usually see in our single job per node tests. What's a good way to test?

dan131riley:
48GB for the L40 is correct, but the CUDA device driver allocates some of that.  We actually see the same thing on the host system specs, max available mem is always less that the actual physical mem.

The caching allocator works a lot like malloc().  Within a process, memory available to the caching allocator is shared between modules so long as modules free caching allocator memory in a timely fashion.  Memory allocated by one process is held onto by the caching allocator, so isn't normally available to other processes.  I don't remember offhand if there is a call to force releasing caching allocator memory--if there is, I wouldn't expect it to be very effective because the caching allocator is likely to fragment its address space.

Last I heard I thought the preferred operating mode with GPUs (at least for HLT) was one process per socket with a GPU bound to each socket.  I don't know how that's expected to scale up to 96 cores per socket, but I can't imagine it going well.  Nvidia does have the "CUDA Multi Process Service management" (MPS) for sharing GPU resources between processes--I've used it in the past for benchmarking, but not recently.  I don't know if it figures in the HLT plans, and Nvidia seems to keep tweaking how it works with each new architecture generation.

slava77:
> Last I heard I thought the preferred operating mode with GPUs (at least for HLT) was one process per socket with a GPU bound to each socket.

Ah, indeed, it's not 8 or 16 threads.
I got `32 threads and 24 streams` for the current HLT from Andrea in https://mattermost.web.cern.ch/cms-tsg/pl/pfc9i5nt8pf788btfg7y8ybowy in January.

GNiendorf:
Closing, see new PR: https://github.com/SegmentLinking/cmssw/pull/39

Issue #380, opened 2024-03-29
Remove PixelMaps + ModuleMaps from Globals


GNiendorf:
See this conversation for more details: https://github.com/SegmentLinking/TrackLooper/pull/374#discussion_r1523509608
Also discussed in @slava77's recent PR: https://github.com/SegmentLinking/TrackLooper/pull/377

When the code was moved from CUDA to Alpaka, I had to put the sizes of the maps as constants in Constants.h here:
https://github.com/SegmentLinking/TrackLooper/blob/ad8ed39d7a701ce505363352f2b406c82d3744a6/SDL/Constants.h#L143-L148

This is because the module and pixel maps are global variables, and I could not find a way to calculate the sizes at runtime and make the relevant arrays on the GPU using Alpaka. CUDA is much more relaxed about this, so this was not an issue before we moved to Alpaka. The downside of having the sizes stored in the Constants.h is that if you change the pixel maps in /data/ and the number of connections changes then it would throw a runtime error. To make this clear and avoid future issues, I put warnings in the code to check that the sizes in Constants.h were the sizes calculated at runtime. See the below

https://github.com/SegmentLinking/TrackLooper/blob/ad8ed39d7a701ce505363352f2b406c82d3744a6/SDL/EndcapGeometry.cc#L41-L47

https://github.com/SegmentLinking/TrackLooper/blob/ad8ed39d7a701ce505363352f2b406c82d3744a6/SDL/ModuleMethods.h#L80-L86

The best solution here is to remove these variables as globals (probably initialize them in sdl.cc and pass their pointer to each Event object that is created), calculate their sizes at runtime when they are initialized, and then allocate the relevant arrays with these computed sizes. @slava77's recent PR fixed a lot of the issues I ran into when I attempted to do this when moving the code over to Alpaka, so I think this can be done now.

PR #401, opened 2024-05-13
Remove global variables

ariostas:
This PR removes the remaining global variables. The standalone version works now, but I still need to work on the CMSSW side. I'll mark it as ready for review once everything works and there's a companion PR on the CMSSW repo.


GNiendorf:
This PR will also remove the sizes defined in Constants.h right? Now that these are no longer globals. https://github.com/SegmentLinking/TrackLooper/issues/380

ariostas:
> This PR will also remove the sizes defined in Constants.h right? Now that these are no longer globals. https://github.com/SegmentLinking/TrackLooper/issues/380

Oh yeah, I wasn't planning on it, but we should do that on this PR too

ariostas:
> Oh yeah, I wasn't planning on it, but we should do that on this PR too

That turned out to be much more convoluted than I anticipated. I'll try again once this PR is more complete, but I might also defer it to a future PR.

ariostas:
I think it should all be working now. I'll test it with the CI while I check if there's things that could be cleaned up a bit. Also, I'll check to see how much it needs to be changed to fix the issue @GNiendorf mentioned.

/run standalone
/run cmssw 26

ariostas:
After thinking about it a bit more, I think that a way to make this cleaner is to have the load functions construct the shared pointers instead of only filling them, which would make #380 much easier to address. Also, I should probably move the `LSTESData.h` header that I added in https://github.com/SegmentLinking/cmssw/pull/26 to here so we can pass around all this data more cleanly. I'll ask for opinions during the meeting today and I'll have it ready within the next couple of days.

SDL/Event.h
```diff
@@ -84,10 +84,37 @@ namespace SDL {
...
     int8_t* pixelTypeCPU;
 
+    // Stuff that used to be global
+    uint16_t nModules;
+    uint16_t nLowerModules;
+    std::shared_ptr<modulesBuffer<Dev>> modulesBuffers;
+    std::shared_ptr<pixelMap> pixelMapping;
+    std::shared_ptr<EndcapGeometry<Dev>> endcapGeometry;
+    std::shared_ptr<TiltedGeometry<Dev>> tiltedGeometry;
+    std::shared_ptr<ModuleConnectionMap<Dev>> moduleConnectionMap;
```
slava77, line 94: can this all be const?
This is required for cases where data can be used directly from the event setup; otherwise we'd have to make a copy somewhere, which is not good

ariostas: Okay, I can adjust things to make them const. It will be easier once I make the change I mentioned above.

bin/sdl.cc
```diff
@@ -303,10 +303,25 @@ void run_sdl()
         queues.push_back(SDL::QueueAcc(devAcc));
     }
 
+    // Create objects that used to be global
+    uint16_t nModules;
+    uint16_t nLowerModules;
+    std::shared_ptr<SDL::modulesBuffer<SDL::Dev>> modulesBuffers = std::make_shared<SDL::modulesBuffer<SDL::Dev>>(devAcc);
```
slava77, line 309: minor

```suggestion
    auto modulesBuffers = std::make_shared<SDL::modulesBuffer<SDL::Dev>>(devAcc);
```
could be less verbose

bin/sdl.cc
```diff
@@ -303,10 +303,25 @@ void run_sdl()
...
+    std::shared_ptr<SDL::pixelMap> pixelMapping = std::make_shared<SDL::pixelMap>();
+    std::shared_ptr<SDL::EndcapGeometry<SDL::Dev>> endcapGeometry = std::make_shared<SDL::EndcapGeometry<SDL::Dev>>(devAcc);
+    std::shared_ptr<SDL::TiltedGeometry> tiltedGeometry = std::make_shared<SDL::TiltedGeometry>();
+    std::shared_ptr<SDL::ModuleConnectionMap> moduleConnectionMap = std::make_shared<SDL::ModuleConnectionMap>();
     // Load various maps used in the SDL reconstruction
     TStopwatch full_timer;
     full_timer.Start();
-    loadMaps(devAcc, queues[0]);
+    SDL::LST<SDL::Acc>::loadAndFillES(queues[0],
```
slava77, line 317: nice cleanup of redundant code

SDL/EndcapGeometry.cc
```diff
@@ -82,4 +69,9 @@ void SDL::EndcapGeometry<SDL::Dev>::fillGeoMapArraysExplicit(SDL::QueueAcc& queu
   alpaka::wait(queue);
 }
 
-float SDL::EndcapGeometry<SDL::Dev>::getdxdy_slope(unsigned int detid) { return dxdy_slope_[detid]; }
+float SDL::EndcapGeometry<SDL::Dev, true>::getdxdy_slope(unsigned int detid) const {
+  return (detid < dxdy_slope_.size() ? dxdy_slope_.at(detid) : 0);
```
ariostas, line 73: This part was reading data out of bounds. We should check why that is happening and how that was affecting performance.

ariostas:
This is PR is close to being ready for review. I did end up fixing #380, but it needed quite a few changes. I made things as const as possible. Also, I found out the out of bounds issue I mentioned above, which we should look into. I'll run the CI to make sure that things look fine. I still have to work on the CMSSW side.

/run standalone

SDL/EndcapGeometry.cc
GNiendorf: Isn't dxdy_slope_ a map? Why does this change make sense?

ariostas:
Thanks @GNiendorf for pointing out that `dxdy_slope_` was a map, not a vector.

/run standalone

SDL/EndcapGeometry.h
```diff
@@ -18,28 +18,41 @@
...
 namespace SDL {
-  template <typename TDev>
-  class EndcapGeometry {};
+
+  // Only the full one contains alpaka buffers
+  template <typename TDev, bool full>
+  class EndcapGeometry;
+
+  template <>
+  class EndcapGeometry<DevHost, false> {
```
slava77, line 27: isn't this an ODR violation? `EndcapGeometry<DevHost, false>` will appear in all alpaka backend builds.

A separate non-template `EndcapGeometryZZZ` (put something clear in ZZZ) would be more appropriate; it will be in non-alpaka part of LSTCore

SDL/Event.h
```diff
@@ -84,10 +86,27 @@ namespace SDL {
     int* superbinCPU;
     int8_t* pixelTypeCPU;
 
+    // Stuff that used to be global
+    const uint16_t nModules;
+    const uint16_t nLowerModules;
+    const unsigned int nPixels;
+    const std::shared_ptr<const modulesBuffer<Dev>> modulesBuffers;
+    const std::shared_ptr<const pixelMap> pixelMapping;
+    const std::shared_ptr<const EndcapGeometry<Dev, true>> endcapGeometry;
```
slava77, line 95: CMSSW naming rules suggest using a trailing underscore for (private) data members (`m_` or `the` or some distrinct prefix/suffix are also acceptable). This makes the implementation more readable.
This PR could be a good opportunity to denote these are data members now

SDL/LST.h
```diff
@@ -17,22 +17,76 @@ namespace SDL {
...
+                    std::shared_ptr<pixelMap> pixelMappingIn)
+        : nModules(nModulesIn),
+          nLowerModules(nLowerModulesIn),
+          nPixels(nPixelsIn),
+          modulesBuffers(std::const_pointer_cast<const modulesBuffer<Dev>>(modulesBuffersIn)),
+          endcapGeometry(std::const_pointer_cast<const EndcapGeometry<Dev, true>>(endcapGeometryIn)),
+          pixelMapping(std::const_pointer_cast<const pixelMap>(pixelMappingIn)) {}
+  };
+
+  std::unique_ptr<LSTESHostData> loadAndFillESHost();
```
slava77, line 73: another ODR violation

SDL/ModuleConnectionMap.h
```diff
@@ -15,11 +15,7 @@
 #endif
 
 namespace SDL {
-  //FIXME: move to non-alpaka single arch build
-  template <typename>
-  class ModuleConnectionMap;
-  template <>
-  class ModuleConnectionMap<SDL::Dev> {
+  class ModuleConnectionMap {
```
slava77, line 18: the `FIXME` was here for a reason

SDL/TiltedGeometry.h
```diff
@@ -16,10 +16,7 @@
 #endif
 
 namespace SDL {
-  template <typename>
-  class TiltedGeometry;
-  template <>
-  class TiltedGeometry<SDL::Dev> {
+  class TiltedGeometry {
```
slava77, line 19: and another n-plicate definition

slava77 (commented):
a few classes and methods will now appear as visible symbols in all `sdl_` library backend builds.
Is there some first principles justification that this is safe?

SDL/EndcapGeometry.h
ariostas: I think that's not an ODR violation. That's why I added the second template parameter so that for cpu builds it doesn't end up having a duplicate definition. But I agree that a non-template class would be better.

SDL/Event.h
ariostas: That's a good point, thanks!

ariostas:
> a few classes and methods will now appear as identical visible symbols in all sdl_ library backend builds.
> Is there some first principles justification that this is safe?

For now, should we keep using templated classes even for things that don't use any alpaka buffers or should we figure out how to move them into a non-alpaka part now? As far as I can tell, CMSSW only loads one library at a time, so there probably wouldn't be symbol conflicts, but I'm not sure if in more general scenarios there could be issues.

ariostas:
I think for now it doesn't matter if we have potential symbol conflicts. When we move to CMSSW (hopefully pretty soon) it will be a lot easier to separate non-alpaka parts by moving them out of the `alpaka` directory.

SDL/LSTESData.h
```diff
@@ -0,0 +1,94 @@
...
+    const std::shared_ptr<const MapPLStoLayer> mapPLStoLayer;
+    const std::shared_ptr<const EndcapGeometryHost<Dev>> endcapGeometry;
+    const std::shared_ptr<const TiltedGeometry<Dev>> tiltedGeometry;
+    const std::shared_ptr<const ModuleConnectionMap<Dev>> moduleConnectionMap;
+
+    LSTESHostData(std::shared_ptr<MapPLStoLayer> mapPLStoLayerIn,
+                  std::shared_ptr<EndcapGeometryHost<Dev>> endcapGeometryIn,
+                  std::shared_ptr<TiltedGeometry<Dev>> tiltedGeometryIn,
+                  std::shared_ptr<ModuleConnectionMap<Dev>> moduleConnectionMapIn)
+        : mapPLStoLayer(std::const_pointer_cast<const MapPLStoLayer>(mapPLStoLayerIn)),
```
slava77, line 50: is the cast actually needed?

slava77: https://godbolt.org/z/o8Ms81c97 from non-const to const works

as for the member constness, it's probably an overkill, since ES will share only `const LSTESHostData`

ariostas: Ah okay, I'll fix that

slava77:
/run standalone

@ariostas will the previous run cmssw actually complete or does the request/job get killed by the PR update?

ariostas:
> will the previous run cmssw actually complete or does the request/job get killed by the PR update?

It will complete, but let's run a new one to make sure that the new changes don't break anything.

/run cmssw 26

slava77:
> Here is a timing comparison:

pLS column changed from 543 to 498 (a similar reduction as in a test [on May 14](https://github.com/SegmentLinking/TrackLooper/pull/401#issuecomment-2110502247) , but not on [May 17](https://github.com/SegmentLinking/TrackLooper/pull/401#issuecomment-2117984555)); I'm curious if a rerun (after the more technical const update) will confirm the reduction.

ariostas:
There seems to be some issue with CUDA. I'll look into it.

ariostas:
I moved some functions back into an unnamed namespace. I didn't realize that there was a reason for it. Now it also works with CUDA.

I ran a timing comparison on cgpu-1, but the pLS time is too small to notice a difference.
```
This PR
Total Timing Summary
Average time for map loading = 477.75 ms
Average time for input loading = 7560.01 ms
Average time for SDL::Event creation = 0.0114327 ms
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       Reset    Event     Short             Rate
   avg     19.5      1.3      0.4      1.5      1.5      0.4      0.6      0.5      1.2      0.1      27.0       7.1+/-  1.4      29.2   explicit_cache[s=1]
   avg      7.0      1.4      0.6      2.0      1.9      0.4      1.0      0.7      1.7      0.2      16.9       9.4+/-  2.1       9.8   explicit_cache[s=2]
   avg      9.6      1.8      0.8      3.1      3.1      0.5      2.0      1.2      2.7      0.4      25.3      15.2+/-  3.2       7.0   explicit_cache[s=4]
   avg     15.2      2.3      1.2      4.3      4.2      0.6      3.1      1.9      3.8      0.5      37.2      21.3+/-  4.0       6.7   explicit_cache[s=6]
   avg     21.4      2.9      1.5      5.5      5.9      0.8      4.8      2.7      4.7      1.0      51.3      29.1+/-  6.8       6.8   explicit_cache[s=8]

master
Total Timing Summary
Average time for map loading = 445.219 ms
Average time for input loading = 7561.65 ms
Average time for SDL::Event creation = 0.0083333 ms
   Evt    Hits       MD       LS      T3       T5       pLS       pT5      pT3      TC       Reset    Event     Short             Rate
   avg     20.4      1.3      0.4      1.5      1.5      0.4      0.6      0.5      1.2      0.1      27.9       7.1+/-  1.4      30.3   explicit_cache[s=1]
   avg      6.8      1.4      0.5      2.0      1.9      0.4      1.0      0.7      1.7      0.2      16.6       9.4+/-  2.0       9.6   explicit_cache[s=2]
   avg      9.6      1.8      0.8      3.1      3.0      0.5      2.0      1.2      2.6      0.4      25.0      14.9+/-  3.5       7.0   explicit_cache[s=4]
   avg     15.0      2.3      1.2      4.3      4.4      0.6      3.2      1.9      3.8      0.5      37.3      21.6+/-  4.7       6.7   explicit_cache[s=6]
   avg     21.5      3.0      1.5      5.7      5.7      0.8      4.5      2.5      5.0      0.9      51.0      28.7+/-  8.2       6.8   explicit_cache[s=8]
```

ariostas:
I also verified that it works with GPU in CMSSW, so it's all ready now.

slava77 (approved):

PR #402, opened 2024-05-14
pT3 new pt definiton

YonsiG:
This introduces a new definition of pT3 pt in the standalone write out. Average the pt from radius and pLS pt has a narrower resolution than only using the pt from pLS, compared to sim track pt.
slides can be found here: https://indico.cern.ch/event/1410981/

### DIFF
--- a/SDL/Triplet.h
+++ b/SDL/Triplet.h
@@ -912,7 +912,7 @@ namespace SDL {
 
             uint16_t outerOuterLowerModuleIndex = segmentsInGPU.outerLowerModuleIndices[outerSegmentIndex];
 
-            float zOut, rtOut, deltaPhiPos, deltaPhi, betaIn=0, circleRadius=-999, circleCenterX, circleCenterY;
+            float zOut, rtOut, deltaPhiPos, deltaPhi, betaIn = 0, circleRadius = -999, circleCenterX, circleCenterY;
             float zLo, zHi, rtLo, rtHi, zLoPointed, zHiPointed, sdlCut, betaInCut;
 
             bool success = runTripletConstraintsAndAlgo(acc,
--- a/SDL/Triplet.h
+++ b/SDL/Triplet.h
@@ -913,7 +913,7 @@ namespace SDL {
 
             uint16_t outerOuterLowerModuleIndex = segmentsInGPU.outerLowerModuleIndices[outerSegmentIndex];
 
-            float zOut, rtOut, deltaPhiPos, deltaPhi, betaIn, circleRadius=-999, circleCenterX, circleCenterY;
+            float zOut, rtOut, deltaPhiPos, deltaPhi, betaIn=0, circleRadius=-999, circleCenterX, circleCenterY;
             float zLo, zHi, rtLo, rtHi, zLoPointed, zHiPointed, sdlCut, betaInCut;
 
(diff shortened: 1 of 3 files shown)


YonsiG:
Performance comparison separately for break up objects
master: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/T3_remove_redundant/pT3_pt_redef/master_PU200_NEVT-1_37e559-PU200/compare/
new pt for pT3: http://uaf-10.t2.ucsd.edu/~yagu/SDL_GPU_plots/T3_remove_redundant/pT3_pt_redef/new_def_PU200_NEVT-1_a9bf9a-PU200/compare/

SDL/Triplet.h
```diff
@@ -913,7 +912,7 @@ namespace SDL {
 
             uint16_t outerOuterLowerModuleIndex = segmentsInGPU.outerLowerModuleIndices[outerSegmentIndex];
 
-            float zOut, rtOut, deltaPhiPos, deltaPhi, betaIn, circleRadius, circleCenterX, circleCenterY;
+            float zOut, rtOut, deltaPhiPos, deltaPhi, betaIn = 0, circleRadius = -999, circleCenterX, circleCenterY;
```
slava77, line 915: (from inspecting and cleaning up some other cases), I think it's better to make the initialization later in a place where it's clear that the good value is not going to be computed. This would be around the last `return pass` that can lead to a `success == true`.
This way we'll benefit from the compiler diagnostic to tell us that the variable is not used before it's set to something useful.

YonsiG: Thank you for suggestion. However, I am still a bit confused, how can the compiler diagnostics tell us the variable is not used if we initialize them and give them values anyway?

slava77: my comment would address the use cases of these variables inside `runTripletConstraintsAndAlgo`.

YonsiG: I added an initialization after the runTripletConstraintsAndAlgo in the later commit, if the "success" is not correct, initialize all the variables as -999. Does this look ok?

SDL/Triplet.h
```diff
@@ -941,6 +940,16 @@ namespace SDL {
                                                         zHiPointed,
                                                         sdlCut,
                                                         betaInCut);
+            if (!success) {
+              zOut = -999;
+              rtOut = -999;
```
slava77, line 945: I don't think this is the right place to add this change.

`if (!success)` case is not going to be used at all.
The point was to add the initialization inside `runTripletConstraintsAndAlgo`

Did the compiler warning even go away about `betaIn`?

slava77: I was looking at the underlying calls and for `betaIn`, there is no initialization of it in `passPointingConstraintBBB`

For all other variables, did the compiler complain about them being not initialized?

YonsiG: No, only the betaIn complains. Other variables don't have complaints

slava77 (approved):

slava77:
the plots look as expected: 
- no differences in cmssw part; 
- changes in the fake rates and duplicate rates are due to the migration of reconstructed pt to new values
- the timing increase in the "rate" seems to be random (by column it shows up more in MD and Hits, not related to this PR)

PR #404, opened 2024-05-21
Add Occupancy Branches Back

GNiendorf:
Removed in #212, adding them back so that I can tune the occupancies for the low pT threshold. Keeping this as a draft PR since it seems like some branches are broken and this needs more cleanup.

### DIFF
--- a/code/core/write_sdl_ntuple.cc
+++ b/code/core/write_sdl_ntuple.cc
@@ -322,16 +322,15 @@ void setOptionalOutputBranches(SDL::Event<SDL::Acc>* event)
 //________________________________________________________________________________________________________________________________
 void setOccupancyBranches(SDL::Event<SDL::Acc>* event)
 {
-    SDL::trackCandidates& trackCandidatesInGPU = (*event->getTrackCandidates());
-    SDL::triplets& tripletsInGPU = (*event->getTriplets());
-    SDL::segments& segmentsInGPU = (*event->getSegments());
-    SDL::miniDoublets& mdsInGPU = (*event->getMiniDoublets());
-    //SDL::hits& hitsInGPU = (*event->getHits());
     SDL::modules& modulesInGPU = (*event->getModules());
+    SDL::miniDoublets& mdsInGPU = (*event->getMiniDoublets());
+    SDL::segments& segmentsInGPU = (*event->getSegments());
+    SDL::triplets& tripletsInGPU = (*event->getTriplets());
     SDL::quintuplets&  quintupletsInGPU = (*event->getQuintuplets());
     SDL::pixelQuintuplets& pixelQuintupletsInGPU = (*event->getPixelQuintuplets());
     SDL::pixelTriplets& pixelTripletsInGPU = (*event->getPixelTriplets());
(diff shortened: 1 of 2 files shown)


code/core/write_sdl_ntuple.cc
```diff
@@ -297,10 +314,80 @@ void setOptionalOutputBranches(SDL::Event<SDL::Acc>* event)
...
+        moduleRod.push_back(modulesInGPU.rods[lowerIdx]);
+        moduleEta.push_back(modulesInGPU.eta[lowerIdx]);
+        moduleR.push_back(modulesInGPU.r[lowerIdx]);
+        isTilted = (modulesInGPU.subdets[lowerIdx] == 5 and modulesInGPU.sides[lowerIdx] != 3);
+        moduleIsTilted.push_back(isTilted);
+        moduleModule.push_back(modulesInGPU.modules[lowerIdx]);
+        segmentOccupancy.push_back(segmentsInGPU.totOccupancySegments[lowerIdx]);
+        mdOccupancy.push_back(mdsInGPU.totOccupancyMDs[lowerIdx]);
+
+        if(idx < *(modulesInGPU.nLowerModules))
```
GNiendorf, line 368: @YonsiG or @slava77 Do you know why the t5 and t3 occupancies are indexed from 0 to nLowerModules-1 and the md and segment occupancies are indexed from 0 to nLowerModules?

GNiendorf: I decided to get rid of this, I don't think it will matter when choosing the occupancies.

code/core/write_sdl_ntuple.cc
```diff
@@ -297,10 +314,72 @@ void setOptionalOutputBranches(SDL::Event<SDL::Acc>* event)
...
+        moduleRod.push_back(modulesInGPU.rods[lowerIdx]);
+        moduleEta.push_back(modulesInGPU.eta[lowerIdx]);
+        moduleR.push_back(modulesInGPU.r[lowerIdx]);
+        bool isTilted = (modulesInGPU.subdets[lowerIdx] == 5 and modulesInGPU.sides[lowerIdx] != 3);
+        moduleIsTilted.push_back(isTilted);
+        moduleModule.push_back(modulesInGPU.modules[lowerIdx]);
+        segmentOccupancy.push_back(segmentsInGPU.totOccupancySegments[lowerIdx]);
+        mdOccupancy.push_back(mdsInGPU.totOccupancyMDs[lowerIdx]);
+        quintupletOccupancy.push_back(quintupletsInGPU.totOccupancyQuintuplets[lowerIdx]);
+        tripletOccupancy.push_back(tripletsInGPU.totOccupancyTriplets[lowerIdx]);
```
VourMa, line 363: I can see in #212 that these lines are within an `if(idx < *(modulesInGPU.nLowerModules))` clause. I can't remember why that was there but is it not needed anymore?

GNiendorf: Added this back like we discussed.

GNiendorf:
@VourMa This PR should be good to go.

VourMa (approved):

PR #406, opened 2024-05-24
Dynamic Allocation Tests

GNiendorf:
I see a 5x-10x reduction (from 500k to 50k-100k per event) in the nTotalQuintuplets variable from the changes below from @slava77's suggestion. Using this PR for tests. Obviously no timing difference with the cache enabled, but I do see improvement when it is turned off:

Timing with dynamic allocation (this PR)
Screenshot 2024-05-24 at 4 57 39 PM

Master Timing
Screenshot 2024-05-24 at 4 53 25 PM

### DIFF
--- a/SDL/Quintuplet.h
+++ b/SDL/Quintuplet.h
@@ -3131,15 +3131,13 @@ namespace SDL {
       alpaka::syncBlockThreads(acc);
 
       // Initialize variables outside of the for loop.
-      int occupancy, category_number, eta_number;
+      int occupancy = 0;
 
       for (int i = globalThreadIdx[2]; i < *modulesInGPU.nLowerModules; i += gridThreadExtent[2]) {
         // Condition for a quintuple to exist for a module
         // TCs don't exist for layers 5 and 6 barrel, and layers 2,3,4,5 endcap
-        short module_rings = modulesInGPU.rings[i];
         short module_layers = modulesInGPU.layers[i];
         short module_subdets = modulesInGPU.subdets[i];
-        float module_eta = alpaka::math::abs(acc, modulesInGPU.eta[i]);
 
         if (tripletsInGPU.nTriplets[i] == 0)
           continue;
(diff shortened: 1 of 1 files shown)


SDL/Quintuplet.h
```diff
@@ -3150,52 +3148,7 @@ namespace SDL {
...
-          occupancy = 191;
-        else if (category_number == 3 && eta_number == 3)
-          occupancy = 106;
-        else {
-          occupancy = 0;
-#ifdef Warnings
-          printf("Unhandled case in createEligibleModulesListForQuintupletsGPU! Module index = %i\n", i);
-#endif
-        }
+        occupancy = 2 * tripletsInGPU.nTriplets[i];
```
slava77, line 3151: this will look bad; probably more so at high pt.
But we need breakdown plots; CI is not showing them

GNiendorf: Here are the breakdown plots, I don't see any issues:

https://www.classe.cornell.edu/~gsn27/www/www/PR406/

slava77: ehm, I was wrong at least at base/integrated  level  ;)

slava77:
@GNiendorf 
did you run this locally?
I'm curious how often do we see a truncation warning with this `2*n_T3` vs the old case limit.
Just counting the warnings may be enough.

GNiendorf:
I'll check the warning printouts and post the results soon. I also want to try this method out on the other objects.

SDL/Quintuplet.h
slava77: > You can also see the breakdown plots from the CI by downloading the full zip file.

ah OK.



> Here are the breakdown plots. I don't see any issues, maybe I'm missing something though.

yes, it looks OK; it takes an effort to see a difference.

The other side (opposite to high pt high occupancy) to worry about is the low end: a module with one triplet can have more than two matching. This may become more visible without PU.

slava77: the low occupancy end could be moderated by e.g. `2*n_T3 + 10` ("random" guess for values; I'm still implying some plots need to be made to show the boundary is OK).

GNiendorf:
@slava77 Excess alert printouts go way up with these changes, although a lot of these alerts come from the same module numbers.

t5 excess warnings (175 events)- 
45,927 - with my changes
85 - Master

Here are the current excess alerts for the other objects as a reference.
388,149 - sement excess alert 
104,239 - triplet excess alert
6,291 - Minidoublet excess alert

slava77:
> Excess alert printouts go way up with these changes, although a lot of these alerts come from the same module numbers. Maybe these are mostly duplicates that get cleaned up later on?

Are the counts on low end of the distribution or somewhat spread out?
I'm curious if my https://github.com/SegmentLinking/TrackLooper/pull/406#discussion_r1614111534 could help easily

do we have per-module truncated and non-truncated occupancies to be able to get some more detail where to truncate?

slava77:
> Maybe these are mostly duplicates that get cleaned up later on?

it could be. The main problem I can expect now would be in variability of results.

Unless there are exact duplicates that contribute significantly; offhand I think we don't have any, but then I seem to recall that there was a recent thread suggestive that there may be.

GNiendorf:
I'm not sure what an acceptable amount of cutoff would be, especially since the other objects seem to get cutoff at a much higher rate. Here I added an offset of +30 and the excess printouts go from 45,927 to 3,106. We still see a 5x improvement on most events over the current master for the nQuints allocation size.

GNiendorf:
Closing this PR since Tracklooper is discontinued.

VourMa:
Do you intend to port it to SegmentLinking/cmssw, or is this development discontinued as a whole?

PR #407, opened 2024-05-29
fix typo in Triplet.h line 352

jchismar:
Replaced + with *

### DIFF
--- a/SDL/Triplet.h
+++ b/SDL/Triplet.h
@@ -349,7 +349,7 @@ namespace SDL {
     float drt_InSeg = rtMid - rtIn;
     float dz_InSeg = zMid - zIn;
     float dr3_InSeg =
-        alpaka::math::sqrt(acc, rtMid * rtMid + zMid * zMid) - alpaka::math::sqrt(acc, rtIn * rtIn + zIn + zIn);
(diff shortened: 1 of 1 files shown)


slava77 (approved):

SDL/Triplet.h
```diff
@@ -349,7 +349,7 @@ namespace SDL {
     float drt_InSeg = rtMid - rtIn;
     float dz_InSeg = zMid - zIn;
     float dr3_InSeg =
-        alpaka::math::sqrt(acc, rtMid * rtMid + zMid * zMid) - alpaka::math::sqrt(acc, rtIn * rtIn + zIn + zIn);
+        alpaka::math::sqrt(acc, rtMid * rtMid + zMid * zMid) - alpaka::math::sqrt(acc, rtIn * rtIn + zIn * zIn);
```
slava77, line 352: this would can potentially affect BBB T3s: the value of `dr3_InSeg` was slightly larger with the bug, which meant a slightly larger z selection window based on the track direction due to a larger component accounting for the multiple scattering. Realistically only T3s starting in barrel layer 1 may show something, but it looks like it was not visible even there.

PR #408, opened 2024-05-29
Make compilation flags for duplicate cleaning and triplet pLSs into runtime flags

VourMa:
As per title. The default behavior should not change, as it will be (hopefully) demonstrated by the tests, while I have run comparisons for the application of the flags, and the results are as expected:

**Master vs. no duplicate cleaning**
https://uaf-10.t2.ucsd.edu/~evourlio/SDL/PR408/runtimeFlags__VS_2_enabled_50fb47D-PU200_50fb47D-PU200/summary/

**Master vs. triplet pLSs in TCs**
https://uaf-10.t2.ucsd.edu/~evourlio/SDL/PR408/runtimeFlags__VS_3_enabled_50fb47D-PU200_50fb47D-PU200/summary/

It goes together with the corresponding PR in cmssw (SegmentLinking/cmssw#31).


SDL/LST.h
```diff
@@ -9,6 +9,11 @@
 #include "LSTESData.h"
 #endif
 
+#include "Event.h"
+
+#include "Math/Vector3D.h"
```
slava77, line 14: can these be avoided?
Especially the first one; to keep the dependencies cleaner.

VourMa: These were not added, they were just moved from the `.cc` to the `.h` file. Unless the movement is the problem.

slava77: > Unless the movement is the problem.

yes

VourMa: Sure, I will revert.

VourMa: Reverted in [37b465a](https://github.com/SegmentLinking/TrackLooper/pull/408/commits/37b465af7330a92688c3ceb58416e25d7946b871). I also included back the explicit `SDL/Event.h` dependency for `sdl.h` to make it clearer, even though it is implicitly included through `write_sdl_ntuple.h`.

VourMa:
Superseded by SegmentLinking/cmssw#37.

PR #410, opened 2024-06-11
New pT5 r-z chi2 cut definition study based on helix approximation

YonsiG:
The original pT5 rz chi2 cut is based on linear approximation. This PR updates the trajectory estimation as a helix, the helix does not necessarily need to start from the origin. However, it still takes the pixel information as input momentum and initial point. 
After this update, the pT5 efficiency in transition region is significantly improved. While also, the fake rate is reduced.


VourMa:
Superseded by SegmentLinking/cmssw#40

PR #412, opened 2024-07-29
LSTOD LS/MD Changes

Hoobidoobidoo:
Added LS/MD cut info into the LSTNtuple.root output and fixing the plotting scripts to print new variables from ntuple.

### DIFF
--- a/SDL/MiniDoublet.h
+++ b/SDL/MiniDoublet.h
@@ -1,5 +1,6 @@
 #ifndef MiniDoublet_cuh
 #define MiniDoublet_cuh
+#define OUTPUT_MD_CUTS
 
 #include "Constants.h"
 #include "EndcapGeometry.h"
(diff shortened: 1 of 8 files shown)


VourMa:
Before this PR goes much further, I wanted to remind you that the development now continues over to the SegmentLinking/cmssw repo, so these developments have to be ported there. You should also keep an eye for the renaming updates that have happened/are happening.

I also have a curiosity: Which machine are you testing these developments on? If it is on some machine in Florida, which setup file do you `source` when before compiling?

@sgnoohc FYI

ariostas:
@Hoobidoobidoo are you still working on this? If so, I'll help you rebase it to the CMSSW side, but let's do that sooner rather than later so that we don't have pending things on ancient versions of the code.

@slava77 maybe we should mark this repo as a read-only archive?

slava77:
> @slava77 maybe we should mark this repo as a read-only archive?

fine with me

